PR #1, opened 2020-03-06
GH-673: Add docs and update repo settings

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/jira/software/projects/GH/boards/530?selectedIssue=GH-673

### Changes

- Replace most of the `Zero`s in the docs with `WFL`. Leaving the `zero`s in the code snippets as they are. 
- Update gitignore file.
- Setting up a static doc site based on Github Pages which is generated from markdown files automatically.

### Review Instructions

- Check https://broadinstitute.github.io/wfl/


docs/docs/README.md
```diff
@@ -0,0 +1,581 @@
...
+[WorkFlow Launcher (WFL)](https://github.com/broadinstitute/wfl.git) 
+is a workload manager.
+
+It runs as you, with your credentails, from your laptop, and
+communicates with other services as necessary to manage a workload.
+
+It can also be deployed to run as a service in the cloud.
+
+For more on Workflow Launcher's role in the Terra infrastructure see 
+[Workflow Launcher's role in Terra (./docs/docs/terra.md)](./terra.md).
```
rhiananthony: Why does this doc tell the user to refer to terra here and then below describes it's role anyway? Should it instead link to either the area below or delete the area below and keep the role this plays in terra?

docs/docs/README.md
```diff
@@ -0,0 +1,581 @@
+# Welcome to WorkFlow Launcher
+
+## Overview
+
+[WorkFlow Launcher (WFL)](https://github.com/broadinstitute/wfl.git) 
+is a workload manager.
```
rhiananthony: This is a bit confusing if one is not familiar with the workflow launcher.

docs/docs/README.md
```diff
@@ -0,0 +1,581 @@
...
+Workflow Launcher needs to manage its version and the versions of
+`dsde-pipelines.git` which contribute the WDL files. There may be as
+many `dsde-pipelines.git` versions as there are workflow wdls.
+
+The `wfl` jar includes a manifest with at least that information in
+it, and a version command that returns it.
+
+## Role in Terra
+
+The Data Sciences Platform (DSP) is building a new system (around
```
rhiananthony: This describes the new role in relation to the old role. Will outside people know what zamboni is? Should they as a prerequisite to understanding what the WFL is? If not, then we shouldn't refer to what zamboni is or does as that will be confusing. Instead we should just say what the WFL does and is.

docs/docs/README.md
```diff
@@ -0,0 +1,581 @@
...
+    Clojure development feels very different from Scala and Java
+    development. It even differs markedly from development in other
+    *dynamic languages* such as Python or Ruby.
+    
+    Get a demonstration from someone familiar with Clojure
+    development before you spend too much time trying to figure
+    things out on your own.
+    
+    Find a local Cursive user for guidance if you like IntelliJ.
+    [Jay Carey](mailto:jcarey@broadinstitute.org) and [Charley
```
rhiananthony: Both of these people no longer work at the Broad.

docs/docs/README.md
```diff
@@ -0,0 +1,581 @@
...
+    ```
+    
+    They permit fast testing of code changes by storing expressions
+    that you can evaluate in your editor buffer.
+    
+    Feel free to add, edit, or augment them as you see fit.
+
+6.  Documentation
+    
+    This file is written in markdown.
```
rhiananthony: This seems unimportant...

docs/docs/README.md
```diff
@@ -0,0 +1,581 @@
...
+    
+    3.  Sources
+        
+          - /Users/tbl/Broad/zamboni/Client/src/scala/org/broadinstitute/zamboni/client/lightning/clp/Lightning.scala
+          - /Users/tbl/Broad/picard-private/src/java/edu/mit/broad/picard/lightning
+          - /Users/tbl/Broad/gppipeline-devtools/release<sub>client</sub>
+          - /Users/tbl/Broad/gppipeline-devtools/starter<sub>control</sub>
+          - /picard02:/seq/pipeline/gppipeline-devtools/current/defs/prod.defs
+
+#### Cruft
```
rhiananthony: Should this section be removed?

docs/docs/server.md
```diff
@@ -1,21 +1,21 @@
...
 
-We now have the basics of Zero
+We now have the basics of WorkFlow Launcher
 running as a server
 in Google App Engine (GAE).
 
 ## Deploy to Google App Engine
 
-To build and deploy Zero,
+To build and deploy WFL,
```
rhiananthony: Should we spell out Workflow Launcher for all of these replacements in the docs?

docs/docs/terra.md
```diff
@@ -0,0 +1,296 @@
...
+engineers to debug problems and resume or restart failed workflows.
+Zamboni can run workflows on both a local Sun Grid Engine (SGE), and
+on Cromwell on premises and in the cloud.
+
+We think that WFL can fill the role of Zamboni in the new data
+storage and processing system that DSP is developing now.
+
+## History
+    
+WFL began as a project to replace a Zamboni *starter*, with the old
```
rhiananthony: Same comment as before - do future users need to know about Zamboni? If so, then keep this. If not, then it will just confuse them.

docs/docs/README.md
rexwangcc: Good question, I guess when we factored out the role stuff to a separate file, we didn't update this readme. I will delete the below section.

docs/docs/README.md
rexwangcc: Zamboni and the whole background is explained in the standalone `terra.md`. I think it still makes sense to keep it since a lot of users will still be those who work close with us. What do you think?

docs/docs/README.md
rexwangcc: I will update this.

docs/docs/README.md
rexwangcc: Removed!

docs/docs/server.md
rexwangcc: I thought about it but realized it would be too verbose... I think having the title as `WorkFlow Luncher (WFL)` should be enough to setup the context for the readers of this doc, what do you think?

docs/docs/README.md
rexwangcc: I agree, not sure how could make it more explicit. @tbl3rd ?

docs/docs/README.md
rhiananthony: I can live with it. It's up to what you and the team thinks. Just felt that I should ask the question :)

docs/docs/server.md
rhiananthony: That's fine. I agree it's verbose.

docs/docs/README.md
rhiananthony: Definitely think that it should be limited to one section though :)

tbl3rd:
Thanks for doing this and the updates.
Would be good to document the database as well.

docs/docs/README.md
tbl3rd: I don't have time to work on this now. Much improvement is possible.
Be careful not to make it less useful for Broadies while attempting to
make it more appropriate for a general audience.

I doubt WFL will ever be very useful to outsiders. A lot of work will be
necessary before that happens … if ever.

tbl3rd:
IWBNI There were some navigational overview of all the documents in the tree.

rexwangcc:
> Would be good to document the database as well.

Yeah, that's definitely some worthy documenting. Since I haven't looked into the details about the DB and migration parts, I'd be only confident with adding the page, and maybe Saman could add the accurate details to it when we have time later?

docs/docs/README.md
rexwangcc: Good catch, we don't support sv anymore, so makes sense to remove this.

PR #4, opened 2020-03-13
GH-690: Fix versioning and staging of resources in WFL again.

tbl3rd:
https://broadinstitute.atlassian.net/browse/GH-690


wdl/ExternalWholeGenomeReprocessing.wdl
```diff
@@ -5,8 +5,6 @@ import "CopyFilesFromCloudToCloud.wdl" as Copy
 
 workflow ExternalWholeGenomeReprocessing {
 
-  String pipeline_version = "1.2"
```
rexwangcc: This was here since we thought we should give the adapter wdls same version as their scientific counterparts.

tbl3rd: Why?

src/zero/boot.clj
```diff
@@ -95,83 +67,89 @@
...
-    (util/delete-tree (io/file tmp))))
+  (letfn [(frob [{:keys [release top] :as wdl}]
+            [(last (str/split top #"/")) release])]
+    (let [wdls [ukb/workflow-wdl wgs/workflow-wdl xx/workflow-wdl]
+          {:keys [tmp] :as clones} (clone-repos)
+          directory (io/file resources zero/the-name)
+          edn (merge version
+                     (dissoc clones :tmp)
+                     (into {} (map frob wdls)))]
+      (pprint edn)
```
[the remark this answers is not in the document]
tbl3rd: No. Why?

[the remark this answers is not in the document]
rexwangcc: NVM, I forgot this is in build step and worried this might cause too much noise; noise in build is good.

wdl/ExternalWholeGenomeReprocessing.wdl
rexwangcc: If you submit this external workflow without having workflow versions, won't people ask which semantic version is this? I also agree having the hash is enough info this is more like redundant version that we have to maintain overtime.

tbl3rd: The subworkflow WDL is part of the Cromwell input.
The release version is there if people want it.
I don't think a semantic version adds anything here,
and we would have to maintain it somehow.

src/zero/boot.clj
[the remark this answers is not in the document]
tbl3rd: boot build always dumped the version.
Looking for a reason that it shouldn't.

docs/md/README.md
```diff
@@ -267,28 +267,28 @@ the services WFL talks to, and are named accordingly.
...
-| cromwell.clj | Cromwell workflow runner                   |
+| cromwell.clj | Cromwell workflow runner                  |
 | datarepo.clj | DSP DataRepo                              |
 | db.clj       | On-prem and Cloud SQL databases           |
 | gcs.clj      | Google Cloud Storage                      |
 | jms.clj      | Java Message Service queues               |
 | postgres.clj | Cloud SQL postgres databases              |
 | pubsub.clj   | Google Cloud Pub/Sub                      |
-| server.cl    | the WFL server itself                    |
+| server.cl    | the WFL server itself                     |
```
samanehsan: `server.clj`?

src/zero/zero.clj
```diff
@@ -8,31 +8,15 @@
...
-  (str "pipeline-config-" (UUID/randomUUID) "/"))
-
-(def pipeline-config-url
-  "Clone URL for the pipeline-config repo"
-  "git@github.com:broadinstitute/pipeline-config.git")
-
 (def the-github-repos
   "Map Zero source repo names to their URLs"
-  (let [repos ["zero" "dsde-pipelines"]
+  (let [repos ["wfl" "dsde-pipelines" "pipeline-config"]
```
samanehsan: So concise! 💯:octocat:

tbl3rd: 1 less character!

docs/md/README.md
tbl3rd: :sigh: thanks.

PR #7, opened 2020-03-17
GH-682: Update WFL schema to include typed tables.

tbl3rd:
https://broadinstitute.atlassian.net/browse/GH-682

Checkpoint new schema so I don't lose it again.

WARNING: Running the migration will `drop table workload` on the `postgres` database.

Should `workload.pipeline` be named `workload.workflow` instead?
Or maybe change the name of the composite type to `pipeline`?
What say ye?


database/migration/changesets/20200315_CreateWorkloadTable.xml
```diff
@@ -0,0 +1,123 @@
...
+        <comment>
+            Define a new root workload table schema.
+            There is a row in this table for every workload.
+            The 'table' column here names another table.
+            That other table has the composite Postgres type 'workflow',
+            and is named with the 'pipeline' column
+            composed with the 'id' column of this table.
+        </comment>
+        <rollback>
+            I have no idea.
```
rexwangcc: 🤔 what is supposed to be in this block?

database/migration/changesets/20200316_ExternalWholeGenomeReprocessing.xml
```diff
@@ -0,0 +1,43 @@
...
+        <sql>
+            CREATE TYPE ExternalWholeGenomeReprocessing AS (
+            id                     bigint, -- unique to this table
+            base_file_name         text,   -- input to the workflow
+            cromwell               text,   -- the Cromwell metadata URL
+            destination_cloud_path text,   -- input to the workflow
+            final_gvcf_base_name   text,   -- input to the workflow
+            input_cram             text,   -- input to the workflow
+            sample_name            text,   -- input to the workflow
+            unmapped_bam_suffix    text)   -- input to the workflow
```
rexwangcc: ❤️ Nice, I think for now these fields are sufficient to provide flexible input options.

database/migration/changesets/20200315_CreateWorkloadTable.xml
tbl3rd: It's spozed to be how to rollback the changset.
I should just delete those, since I don't think they matter.
Liquibase does not support automated rollback for SQL blocks.

database/migration/changesets/20200315_CreateWorkloadTable.xml
```diff
@@ -0,0 +1,123 @@
...
+            </column>
+            <column name="cromwell" type="text"
+                    remarks="The URL of the Cromwell instance
+                             running the workflows.
+                             Should be a common prefix
+                             of the 'cromwell' column
+                             in the 'table' table.">
+                <constraints nullable="false"/>
+            </column>
+            <column name="finished" type="timestamptz"
```
samanehsan: What do you think about calling this `end_time`? It makes it clearer that this field refers to the time that the workflow finished and not whether it finished

database/migration/changesets/20200315_CreateWorkloadTable.xml
```diff
@@ -0,0 +1,123 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="release" type="text"
+                    remarks="The release tag of the WDL
+                    for workflows in this workload
+                    or the Git commit hash of the .wdl file
+                    if this WDL is not officially released.">
+                <constraints nullable="false"/>
+            </column>
+            <column name="started" type="timestamptz"
```
samanehsan: Could this be "start_time" instead?

database/migration/changesets/20200315_CreateWorkloadTable.xml
```diff
@@ -0,0 +1,123 @@
...
+                             in 'table'.">
+                <constraints nullable="false"/>
+            </column>
+            <column name="output" type="text"
+                    remarks="Common prefix
+                             of the output URLs
+                             in 'table'.">
+                <constraints nullable="false"/>
+            </column>
+            <column name="pipeline" type="workflow"
```
samanehsan: Where is the `workflow` type defined?

samanehsan:
Since we do not have any workload information stored in the database, what do you think about cleaning up the changelog history by removing "changesets/01_db_schema.xml" and "changesets/20200314_DropWorkloadTable.xml"? (We'd need to drop all of the tables in the DB before re-running LB again with the remaining changeset files) :octocat:

database/migration/changesets/20200315_CreateWorkloadTable.xml
tbl3rd: In the `<sql>` block above.

database/migration/changesets/20200315_CreateWorkloadTable.xml
tbl3rd: A column is more than a name. It also has a type.

database/migration/changesets/20200315_CreateWorkloadTable.xml
tbl3rd: It is typed `timestamptz`, not `boolean`.
And we obviously cannot know when it finished until it's finished.

tbl3rd:
> Since we do not have any workload information stored in the database, what do you think about cleaning up the changelog history by removing `changesets/01_db_schema.xml` and `changesets/20200314_DropWorkloadTable.xml`? (We'd need to drop all of the tables in the DB before re-running LB again with the remaining changeset files) :octocat:
> 
> Although in the future/production we would have to change everything with a changelog...so I am also okay with leaving it as is.

Yes. I left it in to test that these changesets could properly run as updates to some existing schema.
We may want to clean out the initial ones after we successfully create our first `workflow` table.

tbl3rd:
> I'd prefer not to have `workflow` and `workload` sitting too close, they look so similar...

I agree. Suggest an alternative.

database/migration/changesets/20200315_CreateWorkloadTable.xml
samanehsan: Thanks, I totally missed that 🤦‍♀

PR #9, opened 2020-03-19
Work around the PostgreSQL enum transaction error.

tbl3rd:
```
2020-03-19 16:10:33.155 EDT [23812] ERROR:  ALTER TYPE ... ADD cannot run inside a transaction block
2020-03-19 16:10:33.155 EDT [23812] STATEMENT:  ALTER TYPE pipeline ADD VALUE 'ExternalWholeGenomeReprocessing'
```


tbl3rd:
BTW: I have not deployed this change yet, but it works locally.

PR #10, opened 2020-03-23
Return 401 error when logged out

samanehsan:
### Purpose
Currently, if you are not logged in the WFL API returns a redirect response. However, swagger will return `{"status":"ok"}` in this case, presumably because the redirect was successful. This is confusing behavior because it makes it seem like the request itself was successful.

### Changes
Make the `authorize` function return a 401 error if a user is not logged in.

### Review Instructions
- No instructions.

### DIFF
--- a/src/zero/api/handlers.clj
+++ b/src/zero/api/handlers.clj
@@ -55,7 +55,10 @@
         (if-let [jwt (some-> request :oauth2/access-tokens :google :id-token
                              decode-jwt valid?)]
           (handler (assoc request :jwt jwt))
-          (response/status (response/response "") 401))))))
+          (-> (response/response {:message "Unauthorized"})
+              (response/header "WWW-Authenticate" "Bearer realm=API access")
+              (response/content-type "application/json")
+              (response/status 401)))))))
 
 (defn succeed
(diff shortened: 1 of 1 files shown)


src/zero/api/handlers.clj
```diff
@@ -55,7 +55,7 @@
         (if-let [jwt (some-> request :oauth2/access-tokens :google :id-token
                              decode-jwt valid?)]
           (handler (assoc request :jwt jwt))
-          (response/redirect landing-uri))))))
+          (response/status (response/response "") 401))))))
```
tbl3rd: Seems a shame to waste that `""`.

tbl3rd: And maybe add an `Authenticate` header?

tbl3rd: https://tools.ietf.org/html/rfc7235#section-4.1
… though maybe Ring does that for us?

samanehsan: I'm not sure...I'll take a look!

samanehsan: It's not added by Ring, so I added it in :octocat: Lmk what you think!

PR #12, opened 2020-03-24
Check authorization headers in API requests

samanehsan:
### Purpose
The WFL API is only checking requests for a google oauth access token.

### Changes
Update the `authenticate` fn to check if the authorization header of an incoming request contains a token.

### Review Instructions

- No instructions.

### DIFF
--- a/src/zero/api/handlers.clj
+++ b/src/zero/api/handlers.clj
@@ -59,7 +59,7 @@
                               last)))]
       (if-let [jwt (some-> request token decode-jwt valid?)]
         (handler (assoc request :jwt jwt))
-        (-> (response/response {:response "Unauthorized"})
+        (-> (response/response {:response {:message "Unauthorized"}})
             (response/header "WWW-Authenticate" "Bearer realm=API access")
             (response/content-type "application/json")
             (response/status 401))))))
--- a/src/zero/api/handlers.clj
+++ b/src/zero/api/handlers.clj
@@ -41,28 +41,27 @@
 
 (defn authorize
   "Add JWT to REQUEST when it contains a broadinstitute.org account.
-  Otherwise redirect HANDLER to the LANDING-URI."
+  Otherwise return a 401 Unauthorized response."
   [handler]
-  (let [{:keys [client-id]} (:google oauth2-profiles)]
-    (fn [request]
-      (letfn [(valid? [{:keys [payload] :as jwt}]
-                (let [{:keys [aud exp hd iss]} payload]
-                  (and (= hd  "broadinstitute.org")
-                       (= iss "https://accounts.google.com")
-                       (= aud client-id)
-                       (> exp (quot (System/currentTimeMillis) 1000))
-                       jwt)))]
-        (let [oauth-token (some-> request :oauth2/access-tokens :google :id-token)
-              bearer-token (some-> request
-                                   (response/get-header "authorization")
-                                   (str/split #" ")
-                                   last)]
-          (if-let [valid-token (some-> (or oauth-token bearer-token) decode-jwt valid?)]
-            (handler (assoc request :jwt valid-token))
-            (-> (response/response {:message "Unauthorized"})
-                (response/header "WWW-Authenticate" "Bearer realm=API access")
-                (response/content-type "application/json")
-                (response/status 401))))))))
+  (fn [request]
+    (letfn [(valid? [{:keys [payload] :as jwt}]
+              (let [{:keys [aud exp hd iss]} payload]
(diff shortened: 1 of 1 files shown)


src/zero/api/handlers.clj
```diff
@@ -43,7 +43,7 @@
   "Add JWT to REQUEST when it contains a broadinstitute.org account.
   Otherwise redirect HANDLER to the LANDING-URI."
   [handler]
-  (let [{:keys [client-id landing-uri]} (:google oauth2-profiles)]
+  (let [{:keys [client-id]} (:google oauth2-profiles)]
```
tbl3rd: Does your prior PR break Google OAuth?

src/zero/api/handlers.clj
```diff
@@ -55,10 +55,17 @@
...
+                                        (response/get-header "authorization")
+                                        (str/split #" ")
+                                        last
+                                        decode-jwt
+                                        valid?)]
+            (handler (assoc request :jwt bearer-token))
+            (-> (response/response {:message "Unauthorized"})
+                (response/header "WWW-Authenticate" "Bearer realm=API access")
+                (response/content-type "application/json")
+                (response/status 401))))))))
```
tbl3rd: This would be better as two token sources validated the same way
instead of two completely distinct extraction and validation paths.

src/zero/api/handlers.clj
samanehsan: No, Google OAuth works through the UI like it has been. But when I send an API request outside of the UI, where would I get that token from?

samanehsan: So that's why I created a JWT from a service account and sent it with the authorization headers...but this authorize function wasn't checking that 🤔

tbl3rd: Thank you for checking.
Probably should remove `LANDING-URI` from the docstring then.
And maybe replace the top `let` with a `get-in`?

src/zero/api/handlers.clj
samanehsan: ^ Does that look better @tbl3rd?

src/zero/api/handlers.clj
samanehsan: Ahh I forgot to update the docstring 🤦‍♀

src/zero/api/handlers.clj
tbl3rd: Better, but howzabout: (untested)
``` clojure
(letfn [(valid? [{:keys [payload] :as jwt}]
              (let [{:keys [aud exp hd iss]} payload]
                (and (= hd  "broadinstitute.org")
                     (= iss "https://accounts.google.com")
                     (= aud (get-in oauth2-profiles [:google :client-id]))
                     (> exp (quot (System/currentTimeMillis) 1000))
                     jwt)))
            (token [request]
              (get-in request [:oauth2/access-tokens :google :id-token]
                      (some-> request
                              (response/get-header "authorization")
                              (str/split #" ")
                              last)))]
      (if-let [jwt (some-> request token decode-jwt valid?)]
        (-> (response/response request)
            (response/header "WWW-Authenticate" "Bearer realm=API access")
            (response/content-type "application/json")
            (response/status 401))))
```
Maybe just return `request` for debugging.
Worried that no one will be looking for a `:message` anyway.
I also worry about the `split`, but that's probably OK.

src/zero/api/handlers.clj
tbl3rd: I forget all the time. Only noticed on the 3rd read, FWIW.

src/zero/api/handlers.clj
tbl3rd: Indentation looks messed up, so please check the syntax in a real editor.

tbl3rd: Or you could `assoc-in` an `"Unauthorized"` result message.

samanehsan: Thanks for the suggestion -- this looks much clearer now! :octocat: The only thing is I get a 500 error when I try to return the `request` as the response body

tbl3rd: Weird. Ah! Try something like this if your up for it.
``` clojure
{:request request :response {:message "Unauthorized"}}
```
Or … just merge it and maybe we'll take a later crack at it.

tbl3rd: Ring is probably trying trying to interpret the request as a response.

samanehsan: Weirdly even including `{:request request}` causes a 500 error. I thought the `request` was just a map 🤔 Anyways I think we can revisit this later! Thanks for your help 💯

PR #14, opened 2020-03-27
GH-640: Add test for wgs endpoint

samanehsan:
### Purpose
Add an integration test that sends a request to the `wgs` endpoint and checks that the resulting test workflow succeeds: https://broadinstitute.atlassian.net/jira/software/projects/GH/boards/530?selectedIssue=GH-640

### Changes
Add a `wgs_test.clj` file that can be run with the command:
`clojure -A:integration <cromwell-env> <max> <input_path> <output_path>`

Note: `cromwell/wait-for-workflow-complete` gets the cromwell url out of the wfl `env`, which is why I included `resources` in the extra paths instead of just `integration`. There is probably a better way to do this...

To-Do:
- [x] Create temporary output directory and delete it after the test completes, otherwise wfl will see that a result already exists and will not run the test workflow

### Review Instructions

- No instructions.

### DIFF
--- a/integration/zero/wgs_test.clj
+++ b/integration/zero/wgs_test.clj
@@ -1,0 +1,52 @@
+(ns zero.wgs-test
+  (:require [clojure.data.json :as json]
+            [clojure.string :as str]
+            [buddy.sign.jwt :as jwt]
+            [zero.service.cromwell :as cromwell]
+            [zero.environments :as env]
+            [zero.util :as util]))
+
+(def wfl-url
+  "http://localhost:3000")
+
+(def test-input-path
+  "gs://broad-gotc-test-storage/single_sample/plumbing/bams/2m/")
+
+(def test-output-path
+  "gs://broad-gotc-dev-zero-test/wgs-test-output/")
+
+(def credentials-path
+  (get-in (env/stuff :gotc-dev) [:server :vault]))
+
+(defn create-jwt
+  []
+  (let [{:keys [oauth2_client_id oauth2_client_secret]}
+        (util/vault-secrets credentials-path)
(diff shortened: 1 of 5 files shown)


integration/zero/wgs_test.clj
```diff
@@ -0,0 +1,52 @@
...
+                                      :input_path  input-path
+                                      :output_path output-path})}
+      cromwell/request-json :body :results))
+
+(defn -main
+  [& args]
+  (let [workflow-results (start-wgs-workflow "wgs-dev" "1" test-input-path test-output-path)
+        workflow-id (first workflow-results)
+        status (:status (cromwell/wait-for-workflow-complete :wgs-dev workflow-id))]
+    (print (str workflow-id "completed with status: " status))))
```
tbl3rd: Maybe just show the status and workflow ID?
Or use `select-keys` on the metadata?
Does this return a non-0 status when it fails?

integration/zero/wgs_test.clj
```diff
@@ -0,0 +1,53 @@
...
+                                      :output_path output-path})}
+      cromwell/request-json :body :results))
+
+(defn -main
+  [& args]
+  (let [workflow-results (start-wgs-workflow "wgs-dev" "1" test-input-path test-output-path)
+        workflow-id (first workflow-results)
+        status (:status (cromwell/wait-for-workflow-complete :wgs-dev workflow-id))]
+    (println (str workflow-id ": " status))
+    (if (= status "Succeeded") (System/exit 0) (System/exit 1))))
```
tbl3rd: `(System/exit (if (= status "Succeeded") 0 1))` ?

samanehsan:
I made a few more changes @tbl3rd! :octocat:

PR #11, opened 2020-03-23
GH-652: Support random WDL input associations.

tbl3rd:
https://broadinstitute.atlassian.net/browse/GH-652

This goes with https://github.com/broadinstitute/pipeline-config/pull/26 too.

PR #15, opened 2020-04-08
GH-655: WFL-Cromwell communication

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-655

### Changes

- Implement the function to generate `GoogleCredentials` objects and `AccessToken` from service account files, this is also taking the deployment switch from GAE to GKE into consideration. The function depends on a few environment variables defined in the `server.clj`.
- Update ~a few (but not all)~ ALL  of handlers to use the new way of authentication.
- Fix a security alert for the UI.
- Clean up some testing comments in the Helm chart files. 
- **Removed all about OAUTH from the API server. ** (DO NOT deploy this until the login page for the UI is implemented and merged.)
- Update the Helm chart for WFL to include a secret object that stores service account.
- **PR https://github.com/broadinstitute/pipeline-config/pull/25 needs to be merged before this one and we shall update the commit hash in `boot.clj` before merging this one.**

## Open Questions

- [x] This PR does not remove the `authorize` handler wrapper, which only validates the user provided token, decode into JWTs and store them into the ring session's request body. According to @tbl3rd and @hjfbynara we will likely to get the user info(email) from the OIDC proxy in the future, so should we remove this wrapper now or later once we move to GKE? Also, 
is `get the user info(email) from the OIDC proxy` true?
- [x] There are 2 ways of passing the auth tokens:
    - [x] top down: generate the token and passing them from the handlers all the way down to the underlying functions
    - [x] bottom up: generate the token from the underlying functions, basically replace all of the places that are currently using `once/get-local-auth-header`.
  This PR updates a few handlers using the first approach, but it's very likely we need to do a lot with the second approach. Question is it will **break the WFL CLI**, shall we stick with the 1st approach, even it's much more complicated? (Requires refactor a lot of functions to support polymorphism)
- [x] Currently a lot of GCloud service to service communication are done using `once/get-local-auth-header` as well, leveraging the fact that GAE handles some authentication for us. Moving to GKE will change this and we will need to do a lot by ourselves. How could we do it? (The 2nd approach above seems to be more appealing taking this into consideration)

**[Updated on 05/01]: All open questions are resolved now thanks to @tbl3rd and @samanehsan** 

### Review Instructions / QA

- There is a sister PR ready for review as well: https://github.com/broadinstitute/pipeline-config/pull/25
~- Note: this PR won't be **Finished** until we move to GKE completely, so please don't expect this PR to be fully QAed by then!~ We have moved to GKE, which is done in another PR: #23 

1. Make sure you are on the list of `vault read secret/dsde/gotc/dev/wfl/users`, if not, add yourself with `curl --silent -H "authorization:Bearer $(gcloud auth print-access-token)"  -X GET https://www.googleapis.com/oauth2/v1/userinfo\?alt\=json | jq .id`.
2. Make sure the deployed auth proxy has you on the whitelist, you could check by: `curl -X GET "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org/api/workflows/v1/query?status=Running" -H "accept: application/json" -H "authorization:Bearer $(gcloud auth print-access-token)"`, if not authorized, follow the steps while on the VPN/Broad network:
    - `helm repo add gotc-charts https://broadinstitute.github.io/gotc-helm-repo/`
    - `helm list` (sanity check the list)
    - `helm upgrade gotc-dev gotc-charts/authproxy -f custom.yaml`, the custom yaml should be rendered from https://github.com/broadinstitute/gotc-deploy/blob/master/deploy/gotc-dev/helm/cromwell-auth_proxy-values.yaml.ctmpl, with the right `user-id`s in it.
3. with `./ops/server.sh` up and running, run something like `curl -X GET "http://localhost:3000/api/v1/statuscounts?environment=gotc-dev" -H "accept: application/json" -H "authorization:Bearer $(gcloud auth print-access-token)"`

---

Alternatively, after **step 1.**, you could run:
```
curl -X GET "https://workflow-launcher.gotc-dev.broadinstitute.org/api/v1/statuscounts?environment=wgs-dev" -H "accept: application/json" -H "authorization:Bearer $(gcloud auth print-access-token)"
```
and you could see the results returned all the way back from Cromwell to WFL API.


src/zero/api/handlers.clj
```diff
@@ -40,15 +41,12 @@
...
-  Otherwise return a 401 Unauthorized response."
+  "Add JWT to REQUEST all the time. Will be deprecated soon."
   [handler]
   (fn [request]
     (letfn [(valid? [{:keys [payload] :as jwt}]
               (let [{:keys [aud exp hd iss]} payload]
-                (and (= hd "broadinstitute.org")
-                     (= iss "https://accounts.google.com")
-                     (= aud (get-in oauth2-profiles [:google :client-id]))
+                (and (= iss "https://accounts.google.com")
```
tbl3rd, line 49: What is the motivation here?

src/zero/api/handlers.clj
```diff
@@ -59,7 +57,7 @@
                               last)))]
       (if-let [jwt (some-> request token decode-jwt valid?)]
         (handler (assoc request :jwt jwt))
-        (-> (response/response {:response {:message "Unauthorized"}})
+        (-> (response/response {:response {:message "Invalid or expired token provided!"}})
```
tbl3rd, line 60: And here?

src/zero/api/handlers.clj
```diff
@@ -76,20 +74,22 @@
 
 (defn status-counts
   "Get status counts for environment in REQUEST."
-  [{:keys [parameters] :as _request}]
+  [{:keys [parameters jwt] :as _request}]
```
tbl3rd, line 77: Where is `jwt` used?

src/zero/once.clj
```diff
@@ -31,3 +34,37 @@
...
+    (let [scopes ["email" "profile" "openid"]
+          credentials (GoogleCredentials/fromStream (FileInputStream. file))]
+      (.createScoped credentials scopes))))
+
+(defn service-account-per-env
+  "Map of environment to service account path."
+  []
+  (letfn [(get-sa [environment] {(keyword environment)
+                                 (get-in env/stuff [(keyword environment) :project-service-account])})]
+    (into {} (map get-sa env/registered-envs))))
```
tbl3rd, line 51: How do I build with this change?
I'm guessing this will be simpler after merging PR#11.
https://github.com/broadinstitute/wfl/pull/11

src/zero/once.clj
```diff
@@ -31,3 +34,37 @@
...
   []
   (util/bearer-token-header-for @the-cached-credentials))
+
+(defn new-credentials-from-service-account
+  "Generate scoped GoogleCredentials from a service account FILE."
+  [^String file]
+  (when-not (nil? file)
+    (let [scopes ["email" "profile" "openid"]
+          credentials (GoogleCredentials/fromStream (FileInputStream. file))]
+      (.createScoped credentials scopes))))
```
tbl3rd, line 44: This is just `(when file …)`

src/zero/util.clj
```diff
@@ -65,13 +65,8 @@ vault.client.http/http-client           ; Keep :clint eastwood quiet.
...
                       .getTokenValue)]
-    (when-not token
-      (let [lines ["%1$s: Cannot generate token from Google Credentials."
-                   "%1$s: Run 'gsutil auth list' to check your account."
-                   "%1$s: Run 'gsutil auth login' and try again."]
-            err (format (str/join \newline lines) zero/the-name)]
-        (throw (Exception. err))))
-    {"Authorization" (str "Bearer " token)}))
+    (when token
+      {"Authorization" (str "Bearer " token)})))
```
tbl3rd, line 69: Again. I don't understand the motivation for this.

src/zero/once.clj
rexwangcc: 😅 you are right...

src/zero/once.clj
rexwangcc: Yes! #11 simplifies this a lot, thanks!

rexwangcc:
> We should talk about your Open Questions.
So this PR is not really easy to be tested until we have both the Cromwell Auth Proxy and WFL Proxy set up. The reason why you see this PR removes a lot of stuff is that with the WFL auth proxy in hand, we won't need to store the JWT anywhere anymore. The only approach that might still be used to validates the token is to make sure if it's issued by Google OAuth. We may remove it as well if Proxy handles all of the parts and throw WFL a simple email. Maybe we could talk through the open questions.

tbl3rd:
I would like to approve this, but I cannot build or load it for review.

rexwangcc:
> I would like to approve this, but I cannot build or load it for review.

Thanks, sorry forgot to mention this PR was stale and I'm still on the way finishing it up.

helm/wfl/values.yaml
```diff
@@ -104,6 +110,7 @@ securityContext: {}
 
 ingress:
   enabled: false
-  servicePort: http
+  servicePort: wfl-ui
```
tbl3rd, line 120: Where is wfl-ui defined?  Never mind.  I see it in service.yaml.

src/zero/api/routes.clj
```diff
@@ -155,22 +152,22 @@
...
+            :handler    handlers/post-create}}]
    ["/api/v1/start"
     {:post {:summary    "Start workloads."
             :parameters {:body ::uuid-kvs}
             :responses  {200 {:body ::workload-responses}}
-            :handler    (handlers/authorize handlers/post-start)}}]
+            :handler    handlers/post-start}}]
    ["/swagger/swagger.json"
     {:get {:no-doc true ;; exclude this endpoint itself from swagger
            :swagger {:info {:title (str zero/the-name "-API")
```
tbl3rd, line 173: Nice.

src/zero/environments.clj
```diff
@@ -68,6 +68,10 @@
                             (->> (assoc m kw)))))]
     (reduce make {} ["development" "production"])))
 
+(def registered-envs
+  "Register a list of the available ENVIRONMENT names here."
+  ["development", "production"])
+
```
tbl3rd, line 74: I don't understand registered-envs.

src/zero/util.clj
```diff
@@ -68,8 +68,8 @@ vault.client.http/http-client           ; Keep :clint eastwood quiet.
                       .getTokenValue)]
     (when-not token
       (let [lines ["%1$s: Cannot generate token from Google Credentials."
-                   "%1$s: Run 'gsutil auth list' to check your account."
-                   "%1$s: Run 'gsutil auth login' and try again."]
+                   "%1$s: Run 'gsutil auth list' to check your account if running locally"
+                   "%1$s: Try to login if running on a server."]
```
tbl3rd, line 72: How will you see this on a server?

src/zero/server.clj
```diff
@@ -44,13 +42,11 @@
   [env]
   (let [environment (env env/stuff)
         {:keys [cookie_secret
-                oauth2_client_id oauth2_client_secret
                 password username]}
         (util/vault-secrets (get-in environment [:server :vault]))
-        result {"COOKIE_SECRET"          cookie_secret
+        result {"WFL_LIVE_SERVER_MODE"   "true"
+                "COOKIE_SECRET"          cookie_secret
```
tbl3rd, line 48: Do we still need a cookie secret?

.github/workflows/helmchart-release.yml
```diff
@@ -1,16 +1,12 @@
 name: Helm Charts Release
 
+# NOTE: this workflow does NOT work for PR commits!
```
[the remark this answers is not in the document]
rexwangcc: Ah, so we used to add:
```
pull_request:
```
to Github Actions and push changes to a PR to test if they work, so we don't have to merge them to master and then fix them if there are any issues. However this particular one does not work that way, the commits to PR will trigger the action but it does not detect any changes to the Helm:
```
Discovering changed charts ...
Nothing to do. No chart changes detected.
```
thus won't release a new version for us, see: https://github.com/broadinstitute/wfl/runs/627504235?check_suite_focus=true

This comment is to remind us we have to merge PR to master if we want to make any changes to the Action later.

helm/wfl/Chart.yaml
```diff
@@ -14,11 +14,10 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
```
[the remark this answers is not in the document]
rexwangcc: Yeah, they are auto-generated by Helm 😅 I could remove them. Agree that they don't add any values.

helm/wfl/templates/NOTES.txt
```diff
@@ -1,21 +1,8 @@
-1. Get the application URL by running these commands:
+1. Get the Workflow Launcher URL by running these commands:
 {{- if .Values.ingress.enabled }}
 {{- range $host := .Values.ingress.hosts }}
-  {{- range .paths }}
-  http{{ if $.Values.ingress.tls }}s{{ end }}://{{ $host.host }}{{ . }}
-  {{- end }}
+  http{{ if $.Values.ingress.tls }}s{{ end }}://{{ $host.name }}{{ . }}
```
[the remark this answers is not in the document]
rexwangcc: `NOTES.txt ` is just a block of dynamically rendered helper messages that will show up when someone run `helm install` or `helm upgrade`. Taking a closer look I also think we will probably never have ingress enabled w/o TLS. Will remove the condition for less confusion...

helm/wfl/templates/deployment.yaml
```diff
@@ -71,15 +70,9 @@ spec:
...
-    {{- with .Values.tolerations }}
-      tolerations:
-        {{- toYaml . | nindent 8 }}
-    {{- end }}
+        - name: serviceaccount
+          secret:
+            secretName: {{ template "wfl.fullname" . }}
+            items:
+              - key: service-account
+                path: service-account.json
```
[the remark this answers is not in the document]
rexwangcc: I guess they don't have to. `serviceaccount` was used here when I figured we had too many same keys in the yaml and thus will bring confusion, now I think I have a  2nd thought, especially serviceaccount looks pretty awkward as one word... Will change it

helm/wfl/templates/secrets.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "wfl.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+  {{- include "wfl.labels" . | nindent 4 }}
+stringData:
+  service-account: {{ .Values.api.secrets.secretsgeneric.serivceAccountCredentialFileJson | toJson }}
+{{- end }}
```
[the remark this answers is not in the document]
rexwangcc: Yeah, as https://helm.sh/docs/chart_best_practices/templates/#formatting-templates stated, `-`s are used to control subtle blank lines/whitespaces during the rendering of the templates. (_a beautiful templated yaml language :)_)

helm/wfl/Chart.yaml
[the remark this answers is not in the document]
tbl3rd: Leave them in if removing them would surprise people.

src/zero/util.clj
rexwangcc: 🤔 Good question... You won't ... we need to make the error handling better and be able to pass the messages all the way from API to UI.

src/zero/environments.clj
rexwangcc: It's used in `once.clj` module, for creating a map between environments and their service account paths. But since now we have made the paths to service accounts identical to each other in the environments.clj, not sure if we really need this anymore...

.github/workflows/helmchart-release.yml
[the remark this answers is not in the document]
samanehsan: I think the Github action only releases helm charts when the chart.yml changes

src/zero/boot.clj
```diff
@@ -112,10 +112,10 @@
             (let [out (io/file resources (.getName in))]
               (io/make-parents out)
               (io/copy in out)))]
-    (let [config (clone "pipeline-config" "zero/environments.clj")]
+    (let [config (clone "pipeline-config" "wfl/environments.clj")]
       (stage (clone "dsde-pipelines" "tasks/CopyFilesFromCloudToCloud.wdl"))
       (util/shell-io! "git" "-C" (.getParent config)
-                      "checkout" "e44decd39db39725f79bddc42d944181d0621b1d")
+                      "checkout" "a51348d8f755a2c92aad84cc94a9235d5cc90450")
```
samanehsan, line 118: Reminder to update this commit hash, now that the pipeline-config PR is merged :octocat:

src/zero/environments.clj
samanehsan: I'm also wondering this 🤔

src/zero/boot.clj
rexwangcc: Thanks!!

src/zero/environments.clj
rexwangcc: Per the video chat, that's purely me attempting to standardizing the envs and distinguishing depoy-able envs from project envs. We probably don't need this anymore. I will create a PR to pipeline-config to update this.

src/zero/once.clj
```diff
@@ -26,8 +30,48 @@
...
+(defn get-auth-header!
+  "Return a valid auth header. Refresh and generate the access
+   token with gcloud command if invoked from command line,
+   generate the access token from service account if invoked
+   from a live server."
   []
-  (util/bearer-token-header-for @the-cached-credentials))
+  (if (and (System/getenv "WFL_LIVE_SERVER_MODE") (System/getenv "ENVIRONMENT"))
+    (util/bearer-token-header-for @the-cached-credentials-from-service-account)
+    (util/bearer-token-header-for @the-cached-credentials)))
```
tbl3rd, line 61: The `and` is a little weird here.  What does `"ENVIRONMENT"` add?

rexwangcc: It does not add anything, I was thinking making the conditions for `@the-cached-credentials-from-service-account` explicit (both `WFL_LIVE_SERVER_MODE ` and `ENVIRONMENT ` have to be set properly) was a good thing 🤔

src/zero/server.clj
rexwangcc: I believe so, the `wrap-defaults` requires `cookie-store` which requires cookie secret

PR #19, opened 2020-04-14
[GH-707] Add DNS terraform config

samanehsan:
### Purpose
https://broadinstitute.atlassian.net/jira/software/projects/GH/boards/530?selectedIssue=GH-707

### Changes
Add terraform config to create a public IP address and DNS record for the WFL UI.

Note: This uses the same shared terraform state file as the auth proxies for the WFL & gotc-cromwell

### Review Instructions

Thoughts on the DNS name `workflow-launcher`? The one for the API is called `wfl-auth` but that can change as well.

### DIFF
--- a/deploy/gotc-dev/terraform/broad-gotc-dev-provider.tf
+++ b/deploy/gotc-dev/terraform/broad-gotc-dev-provider.tf
@@ -1,0 +1,18 @@
+# Configure the Google Cloud provider
+# Every google project that resources are going ot be created in will have its own
+#  google provider definition using the google project name as its alias so as to be
+#  clear
+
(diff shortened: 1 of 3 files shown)


deploy/gotc-dev/terraform/broad-gotc-dev-provider.tf
```diff
@@ -0,0 +1,18 @@
+# Configure the Google Cloud provider
+# Every google project that resources are going ot be created in will have its own
```
rexwangcc, line 2:
```suggestion
# Every google project that resources are going to be created in will have its own
```

PR #24, opened 2020-04-24
A draft chart releaser for WFL

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/jira/software/projects/GH/boards/530?selectedIssue=GH-708

### Changes

- A draft chart releaser borrowed directly from gotc-deploy.

### Review Instructions

- Note: the `CR_TOKEN` is not set yet. We have to discuss what token we wanna use for it.
- Note: Helm has this conception of `--merge` which will help avoid overriding existing chart index file in the chart repo. I'm not sure the chart-releaser action we use here and in the gotc-deploy repo can do the same thing.


Fab-T:
So far so good, you just need to create your first helm chart and store it in helm folder.
CR will never override an existing release and fail with Error : already exists.
I believe the GH Chart tester should check when doing a PR if you are updating a chart but not bumped the version so it doesn't fail with this error at merge.

PR #26, opened 2020-04-26
Ra master docker build n publish

rhiananthony:
### Purpose
This PR is to create a gitaction that will build and publish the wdfl api and wfl ui docker images

Since deployment keys are unique, and we required all three repos involved to have a deployment key, a change had to be made to the clojure code which generates the URLs that are used to clone repos. This is now wrapped in a try/catch statement, which will first try the aliased URLs and then try the non-aliased URLs. We could have alternatively altered the code to allow the wfl repo to be cloned via the https URL and then use the SSH URL for the remaining repos, but that was a more involved change than this and so was considered less of an ideal solution.

### Review Instructions
Verify that the workflow succeeds and a new image is pushed to the location in dockerhub:
broadinstitute/workflow-launcher-api
broadinstitute/workflow-launcher-ui

### DIFF
--- a/.github/workflows/docker_build_and_publish.yml
+++ b/.github/workflows/docker_build_and_publish.yml
@@ -44,8 +44,12 @@ jobs:
       run: |
         echo "${{ secrets.dsde_pipelines_deploy_key }}" > ~/.ssh/dsde_pipelines_deploy_key
         echo "${{ secrets.pipeline_config_deploy_key }}" > ~/.ssh/pipeline_config_deploy_key
+        echo "${{ secrets.wfl_deploy_key }}" > ~/.ssh/wfl_deploy_key
+
         chmod 600 ~/.ssh/dsde_pipelines_deploy_key
         chmod 600 ~/.ssh/pipeline_config_deploy_key
+        chmod 600 ~/.ssh/wfl_deploy_key
+
         echo """
           Host dsde-pipelines.github.com
           HostName github.com
@@ -60,7 +64,8 @@ jobs:
           Host wfl.github.com
           HostName github.com
           User git
-          IdentityFile ~/.ssh/wfl_deployment_key""" > ~/.ssh/config
+          IdentityFile ~/.ssh/wfl_deploy_key""" > ~/.ssh/config
+        du -h ~/.ssh/dsde_pipelines_deploy_key
         cat ~/.ssh/config
         boot build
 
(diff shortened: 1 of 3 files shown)


src/zero/boot.clj
```diff
@@ -76,11 +76,19 @@
...
                               "--config" "advice.detachedHead=false" url))]
       (io/make-parents (io/file tmp "Who cares, really?"))
-      (run! clone (vals zero/the-github-repos)))
+
+      (try
+        (run! clone (vals zero/the-other-github-repos))
+        (catch Exception e
+          (run! clone (vals zero/the-github-repos))
+          )))
+
```
tbl3rd, line 85: Do you need both of these?

src/zero/zero.clj
```diff
@@ -17,7 +17,13 @@
 (def the-github-repos
   "Map Zero source repo names to their URLs"
   (let [repos ["wfl" "dsde-pipelines" "pipeline-config"]
-        git   (partial str "https://github.com/broadinstitute/")]
+        git   (partial str "git@github.com:broadinstitute/")]
+    (zipmap repos (map git repos))))
+
```
tbl3rd, line 22: Factor out the `fn` applied and these are the same.

.github/workflows/docker_build_and_publish.yml
```diff
@@ -0,0 +1,92 @@
...
+      id: get_branch
+      shell: bash
+      run: echo "##[set-output name=BRANCH_NAME;]$(echo ${GITHUB_HEAD_REF#refs/heads/})"
+
+    - name: Install Clojure tools
+      uses: DeLaGuardo/setup-clojure@2.0
+      with:
+        leiningen: 2.9.1
+        boot: 2.8.3
+        tools-deps: 1.10.1.469
```
tbl3rd, line 45: Do we need leiningen and tools-deps?

src/zero/zero.clj
rhiananthony: This is true, but not understanding clojure very well I am not quite sure how to combine these two.

.github/workflows/docker_build_and_publish.yml
rhiananthony: I'm not sure. I can look.

rhiananthony: It appears that we do not need those - I removed them both.

PR #28, opened 2020-05-01
Some ui fixes during hornet engineering time

rexwangcc:
### Purpose

### Changes

- Fix Dashboard.
- Fix several places that still have "Zero".
- Remove redundant code and Google Auth button.
- Update the API `/version` endpoint to separate pipeline-versions and system versions.

### Review Instructions

- No instructions.

### DIFF
--- a/src/zero/api/routes.clj
+++ b/src/zero/api/routes.clj
@@ -1,6 +1,7 @@
 (ns zero.api.routes
   "Define routes for API endpoints"
   (:require [clojure.spec.alpha                 :as s]
+            [clojure.string                     :as str]
             [muuntaja.core                      :as muuntaja-core]
             [reitit.coercion.spec]
             [reitit.ring                        :as ring]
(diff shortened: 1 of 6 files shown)


ui/src/router/index.js
```diff
@@ -16,23 +16,23 @@ const routes = [
     // route level code-splitting
     // this generates a separate chunk (about.[hash].js) for this route
```
samanehsan, line 17: Oh there's another reference to "about". Can you update it to say (query.[hash].js)?

rexwangcc: Oh, good catch!!! Thanks, I will address it right now.

src/zero/api/routes.clj
```diff
@@ -121,10 +122,14 @@
...
-                                    :build    pos-int?
-                                    :built    string?}}}
+            :handler (handlers/success (let [versions (zero/get-the-version)
+                                             pipeline-versions-keys (keep (fn [x] (when (and (string? x) 
+                                                                                             (str/ends-with? x ".wdl")) x)) 
+                                                                          (keys versions))]
+                                         {:pipeline-versions (select-keys versions pipeline-versions-keys)
+                                          :version (apply dissoc versions pipeline-versions-keys)}))
+            :responses {200 {:body {:version map?
+                                    :pipeline-versions map?}}}
```
tbl3rd, line 132: Seems like the wrong place to separate these.
Either do it when constructing the version itself,
or do it in the UI, which is the code that cares.
Also, map? is too loose.
We want /version to at least return a :version.

rexwangcc: > map? is too loose.

That makes sense, will fix it in a follow up PR.

> Either do it when constructing the version itself,
or do it in the UI, which is the code that cares.

The problem I had with doing this in the frontend is that we have to use specific getters like: https://github.com/broadinstitute/wfl/pull/28/files#diff-88ce83892fa39769e0067bd6efa72316L35 which will require updates every time we add or update a pipeline (name). If that's fine with you I could revert it!

PR #29, opened 2020-05-04
[GH-742] Update login UI

samanehsan:
### Purpose
Addresses: https://broadinstitute.atlassian.net/browse/GH-742

### Changes
Users who are not logged in are redirected to this login page:
Screen Shot 2020-04-30 at 4 17 01 PM

After signing in, users are redirected to the dashboard page. They will also be able to see the icon to toggle the sidebar and a logout button in the navigation bar:
Screen Shot 2020-05-04 at 11 24 03 AM

### Review Instructions
- Note: The API requests that require authentication will return 401 errors until the change to the WFL server goes in.

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -20,6 +20,7 @@
 </template>
 
 <script>
+import axios from "axios";
 import TopBar from "./components/TopBar.vue";
 import SideBar from "./components/SideBar.vue";
 
@@ -35,12 +36,24 @@ export default {
     };
   },
   created: function() {
+    const store = this.$store
     window.gapi.load('auth2', initAuth);
     function initAuth() {
       window.gapi.auth2.init({
         client_id: '450819267403-n17keaafi8u1udtopauapv0ntjklmgrs.apps.googleusercontent.com'
       });
     }
+    axios.interceptors.response.use(null, (error) => {
+      const unauthorized = (error.response && error.response.status === 401)
+      if(unauthorized && error.config && !error.config.__isRetryRequest) {
+        const user = window.gapi.auth2.getAuthInstance().currentUser.get()
+        store.dispatch('auth/refreshToken', user).then(() => {
+          error.config.__isRetryRequest = true
+          return axios.request(error.config);
+        })
+       }
+       return Promise.reject(error);
+    });
   }
 };
 </script>
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -20,7 +20,6 @@
 </template>
 
 <script>
-import axios from "axios";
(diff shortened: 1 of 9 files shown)


samanehsan:
I realized I also need to check that the user's token is not expired. I think axios provides a pre-request hook so we could add the logic there.

ui/src/App.vue
```diff
@@ -34,11 +35,15 @@ export default {
       modules: [{ text: "loading..." }]
     };
   },
-
   created: function() {
-  },
-
-  mounted: function() {}
+    window.gapi.load('auth2', initAuth);
+    function initAuth() {
```
rexwangcc, line 39: Since this is the main App, maybe we could consider making the function `async initAuth()` so it won't block other loads, as an optimization later?

ui/src/components/TopBar.vue
```diff
@@ -5,9 +5,9 @@
         <v-row justify="space-around" align="center">
           <v-col cols="5">
             <v-toolbar-title>
-              <v-app-bar-nav-icon class="mr-2" @click.stop="toggleSideBar"></v-app-bar-nav-icon>
+              <v-app-bar-nav-icon v-if="isAuthenticated" class="mr-2" @click.stop="toggleSideBar"></v-app-bar-nav-icon>
```
rexwangcc, line 8: Like this condition!

ui/src/components/TopBar.vue
```diff
@@ -53,7 +54,23 @@ export default {
       swagger_link: "/swagger/index.html"
```
rexwangcc, line 54: This was removed in another PR, might need to rebase, sorry

ui/src/components/TopBar.vue
```diff
@@ -53,7 +54,23 @@ export default {
...
+      this.$router.push('/login');
+    }
+  },
   methods: {
+    logout() {
+      window.gapi.auth2.getAuthInstance().signOut().then(user => {
+        this.$store.dispatch('auth/updateUser', user);
+        sessionStorage.clear();
+      });
+    },
```
rexwangcc, line 72: I haven't tested this locally, but I assume this will also redirect the user back to the login page?

ui/src/router/index.js
```diff
@@ -41,4 +51,17 @@ const router = new VueRouter({
   routes
 })
 
+
+router.beforeEach((to, from, next) => {
+    const isPublic = to.matched.some(record => record.meta.public);
+    const loggedIn = store.getters['auth/authenticated'];
+
+    if (!isPublic && !loggedIn) {
+        return next({ name: 'login' })
```
rexwangcc, line 60: NVM, I see here!

PR #30, opened 2020-05-06
GH-716: test that WFL runs on k8s and on the auth proxy

tbl3rd:
https://broadinstitute.atlassian.net/browse/GH-716

There are also these.
https://github.com/broadinstitute/pipeline-config/pull/29
https://github.com/broadinstitute/gotc-deploy/pull/47


integration/zero/create.clj
```diff
@@ -0,0 +1,63 @@
+(ns zero.create
```
ehigham, line 1: These integration tests (create, start, etc) have been moved to `integration/zero/v1_endpoint_test.clj`

helm/wfl/values.yaml
```diff
@@ -17,16 +17,16 @@ api:
   deployment:
     containerPort: 80
   env:
-    ENVIRONMENT: "wgs-dev"
```
rexwangcc, line 20: Changes to helm charts need a version bump to https://github.com/broadinstitute/wfl/blob/master/helm/wfl/Chart.yaml, otherwise the lint test will fail (failing now)

src/zero/module/copyfile.clj
```diff
@@ -5,7 +5,7 @@
             [zero.module.all :as all]
```
ehigham, line 5: No functional changes to this file. Revert?

tbl3rd: Not now, please.

src/zero/boot.clj
```diff
@@ -163,59 +153,7 @@
            (run! (partial cromwellify-wdl tmp directory) wdls)
            (adapterize-wgs directory)
            (write-the-version-file directory edn)
-           (finally (util/delete-tree (io/file tmp))))
-      (spit "./build.txt" (-> version :build inc (str \newline))))))
-
-(defn google-app-engine-configure
```
rexwangcc, line 169: ❤️

src/zero/once.clj
```diff
@@ -1,77 +1,44 @@
...
+(defn authorization-header-with-bearer-token
+  "An Authorization header with a Bearer TOKEN."
+  [token]
+  {"Authorization" (str/join \space ["Bearer" token])})
+
+(defn service-account-token
+  "Throw or return a bearer token for the service account SA."
+  [{:keys [file vault] :as sa}]
+  (-> (cond file  (io/file file)
+            vault (-> vault util/vault-secrets json/write-str .getBytes)
```
rexwangcc, line 27: If I understand this correctly, in the env file, only one of `file` and `vault` should be set to true when we create new envs?

tbl3rd: Maybe neither. It depends, but it will prefer :file for no good reason.

integration/zero/create.clj
tbl3rd: Rebases kept resurrecting them. I believe they are gone now.

src/zero/once.clj
tbl3rd: I don't like this solution much. We can discuss what else I tried and maybe find something better.

src/zero/once.clj
```diff
@@ -1,77 +1,61 @@
...
+  "Throw or return credentials for the service account SA."
+  [{:keys [file vault] :as sa}]
+  (letfn [(jsonify [edn] (json/write-str edn :escape-slash false))]
+    (-> (cond file  (io/file file)
+              vault (-> vault util/vault-secrets jsonify .getBytes)
+              :else (throw (IllegalArgumentException. (pr-str sa))))
+      io/input-stream GoogleCredentials/fromStream
+      (.createScoped ["https://www.googleapis.com/auth/cloud-platform"
+                      "https://www.googleapis.com/auth/userinfo.email"
+                      "https://www.googleapis.com/auth/userinfo.profile"]))))
```
rexwangcc, line 33: just a curious question, where are these coming from? Previously we were using the OAuth2 scopes: https://developers.google.com/identity/protocols/oauth2/scopes#oauth2 which includes `openid`

tbl3rd: I saw an error message naming those scopes. Not driven by any advance design.
Maybe openid encompases email and enough of profile to do the trick?
I'll try it.

rexwangcc: Sounds good, it seems the scopes changed to use `userinfo.email` from `email`

PR #36, opened 2020-05-11
Gh-777 update helmchart workflow wfl

Fab-T:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-777

### Changes

- Updated the workflow for chart release to read token from vault
- Added workflow chart test
- Bumped version of wfl helm chart to test

### Review Instructions

- the reviewer should confirm the PR triggered the workflow chart test and it's ok
- the reviewer should confirm the same changes applied to repo gotc-deploy for workflow chart releaser have been applied in repo wfl
- secrets should contain roleID, secretID, vault_addr
- secret CR_TOKEN should be removed

### DIFF
--- a/.github/workflows/helmchart-release.yml
+++ b/.github/workflows/helmchart-release.yml
@@ -1,6 +1,5 @@
 name: Helm Charts Release
 
-# NOTE: this workflow does NOT work for PR commits!
 on:
   push:
     branches:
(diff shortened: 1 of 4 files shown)


Fab-T:
> LGTM! As i see more vault code snippets in our charts, we may want to package it into a standalone github action.

I would like to put these actions into one place and inherit them when updated.
Right now we are using helmchart-release, helmchart-test in 2 repos and potentially in others and for every change I need to update those repos.
- one idea is creating a new action for getsecret from Vault
 1. pro :  quick
2. cons : it factorize only one portion of the action
- another would be to create a new image so the get one step in the action
1. pro : factorize more
2. cons : longer to implement

PR #37, opened 2020-05-18
A Draft temp script for our deployment and testing.

rexwangcc:
### Purpose

- A Draft temp script for our deployment and testing.

### Changes

- Remove the old Bash script and use Python so it's more modular and extensible.
- Scope out the `local deployment` and focus on `cloud deployment`.
- Let the Github Action also push images with `latest` tag.
- Fix an issue with the ctmpl values file: https://github.com/broadinstitute/gotc-deploy/pull/51

### Review Instructions

- No instructions.


ops/cli.sh
```diff
@@ -0,0 +1,163 @@
...
+
+    docker run -i \
+               --rm \
+               -v "$(pwd)":/working \
+               -v "${HOME}"/.vault-token:/root/.vault-token \
+               broadinstitute/dsde-toolbox:dev \
+               /usr/local/bin/render-ctmpls.sh \
+               -k "${CTMPL_FILE}"
+}
+
```
tbl3rd, line 63: Did you intend `CHART_REPO_ALIAS` to be local? (It isn't used here.)
Or do you expect its value to show up in `deploy_helm_charts` too?

rexwangcc: Fixed

rexwangcc:
`./ops/cli.sh` is close to work, once:
- We figure out a way to let it pull and render the ctmpl file.
- We figure out how to turn off the Apache's rewrite rule for local testing
It will be all set

ops/cli.sh
```diff
@@ -0,0 +1,166 @@
...
+#
+function run_local () {
+    local -r port=18982
+    local -ar getpods=(kubectl get pods --no-headers
+                       -o custom-columns=:metadata.name)
+    line
+    # minikube start is not idempotent, kill
+    # all previous ones for the sake of your
+    # laptop's safety
+    minikube stop
```
ehigham, line 129: I think this should be the responsibility of whoever is running the test. Stopping minikube globally might be annoying if people are working on multiple things simultaneously.
Can we use setup/teardown fixture semantics for this to isolate this deployment from others?

ops/cli.sh
```diff
@@ -0,0 +1,166 @@
...
+    render_helm_values "wfl-values.yaml.ctmpl"
+    line
+    setup_and_update_helm_charts gotc-charts
+    line
+    deploy_helm_charts gotc-dev gotc-charts wfl wfl-values.yaml
+    line
+    local -r pod=$(wait_for_result "${getpods[@]}")
+    success Deployed "$pod" to a local minikube cluster.
+    success Browse http://localhost:$port to view it.
+    sudo kubectl port-forward "$pod" $port:80
```
ehigham, line 148: is there a way of doing this without sudo permissions?

rexwangcc: Unfortunately I haven't found a good way to do it on macOS. Maybe we could print a message and handoff running this command to the developer not the script?

ops/cli.py
```diff
@@ -0,0 +1,249 @@
+"""WFL Deployment Script
```
ehigham, line 1: To facilitate ease of use, perhaps include a python3 shebang so that we can just call it:
```
./ops/cli.py --help
```

rexwangcc: Thanks! Updated :)

samanehsan:
@rexwangcc where is the ctmpl file for the Helm chart values? 🤔

ops/cli.py
```diff
@@ -0,0 +1,249 @@
...
+            CLI._clone_config_repo(
+                dest=Path(f"./{dir_name}/gotc-deploy"), branch="master"
+            )
+
+            file_name = "wfl-values.yaml.ctmpl"
+            shutil.copy(
+                str(Path(f"./{dir_name}/gotc-deploy/deploy/gotc-dev/helm/{file_name}")),
+                Path(f"./{dir_name}/{file_name}"),
+            )
+            CLI._render_ctmpl(ctmpl_file=str(Path(f"./{dir_name}/{file_name}")))
```
samanehsan, line 242: Why is the ctmpl file rendered in two places (the render command and in the deploy command)? Is the render command meant to be a way to review the values?

rexwangcc: I was thinking the `render` command could work as a utility command so we could use it in other situations, e.g. render the auth-proxy CTMPL files.

samanehsan:
If we need to do something like this again in the future, I think the [invoke library](http://www.pyinvoke.org/) would be useful since it simplifies calling command-line tasks with python! :octocat:

rexwangcc:
> If we need to do something like this again in the future, I think the [invoke library](http://www.pyinvoke.org/) would be useful since it simplifies calling command-line tasks with python! :octocat:

Thanks, that looks like a cool lib!!

PR #38, opened 2020-05-20
Create new temp directory for WDL hackery

tlangs:
### Purpose
WFL should not attempt to delete _ALL_ files in a users `java.io.tmpdir` directory. This directory is shared with many other applications, and wiping out all files here can be detrimental for users. If the user has a lot of tmp files, then WFL will appear to hang while `util/delete-tree` recursively deletes every file. Additionally, it can create problems for other applications that put their tmp files in this directory.

### Changes
Made WFL create a new temporary directory for its WDL hackery instead of using a shared tmp directory and deleting everything in it.

### Review Instructions

- No instructions.

### DIFF
--- a/src/zero/wdl.clj
+++ b/src/zero/wdl.clj
@@ -5,6 +5,8 @@
             [zero.util :as util]
             [zero.zero :as zero])
   (:import [java.io File FileNotFoundException]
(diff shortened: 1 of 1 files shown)


tlangs:
There's no line there

PR #42, opened 2020-05-27
use one test runner for wfl

ehigham:
### Purpose
Use the same test runner for unittests and integration tests. There's no need to use two separate ones.

### Changes
1. Replace test runner in deps.edn
2. Add `unittest` identifier to tests.edn

Unit tests can still be run in the usual way:
````
clojure -A:test
````

### Open Questions
Should the integration and unit tests be combined under a common alias for convenience? I wonder if something like this is more useful:
````
clojure -A:test             # run all tests in wfl
clojure -A:unittest       # run all unit tests 
clojure -A:integration  # run all integration tests 
````

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -75,11 +75,9 @@
   {:main-opts   ["-m" "zero.main"]}
 
   :test
-  {:extra-deps  {com.cognitect/test-runner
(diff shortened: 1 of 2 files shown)


ehigham:
Dropping the umbrella test suggestion for now. Can review later for automation.

PR #46, opened 2020-06-03
Add a condition for WFL local development.

rexwangcc:
### Purpose

- Add a condition for WFL UI local development according to http://httpd.apache.org/docs/2.4/mod/core.html#if

### Changes

- We could turn off HTTPS redirect for local development.

### Review Instructions

- No instructions.

### DIFF
--- a/helm/wfl/templates/configmap.yaml
+++ b/helm/wfl/templates/configmap.yaml
@@ -6,6 +6,165 @@ metadata:
   labels:
     {{- include "wfl.labels" . | nindent 4 }}
 data:
+  override: |-
+    #!/bin/sh
+
+    # I'm an override file.  Put things in here to override the openidc-baseimage
+    # defaults or add more settings of your own.
+    #
+    # Note: This script will be executed *after* the main `run.sh` script, but
+    # *before* Apache itself is run.
+
+    # set ALLOW_HEADERS3
+    if [ -z "$ALLOW_HEADERS3" ] ; then
+        export ALLOW_HEADERS3=
+    fi
+
+    # update ALLOW_METHODS3
+    if [ -z "$ALLOW_METHODS3" ] ; then
+        export ALLOW_METHODS3=
+    fi
+
+    # update AUTH_REQUIRE3
+    if [ -z "$AUTH_REQUIRE3" ] ; then
+        export AUTH_REQUIRE3='Require valid-user'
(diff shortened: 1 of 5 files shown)


samanehsan:
Is this for running WFL locally using minikube?

rexwangcc:
> Is this for running WFL locally using minikube?

Yup! Currently without having the ability to turn off HTTPS locally, it's hard to test the requests to API.

rexwangcc:
@tbl3rd It's still not working yet, it seems only turning the redirect is not enough, we also have to turn the liveness probe and readiness probe off :(

ops/cli.sh
```diff
@@ -0,0 +1,179 @@
...
+        fi
+    done
+    test "$ok"
+}
+
+# Wait at least 10 seconds for a result from "@".
+#
+wait_for_result () {
+    local n result
+    for n in 0 1 2 3 4 5 6 7 8 9
```
ehigham, line 61: `for n in {0..9}`

helm/wfl/templates/configmap.yaml
```diff
@@ -6,6 +6,165 @@ metadata:
   labels:
     {{- include "wfl.labels" . | nindent 4 }}
 data:
+  override: |-
```
ehigham, line 9: Is there no other way of doing this? Like a .phony "build" target?

helm/wfl/templates/configmap.yaml
```diff
@@ -6,6 +6,165 @@ metadata:
...
+    #!/bin/sh
+
+    # I'm an override file.  Put things in here to override the openidc-baseimage
+    # defaults or add more settings of your own.
+    #
+    # Note: This script will be executed *after* the main `run.sh` script, but
+    # *before* Apache itself is run.
+
+    # set ALLOW_HEADERS3
```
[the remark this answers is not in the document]
rexwangcc: Thanks for the comment! This is copied from https://github.com/broadinstitute/openidc-proxy/blob/master/override.sh, I'm still not sure how to make Apache respect the Env Vars yet :(

rexwangcc:
Based on an offline discussion with @tbl3rd , it doesn't make sense to stick with local development. Closing this now!

helm/wfl/templates/configmap.yaml
tbl3rd: Ah … make.

PR #47, opened 2020-06-04
Create a dummy testing module

ehigham:
### Purpose
The purpose of this change is to facilitate integration testing.
Currently, we lack a nice way of testing Cromwell integration, i.e. auth and workload submission/verification.
To alleviate problem, we're introducing a dummy pipeline called `copyfile` that simply copies its input file to the output because:
- It's computationally inexpensive to run (-- testing time)
- It doesn't require real data to run
- We can test cromwell authentication quickly
- We can test if the thing ran successfully very simply.

### Changes
714ea6 - Create copyfiole wdl
3d0387 - Add database migration scheme for new pipeline
9c1106 - Slight deviation, but try to organise our test root more effectively (3rd attempt??)
fe9f385 - Preliminary implementation of the module 

It's not quite working yet and there'll be produce code updates coming. Test code is behaving as expected.

### Review Instructions

- No instructions.


database/changelog.xml
```diff
@@ -14,4 +14,6 @@
     <include file="changesets/20200316_ExternalWholeGenomeReprocessingEnum.xml"  relativeToChangelogFile="true"/>
     <include file="changesets/20200317_ExternalWholeGenomeReprocessingTable.xml" relativeToChangelogFile="true"/>
     <include file="changesets/20200424_RenameLoadColumnToItems.xml"              relativeToChangelogFile="true"/>
+    <include file="changesets/20200600_CopyfileEnum.xml"                         relativeToChangelogFile="true"/>
+    <include file="changesets/20200601_CopyfileTable.xml"                        relativeToChangelogFile="true"/>
```
pshapiro4broad, line 18: Not sure if you already know this, but you can use `<includeAll>` to load all changelogs in a directory. It gives you less control but it makes adding changelogs easier, since usually what you want is to load all changelogs.

ehigham: Nice. This'll reduce noise in future pull requests

ehigham: On reflection, the liquibase doc actually suggests advises against it:
https://docs.liquibase.com/concepts/advanced/includeall.html

test/zero/tools/fixtures.clj
```diff
@@ -0,0 +1,35 @@
...
+      ;; <- temporary folder deleted
+  "
+  [uri & body]
+  `(let [name# (join ["wfl-test-" (UUID/randomUUID) "/"])
+         ~uri (gcs/gs-url gcs-test-bucket name#)]
+     (try ~@body
+          (finally
+            (->>
+              (gcs/list-objects gcs-test-bucket name#)
+              (run! (comp delete-test-object :name)))))))
```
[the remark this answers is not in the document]
tbl3rd: I did something similar, but not specific to any bucket, then backed it out.

PR #48, opened 2020-06-13
GH-671: Build-Arrays-Module for AOU

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-671

### Changes

- Add something for aou arrays module. So many things are UNCERTAIN, sigh :(

### Questions and Assumptions
- [x] Which pipeline and version to use?
    - Arrays.wdl
    - Arrays_v1.9
- [x] The testing inputs?
    - https://github.com/broadinstitute/dsde-pipelines/blob/Arrays_v1.9/pipelines/arrays/single_sample/test_inputs/Plumbing/SimpleInput.json
- [x] Do we need external wrapper?
    - No, we use `final_xxx` workflow option now. **Not implemented yet**

### The current design

- The workload schema table (i.e. `workload`) stores workload information that can be used to query for a group of workflows that belong to the same workload. 
- `/workload/create` can add records to workload schema table: If there’s already a record with the same (pipeline-name + pipeline-version + project-name), it will return the existing one, otherwise it will create a new row.
- The workload items tables (e.g. `aou_000001`) only store minimal information to be able to query for unique workflows afterwards.
- `/workload/start` can mark a workload as “started“.
- `/workload/exec` does the combination of both `create` and `start`
- `/workload/append` can append workflow items to a “started” workload items tables.

### Review Instructions

- No instructions.


rexwangcc:
The table contains per-sample information for now, but in the long-term it will be the pub/sub set up info.

database/changesets/20200616_AllOfUsArraysTable.xml
```diff
@@ -16,15 +16,19 @@
         </comment>
         <sql>
             CREATE TYPE AllOfUsArrays AS (
-                id                          bigint,      -- unique to this table
-                analysis_version_number     text,        -- input to the workflow
-                chip_well_barcode           text,        -- input to the workflow
+                analysis_version_number     bigint,      -- input to the workflow (PK part1)
```
ehigham, line 19: I think you can dispense with the `-- input to the workflow` comments :)

src/zero/module/aou.clj
```diff
@@ -174,17 +176,117 @@
...
+            table (format "%s_%09d" pipeline id)
+            kind  (format (str/join " " ["UPDATE workload"
+                                         "SET pipeline = '%s'::pipeline"
+                                         "WHERE id = %s"]) pipeline id)
+            work  (format "CREATE TABLE %s OF %s (PRIMARY KEY (analysis_version_number, chip_well_barcode))" table pipeline)]
+        (jdbc/update! tx :workload {:items table} ["id = ?" id])
+        (jdbc/db-do-commands tx [kind work])
+        [uuid table]))))
+
+(comment
```
ehigham, line 214: Controversial opinion warning:
I'm not a fan of comment blocks (sorry tom). They cluter the code and quickly go stale.
If this is something that's actually useful for debugging, make it a test.

rexwangcc: Thanks! I'm not a big fan of comment blocks in production code either, these blocks will go away when this PR's merged. They are here just for the sake of a faster turn around time for myself :octocat:

database/changesets/20200616_AllOfUsArraysTable.xml
```diff
@@ -0,0 +1,37 @@
...
+               dbms="postgresql"
+               id="20200616"
+               logicalFilePath="20200616_AllOfUsArraysTable.xml">
+        <comment>
+            the AllOfUsArrays table schema
+        </comment>
+        <sql>
+            CREATE TYPE AllOfUsArrays AS (
+                analysis_version_number     bigint,      -- input to the workflow (PK part1)
+                chip_well_barcode           text,        -- input to the workflow (PK part2)
```
rexwangcc, line 20: @pshapiro4broad @gbggrant Is it safe to make the assumption that `(analysis_version_number, chip_well_barcode)` pairs will always be unique, unless we intentionally want to re-run a sample with a specific version?

gbggrant: @rexwangcc - yes, that's a safe assumption to make.  You really woudn't re-run a sample with a specific version in production.

rexwangcc: Thanks!!

tbl3rd:
Trying to build this, I see:

clojure.lang.ExceptionInfo: Unable to resolve symbol: extract-and-validate-per-sample-inputs in this context

I think that's because make-inputs references it before it's defined.

rexwangcc:
> Trying to build this, I see:
> 
> clojure.lang.ExceptionInfo: Unable to resolve symbol: extract-and-validate-per-sample-inputs in this context
> 
> I think that's because make-inputs references it before it's defined.

oops, thanks for catching it, yeah I did not build after the latest change..

src/zero/module/aou.clj
```diff
@@ -1,23 +1,400 @@
 (ns zero.module.aou
-  "Process Arrays for the All Of Us project.")
+  "Process Arrays for the All Of Us project."
+  (:require [clojure.java.io :as io]
+            [clojure.data.json :as json]
```
samanehsan, line 4: According to IntelliJ there are a few unused imports here that can be removed (json, set, all & gcs)

src/zero/module/aou.clj
```diff
@@ -1,23 +1,400 @@
...
+  "The WDL label applied to Cromwell metadata."
+  {(keyword (str zero/the-name "-aou"))
+   (wdl/workflow-name (:top workflow-wdl))})
+
+(def cromwell-label
+  "The WDL label applied to Cromwell metadata."
+  (let [[key value] (first cromwell-label-map)]
+    (str (name key) ":" value)))
+
+#_(def per-sample
```
samanehsan, line 37: This comment looks old...can it be removed?

src/zero/module/aou.clj
```diff
@@ -1,23 +1,400 @@
...
+   :subsampled_metrics_interval_list nil
+   :disk_size                        100
+   :preemptible_tries                3})
+
+(defn- map-aou-environment
+  "Map AOU-ENV to envrionment for inputs preparation."
+  [aou-env]
+  ({:aou-dev "dev" :aou-prod "prod"} aou-env))
+
+(defn array-inputs
```
samanehsan, line 86: Maybe call this `env-inputs` or `array-env-inputs` since it does not return all of the array inputs?

src/zero/module/aou.clj
```diff
@@ -1,23 +1,400 @@
...
+   to the existing one."
   [tx body]
-  (->> body
-       first
-       (filter second)
-       (into {})))
+  (let [{:keys [creator cromwell pipeline project]} body
+        {:keys [release top]} workflow-wdl
+        {:keys [commit version]} (zero/get-the-version)
+        probe-result (jdbc/query tx ["SELECT * FROM workload WHERE project = ? AND wdl = ? AND release = ?"
```
samanehsan, line 236: Why query by the `wdl` field instead of the `pipeline`? The `workflow-wdl` :top has changed independently of the :version before when we refactored the WFL build process so I'm not sure how much we want to rely on that 🤔

src/zero/module/aou.clj
```diff
@@ -1,23 +1,400 @@
...
+                                        :green_idat_cloud_path
+                                        :params_file
+                                        :red_idat_cloud_path
+                                        :reported_gender
+                                        :sample_alias
+                                        :sample_lsid
+                                        :zcall_thresholds_file])]
+    (if (every? (fn [x] (not (nil? x))) (vals per-sample))
+      per-sample
+      (throw (Exception. (format "Invalid sample inputs %s" per-sample))))))
```
tbl3rd, line 113: Suggest: https://github.com/broadinstitute/wfl/blob/tbl/rex/GH-671-Build-Arrays-Module/src/zero/module/aou.clj#L95-L114

src/zero/module/aou.clj
rexwangcc: Nice catch, will do!

src/zero/api/handlers.clj
```diff
@@ -51,6 +51,15 @@
         results (wgs/submit-some-workflows env max input_path output_path)]
     (succeed {:results results})))
 
+(defn append-to-aou-workload
+  "Append new work to an existing started AoU workload describe in BODY of _REQUEST."
```
jack-r-warren, line 55: Maybe specifying workflows appended to workloads would help disambiguate a bit
```suggestion
  "Append new workflows to an existing started AoU workload describe in BODY of _REQUEST."
```

src/zero/module/aou.clj
rexwangcc: removed!

src/zero/references.clj
```diff
@@ -52,3 +52,12 @@
...
      :reference_fasta             reference_fasta}))
+
+(def hg19-arrays-references
+  "HG19 reference files for arrays processing."
+  {:ref_fasta                   "gs://gcp-public-data--broad-references/hg19/v0/Homo_sapiens_assembly19.fasta"
+   :ref_fasta_index             "gs://gcp-public-data--broad-references/hg19/v0/Homo_sapiens_assembly19.fasta.fai"
+   :ref_dict                    "gs://gcp-public-data--broad-references/hg19/v0/Homo_sapiens_assembly19.dict"
+   :dbSNP_vcf                   "gs://gcp-public-data--broad-references/hg19/v0/dbsnp_138.b37.vcf.gz"
+   :dbSNP_vcf_index             "gs://gcp-public-data--broad-references/hg19/v0/dbsnp_138.b37.vcf.gz.tbi"}
+  )
```
jack-r-warren, line 63: I doubt Clojure or any of our unix-line-ending machines care but this might avoid issues if someone with Windows tries to clone this repo
```suggestion
  )
  
```

src/zero/module/aou.clj
rexwangcc: changed to `env-inputs`

src/zero/module/aou.clj
rexwangcc: you are right, I think query `pipeline` is much safer. changed.

src/zero/module/aou.clj
```diff
@@ -1,23 +1,400 @@
...
+  ({:aou-dev "dev" :aou-prod "prod"} aou-env))
+
+(defn array-inputs
+  "Array inputs for ENVIRONMENT that do not depend on the input file."
+  [environment]
+  {:vault_token_path (get-in env/stuff [environment :vault_token_path])
+   :environment      (map-aou-environment environment)})
+
+(comment
+  (array-inputs :aou-dev))
```
jack-r-warren: I don't really have an issue with leaving these `comment`s around to help in future debugging, but it would super helpful if they were briefly commented with an explanation. I also know there was some talk about removing comments when development on this was done--I don't care either way

src/zero/module/aou.clj
rexwangcc: +1 to return filtered inputs and throw missing.
btw, the link doesn't run: 
```
Execution error (ArityException) at zero.module.aou/eval10123 (aou.clj:102).
Wrong number of args (3) passed to: clojure.lang.PersistentVector
```

database/changesets/20200317_ExternalWholeGenomeReprocessingTable.xml
```diff
@@ -16,15 +16,16 @@
         </comment>
```
ehigham, line 16: Does this file really need to be indented for this feature?

samanehsan:
Will Ops need to create/start any Arrays workloads through the UI? Since the functionality slightly differs from the other modules that could be tricky 🤔 In other words, there would be no way to "preview" array workload samples before starting their corresponding analysis workflows.

src/zero/api/handlers.clj
```diff
@@ -51,6 +51,15 @@
         results (wgs/submit-some-workflows env max input_path output_path)]
     (succeed {:results results})))
 
+(defn append-to-aou-workload
```
samanehsan, line 54: The ability to add samples to an existing workload might be useful to have for other pipelines as well! But this can be a future discussion with Ops.

src/zero/module/aou.clj
```diff
@@ -1,23 +1,385 @@
...
+                                        :input    "aou-inputs-placeholder"
+                                        :output   "aou-outputs-placeholder"
+                                        :project  project
+                                        :release  release
+                                        :uuid     (UUID/randomUUID)
+                                        :version  version
+                                        :wdl      top})
+            table (format "%s_%09d" pipeline id)
+            table_seq (format "%s_id_seq" table)
+            kind  (format (str/join " " ["UPDATE workload"
```
samanehsan, line 242: You can use a prepared statement here: https://github.com/broadinstitute/wfl/blob/ec7cc6bf0f68fa4d81f9499796bc3bbece57f42e/src/zero/module/all.clj#L105

src/zero/module/aou.clj
```diff
@@ -1,23 +1,385 @@
...
+  [tx samples table]
+  (letfn [(q [[left right]] (fn [it] (str left it right)))
+          (extract-key-groups [l] [(map :chip_well_barcode l) (map :analysis_version_number l)])
+          (make-query-group [v] (->> v (map (q "''")) (str/join ",") ((q "()"))))
+          (assemble-query [query l] (format query (first l) (last l)))]
+    (let [samples  (map keep-primary-keys samples)
+          existing (->> samples
+                        (extract-key-groups)
+                        (map make-query-group)
+                        (assemble-query (str/join " " ["SELECT * FROM" table "WHERE chip_well_barcode in %s"
```
samanehsan, line 285: A more secure way to do this query is to construct the parameterized string dynamically (since there is a variable number of samples) and then set each parameter like this! https://github.com/broadinstitute/wfl/blob/ec7cc6bf0f68fa4d81f9499796bc3bbece57f42e/src/zero/api/handlers.clj#L105-L111

src/zero/module/aou.clj
tbl3rd: Sorry. That's what I get for not running it. It should work better now.

src/zero/module/aou.clj
```diff
@@ -1,23 +1,385 @@
...
+(defn make-options
+  "Return options for aou arrays pipeline."
+  []
+  {; TODO: add :final_workflow_outputs_dir here
+   ; TODO: add :default_runtime_attributes {:maxRetries 3} here
+   :read_from_cache            true
+   :write_to_cache             true
+   :default_runtime_attributes {:zones "us-central1-a us-central1-b us-central1-c us-central1-f"}})
+
+(defn active-or-done-objects
```
samanehsan, line 140: Does this get used anywhere at the moment?

src/zero/module/aou.clj
```diff
@@ -1,23 +1,385 @@
...
+                      analysis-verion-chip-well-barcode)))))]
+      (->> {:label  cromwell-label
+            :status ["On Hold" "Running" "Submitted" "Succeeded"]}
+           (cromwell/query environment)
+           (keep active?)
+           (filter seq)
+           set))))
+
+(comment
+  (active-objects :gotc-dev {:analysis_version_number 1, :chip_well_barcode "7991775143_R01C01"}))
```
samanehsan, line 165: I'm not sure if you were planning to leave this comment in, but I noticed the fn is called `active-or-done-objects` now!

samanehsan:
The `get-workload` endpoint updates the associated [workflow statuses](https://github.com/broadinstitute/wfl/blob/master/src/zero/service/postgres.clj#L89) and then returns the workload + the updated workflows.

Since the request to get the workflow status from Cromwell uses the uuid I think we'll need to adjust this slightly for Arrays. I think this can happen in a follow-up PR though since this one is pretty big 😅

src/zero/module/aou.clj
rexwangcc: Thanks! It works now with 1 tiny hack:
```diff
(defn get-per-sample-inputs
  "Throw or return per-sample INPUTS."
  [inputs]
  (let [the-keys [:analysis_version_number
                  :bead_pool_manifest_file
                  :chip_well_barcode
                  :cluster_file
                  :extended_chip_manifest_file
                  :gender_cluster_file
                  :green_idat_cloud_path
                  :params_file
                  :red_idat_cloud_path
                  :reported_gender
                  :sample_alias
                  :sample_lsid
                  :zcall_thresholds_file]
        result  (select-keys inputs the-keys)
-        missing (vec (keep (fn [k] (when (nil? (k result))) k) the-keys))]
+        missing (vec (keep (fn [k] (when (nil? (k result))k)) the-keys))]
    (when (seq missing)
      (throw (Exception. (format "Missing per-sample inputs: %s" missing))))
    result))
```

database/changesets/20200317_ExternalWholeGenomeReprocessingTable.xml
rexwangcc: My suspect is no 🤔

src/zero/module/aou.clj
rexwangcc: nice catch! Yeah, Let me rename it just in case.

src/zero/module/aou.clj
rexwangcc: This is not being used, but I tend to keep it since we are very likely to use it in the future. WDYT?

src/zero/module/aou.clj
rexwangcc: Thanks! I'm wondering how could we apply prepared statement for something like `(jdbc/db-do-commands tx [kind idx work link-idx-work])` that requires a sequence of sql executions, would we thread them in a seq of `(jdbc/execute! foo)`s?

rexwangcc:
> The `get-workload` endpoint [updates the associated workflow statuses](https://github.com/broadinstitute/wfl/blob/master/src/zero/service/postgres.clj#L89) and then returns the workload + the updated workflows.
> 
> Since the request to get the workflow status from Cromwell uses the uuid I think we'll need to adjust this slightly for Arrays. I think this can happen in a follow-up PR though since this one is pretty big and I don't think it will actually break this endpoint (it just won't update the statuses correctly) 😅

Sorry, I'm not sure if I understand this correctly, `get-workload` endpoint, which calls `postgres/get-workload-for-uuid` under the hood uses the workload uuid to get back all workflows in the `items` table and update each workflow one by one: 

```
 id | analysis_version_number | chip_well_barcode | status  |            updated            |                 uuid
----+-------------------------+-------------------+---------+-------------------------------+--------------------------------------
  3 |                       1 | 7991775143_R01C01 | Running | 2020-07-17 14:51:10.933556-04 | foo-uuid
  4 |                       2 | 7991775143_R01C01 | Running | 2020-07-17 14:51:12.606813-04 | bar-uuid
```

src/zero/api/handlers.clj
rexwangcc: yes! I think we could make make this an implementation of the `update-workload!` multi-method and apply it to all types of workloads (pipelines)! I agree this might be better in another PR after discussing with hOps

src/zero/module/aou.clj
samanehsan: Ok, let's keep it in then!

samanehsan:
> > The `get-workload` endpoint [updates the associated workflow statuses](https://github.com/broadinstitute/wfl/blob/master/src/zero/service/postgres.clj#L89) and then returns the workload + the updated workflows.
> > Since the request to get the workflow status from Cromwell uses the uuid I think we'll need to adjust this slightly for Arrays. I think this can happen in a follow-up PR though since this one is pretty big and I don't think it will actually break this endpoint (it just won't update the statuses correctly) 😅
> 
> Sorry, I'm not sure if I understand this correctly, `get-workload` endpoint, which calls `postgres/get-workload-for-uuid` under the hood uses the workload uuid to get back all workflows in the `items` table and update each workflow one by one:
> 
> ```
>  id | analysis_version_number | chip_well_barcode | status  |            updated            |                 uuid
> ----+-------------------------+-------------------+---------+-------------------------------+--------------------------------------
>   3 |                       1 | 7991775143_R01C01 | Running | 2020-07-17 14:51:10.933556-04 | foo-uuid
>   4 |                       2 | 7991775143_R01C01 | Running | 2020-07-17 14:51:12.606813-04 | bar-uuid
> ```

Oh okay! You can ignore my previous comment then...I got confused and thought the uuids wouldn't be stored 🤦‍♀️

src/zero/module/aou.clj
samanehsan: That seems like it would work! There is also `(jdbc/db-do-prepared foo)` which looks like the prepared statement equivalent to `(jdbc/db-do-commands foo)`

rexwangcc: `jdbc/db-do-prepared` looks cool! Let me take a look 👀 !

PR #50, opened 2020-06-15
[GH-782] Test pubsub notifications for GCS

samanehsan:
### Purpose
Test that a pubsub message is added to a topic when a file is uploaded to a GCS bucket configured to use pubsub notifications.

### Changes
- Add an acknowledge function to the pubsub module
- Add a test to `pubsub_test.clj` for GH-782

### Review Instructions

- No instructions.

### DIFF
--- a/test/zero/unit/pubsub_test.clj
+++ b/test/zero/unit/pubsub_test.clj
@@ -1,8 +1,10 @@
 (ns zero.unit.pubsub-test
   "Test the Google Cloud Storage namespace."
-  (:require [clojure.string      :as str]
+  (:require [clojure.java.io     :as io]
+            [clojure.string      :as str]
             [clojure.test        :refer [deftest is testing]]
-            [zero.service.pubsub :as pubsub])
+            [zero.service.pubsub :as pubsub]
+            [zero.service.gcs    :as gcs])
   (:import [java.util UUID]))
 
 (def project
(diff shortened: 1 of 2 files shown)


test/zero/unit/pubsub_test.clj
```diff
@@ -76,3 +78,30 @@
...
+
+(def notification-bucket
+  "A GCS bucket that publishes to a PubSub topic on file upload events."
+  "test-storage-notifications")
+
+(def notification-subscription
+  "The subscription to the PubSub topic that the notification-bucket publishes to."
+  "test-storage-notification-subscription")
+
+(deftest bucket-notification-test
```
ehigham, line 90: Looks good. Could you randomise the upload location to avoid sporadics from concurrent testing? There's a macro`zero.tools.fixtures/with-temporary-gcs-folder` that does this for you:
- It gives you a gcs object prefix to prepend to your object names
- It cleans up all resources with that prefix on return

See `test/zero/integration/v1_endpoint_test.clj:61` for an example

samanehsan: I see that macro creates a temp folder in the "broad-gotc-dev-zero-test" bucket, but I used a different bucket for this test since it needs to be configured to publish messages to a topic. I can try making the bucket a parameter of the macro instead...

ehigham: Ah OK, if it's too much work don't worry. It's more a case of avoiding weird errors if one of us happens to run this test at the same time (or concurrently in CI)

samanehsan: That makes sense! I attempted to refactor the existing macro but didn't get very far so I updated the test without it 😅

PR #53, opened 2020-06-19
GH-803: Fix typos.

tbl3rd:
https://broadinstitute.atlassian.net/browse/GH-803

### DIFF
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -99,9 +99,7 @@ class CLI:
         return subprocess.check_call(command, shell=True)
 
     @staticmethod
-    def _subprocess_call_unchecked_because_who_cares_if_it_fails(
-            command: str
-    ):
+    def _subprocess_call_unchecked( command: str ):
         """Run COMMAND in a subprocess and who cares whether it fails!"""
         print(f"Running: {command}")
         return subprocess.call(command, shell=True)
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -99,7 +99,7 @@ class CLI:
         return subprocess.check_call(command, shell=True)
 
     @staticmethod
-    def _subprocess_call_unchecked( command: str ):
+    def _subprocess_call_unchecked(command: str):
         """Run COMMAND in a subprocess and who cares whether it fails!"""
         print(f"Running: {command}")
         return subprocess.call(command, shell=True)
@@ -166,7 +166,7 @@ class CLI:
                 color="green"
             )
         )
-        CLI._subprocess_call_unchecked_because_who_cares_if_it_fails(command)
+        CLI._subprocess_call_unchecked(command)
 
     @staticmethod
     def _is_available(*commands: list):
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -4,6 +4,8 @@
 
 requirements: pip3 install gitpython
 
+Example: ./ops/cli.py -t latest
+
 usage: python3 cli.py -h
 """
 from enum import Enum
@@ -66,7 +68,7 @@ def register(func: Callable) -> Callable:
 class CLI:
     def __init__(self):
         parser = argparse.ArgumentParser(
-            description="WFL Development CLI", usage=self._usage()
+            description="Deploy the workflow-launcher-api.", usage=self._usage()
         )
         parser.add_argument("command", help="command from the above list to run")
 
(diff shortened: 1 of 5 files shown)


ops/cli.py
```diff
@@ -86,35 +82,55 @@ def __call__(self) -> int:
...
         return msg
 
+    @staticmethod
+    def _subprocess_call(command: str):
+        """Run COMMAND in a subprocess."""
+        print(f"Running: {command}")
+        return subprocess.check_call(command, shell=True)
+
+    @staticmethod
+    def _subprocess_call_unchecked_because_who_cares_if_it_fails(
```
ehigham, line 102: If we don't care if this fails then why are we calling it?

ops/cli.py
```diff
@@ -86,35 +82,55 @@ def __call__(self) -> int:
...
+            command: str
+    ):
```
ehigham, line 104: Maybe dial down the sass...
```suggestion
    def _subprocess_call_unchecked( command: str ):
```

ops/cli.py
```diff
@@ -134,138 +150,153 @@ def _render_ctmpl(ctmpl_file: str, **kwargs) -> int:
...
-            )
-
-            file_name = "wfl-values.yaml.ctmpl"
-            shutil.copy(
-                str(Path(f"./{dir_name}/gotc-deploy/deploy/gotc-dev/helm/{file_name}")),
-                Path(f"./{dir_name}/{file_name}"),
-            )
+        with tempfile.TemporaryDirectory() as dir:
+            pwd = os.getcwd()
+            os.chdir(dir)
```
ehigham, line 290: Do you need to change dir or can you just clone to this directory directly? If an exception is thrown then would the current directory be in this stale location?

ops/cli.py
tbl3rd: My point exactly.

ops/cli.py
tbl3rd: Yes, but who cares?
We could wrap it in a try.
This seemed to be the simplest thing that can work.
https://agilefaq.wordpress.com/2007/11/03/what-is-kiss-principle/ :agileaf:

ehigham: Rex already implemented the ability to specify the folder so there was no need for a try or to manually manage system resources.

tbl3rd: After adding the subprocess reporting, I saw pathnames like "././...".
I figured that was due to things like str(Path(f"…")) getting passed around
This way, we just let the system choose the directory and we run in it.
Otherwise you're bouncing between strings and paths to coerce filenames.
I believe this is simpler.  Restore what was there before and compare.
I'll back out the changes if you still prefer that.  Thanks.

ehigham: OK - when you put it like this this sounds sensible.

ops/cli.py
```diff
@@ -68,11 +66,9 @@ def register(func: Callable) -> Callable:
 class CLI:
     def __init__(self):
         parser = argparse.ArgumentParser(
-            description="WFL Development CLI", usage=self._usage()
+            description="Deploy the workflow-launcher-api.", usage=self._usage()
```
rexwangcc, line 69:
```suggestion
            description="Deploy the workflow-launcher-api and workflow-launcher-ui.", usage=self._usage()
```

ops/cli.py
```diff
@@ -86,35 +82,53 @@ def __call__(self) -> int:
         return getattr(self, args.command)(sys.argv[2:])
 
     def _usage(self) -> str:
-        msg, space = "\n", 20
+        max = 0
+        for command in sorted(self.registered_commands):
+            if len(command) > max:
+                max = len(command)
+        msg = "\n"
```
rexwangcc, line 89: smart

ops/cli.py
```diff
@@ -86,35 +82,53 @@ def __call__(self) -> int:
...
 
+    @staticmethod
+    def _subprocess_call(command: str):
+        """Run COMMAND in a subprocess."""
+        print(f"Running: {command}")
+        return subprocess.check_call(command, shell=True)
+
+    @staticmethod
+    def _subprocess_call_unchecked( command: str ):
+        """Run COMMAND in a subprocess and who cares whether it fails!"""
```
rexwangcc, line 103: 🤷‍♂️

PR #55, opened 2020-06-23
only print success when the command succeeded + tidy-ups

ehigham:
### Purpose
The logging in CLI doesn't match the state of the program. For example, we can log "success" when the command actually failed because success is printed before that command is evaluated.

Aim of this change is to tidy-up this behaviour. I also took the liberty of
- grouping together functions that are called in succession (e.g. `print` and `dye_message_with_color`).
- ensuring defaults are defined in one place
- removing unnecessary static methods

### DIFF
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -23,7 +23,6 @@ import tempfile
 if sys.platform.lower() == "win32":
     os.system("color")
 
-
 class AvailableColors(Enum):
     GRAY = 90
     RED = 91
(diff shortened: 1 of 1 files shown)


ops/cli.py
```diff
@@ -35,16 +34,30 @@ class AvailableColors(Enum):
...
+    """Log the info to stdout"""
+    print(_apply_color("default", message))
+
+def success(message: str):
+    """Log the success to stdout"""
+    print(_apply_color("green", message))
 
+def error(message: str):
+    """Log the error to stderr"""
+    print(_apply_color("red", message), file=sys.stderr)
```
rexwangcc, line 56: Like this!

PR #59, opened 2020-06-26
Fix minor issues with WFL Helm and Unit Tests

rexwangcc:
### Purpose

### Changes

- No changes.

### Review Instructions

- No instructions.


test/zero/integration/pull_from_pubsub_test.clj
```diff
@@ -0,0 +1,40 @@
...
+          (is (= (:bucket data) notification-bucket))
+          (testing "acknowledge"
+            (let [ack-ids (map :ackId result)
+                  ack-result (pubsub/acknowledge project notification-subscription ack-ids)
+                  pull-result (pubsub/pull project notification-subscription)]
+              (is (= ack-result {}))
+              (is (= (count pull-result) 0))))))
+      (finally
+        (gcs/delete-object notification-bucket test-file)
+        (io/delete-file "test.txt")))))
```
ehigham, line 40: newline here please

test/zero/integration/pull_from_pubsub_test.clj
```diff
@@ -0,0 +1,40 @@
...
+        (let [result (pubsub/pull project notification-subscription)
+              data (get-in (first result) [:message :data])]
+          (is (= (:name data) test-file))
+          (is (= (:bucket data) notification-bucket))
+          (testing "acknowledge"
+            (let [ack-ids (map :ackId result)
+                  ack-result (pubsub/acknowledge project notification-subscription ack-ids)
+                  pull-result (pubsub/pull project notification-subscription)]
+              (is (= ack-result {}))
+              (is (= (count pull-result) 0))))))
```
ehigham, line 37: have you created a separate issue to handle this test bug?

rexwangcc: I realized simply removing this count won't work since as long as we don't use different topics (queues) for each run, the `ack-result` can also be affected by race conditions and thus give us different results. This whole test needs to be revamped and I'm going to move the work to my AoU PR. as line 8 mentioned. What do you think?

test/zero/integration/pull_from_pubsub_test.clj
rexwangcc: 🤦 good catch

rexwangcc, line 40:
```suggestion
        (io/delete-file "test.txt")))))

```

test/zero/integration/pull_from_pubsub_test.clj
samanehsan: A separate PR sounds good to me :octocat: 

Also one idea I thought of to fix the testing logic is to only acknowledge the message associated with the file that was uploaded by the test and then confirm that the message is no longer in the "pull-result". Then it will not interfere with concurrent runs!

PR #63, opened 2020-07-09
GH-820: Strictly lint helm files

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-820
- The existing Helm chart testing action worked fine but the [chart-testing](https://github.com/helm/chart-testing) it made use of has an outstanding issue ([stale PR](https://github.com/helm/chart-testing/pull/200)) that meant it didn't surface any warnings

### Changes

- Deleted the older helmchart-test.yml
  - Deleted rather than rewritten because I needed both actions simultaneously during my testing and I don't want to lose my commit comments to the new file
  - Deleting commit references a commit for the new action, so someone really needing to can still follow history
- Added a new action, pr-helm-lint.yml that's able to customize command line arguments to pass `--strict`
- **Edited helm/wfl/values.yaml to make helm/wfl/templates/_helpers.tpl pass linting**
  - In a number of places, we used if statements for existence checks on things not in values.yaml, since `nil` is "empty"/falsy
  - Now, all keys are defined in values.yaml, just with empty strings, which are also "empty"/falsy (docs for [default](https://helm.sh/docs/chart_template_guide/function_list/#default) and [if](https://helm.sh/docs/chart_template_guide/control_structures/#ifelse))

### Review Instructions

- For testing the action:
  - Fork WFL, merge this branch into master, and then make a branch with edits to files in helm/wfl and try to PR it
  - You'll have to specifically visit the actions page of your fork on GitHub to allow them to run before you make a PR
- For regression testing the values.yaml file:
  - If we believe Helm's docs, this is a no-op change that just satisfies the linter
  - If there's no change in Helm's output and we can still customize `fullnameOverride`, `nameOverride`, and `serviceAccount.name` then it works
    - I need to get Kubernetes working in order to test this myself. ~~I'll do that, I just figured I could get this in front of other folks' eyes since I'm pretty confident the code here is done.~~ I've now done this, there's no change in behavior I can find (output is the same, arguments still work).


jack-r-warren:
Since the PR is coming from a fork, it is using my fork's GitHub Actions. This means:
- I had to bump the chart version since the linting detects that
- It won't pass the normal build because my repo's actions don't have the permissions that build expects

helm/wfl/Chart.yaml
```diff
@@ -4,10 +4,10 @@ description: A Helm chart for deploy WorkFlow Launcher on Kubernetes
 
 type: application
 
-version: 0.2.6
+version: 0.2.7
 
 home: https://github.com/broadinstitute/wfl
-appVersion: 0.2.6
+appVersion: 0.2.7
```
jack-r-warren, line 10: Even though my change to values.yaml is a no-op, the linting wants a version bump here. If we want, we can revert afa751a to get rid of this version bump (linting will complain but it won't affect anything).

.github/workflows/pr-helm-lint.yml
```diff
@@ -0,0 +1,54 @@
...
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - helm/wfl/**
+  pull_request:
+    paths:
+      - helm/wfl/**
```
jack-r-warren, line 13: I'm targeting helm/wfl specifically because we'd have to reconfigure this action a bit if we added other charts. Basic linting (just errors) will work on everything because chart-testing has logic to find charts based on ct.yaml, but the strict linting is just done via command line and uses a working directory of helm/wfl.

jack-r-warren:
> It would be great to let helm be more tolerant on version bumps, but that's trivial to me and this PR already does a lot good. 

That’s something I have a bit of control over, actually. The version check and a “is there a maintainer” check both come from running `ct lint`, they’re actually the only reason we use `ct` here. I can’t configure the checks but if we don’t care for the version one I can easily remove `ct` from the action altogether.

jack-r-warren:
Discussed with folks after standup, determined that we like the version check. Merging.

PR #64, opened 2020-07-14
[GH-761] Use prepared SQL statements

samanehsan:
### Purpose
Address potential security vulnerabilities from concatenating user input with raw SQL statements:
https://broadinstitute.atlassian.net/jira/software/projects/GH/boards/530?selectedIssue=GH-761

### Changes
Here are the files that currently interact with the database along with the changes I made:
- service.postgres: Modify `get-table` to check that the table exists before using that variable in a SQL query.
- api.handlers: Modify `post-start` to use a parameterized statement instead of concatenating the uuids provided by the user. 
- module.all: Modify `add-workload-table!` to use a parameterized statement when updating the workload.
- module.wfl
- module.copyfile

### Review Instructions

Any ideas what to do about this DB operation to create a new typed table? I don't think there's a way for us to use a parameterized statement here:
https://github.com/broadinstitute/wfl/compare/se-sql-ops-updates?expand=1#diff-3c51ca22431fe3b9448ef2882cbd193aR102

### DIFF
--- a/src/zero/api/handlers.clj
+++ b/src/zero/api/handlers.clj
@@ -9,7 +9,8 @@
             [zero.module.wl :as wl]
             [zero.service.cromwell :as cromwell]
             [zero.service.postgres :as postgres]
-            [zero.zero :as zero]))
+            [zero.zero :as zero])
+  (:import [java.sql PreparedStatement]))
 
 (defn fail
   "A failure response with BODY."
@@ -109,12 +110,15 @@
             ps (jdbc/prepare-statement db-conn query)]
         (doseq [[i uuid] (map-indexed vector uuids)] (.setString ps (+ i 1) (:uuid uuid)))
         (jdbc/with-db-transaction [tx db-config]
-          (->> ps
+          (try
+            (->> ps
                (jdbc/query tx)
                (run! (partial start-workload! tx)))
-          (->> uuids
+            (->> uuids
                (mapv (partial postgres/get-workload-for-uuid tx))
-               succeed))))))
+               succeed)
+            (catch Exception e
+              (fail {:message (.getMessage e)}))))))))
 
 (def post-exec
   "Create and start workload described in BODY of REQUEST"
--- a/src/zero/api/handlers.clj
+++ b/src/zero/api/handlers.clj
@@ -110,15 +110,12 @@
             ps (jdbc/prepare-statement db-conn query)]
         (doseq [[i uuid] (map-indexed vector uuids)] (.setString ps (+ i 1) (:uuid uuid)))
         (jdbc/with-db-transaction [tx db-config]
-          (try
-            (->> ps
+          (->> ps
                (jdbc/query tx)
                (run! (partial start-workload! tx)))
             (->> uuids
                (mapv (partial postgres/get-workload-for-uuid tx))
-               succeed)
-            (catch Exception e
-              (fail {:message (.getMessage e)}))))))))
+               succeed))))))
 
 (def post-exec
   "Create and start workload described in BODY of REQUEST"
--- a/src/zero/api/handlers.clj
+++ b/src/zero/api/handlers.clj
@@ -101,16 +101,20 @@
   [request]
   (let [uuids (-> request :parameters :body distinct)]
     (letfn [(q [[left right]] (fn [it] (str left it right)))]
-      (jdbc/with-db-transaction [tx (postgres/zero-db-config)]
-        (->> uuids
-             (map :uuid)
-             (map (q "''")) (str/join ",") ((q "()"))
-             (format "SELECT * FROM workload WHERE uuid in %s")
-             (jdbc/query tx)
-             (run! (partial start-workload! tx)))
-        (->> uuids
-             (mapv (partial postgres/get-workload-for-uuid tx))
-             succeed)))))
+      (let [db-config (postgres/zero-db-config)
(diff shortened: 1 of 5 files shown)


src/zero/module/copyfile.clj
```diff
@@ -47,5 +47,5 @@
                 (jdbc/update! tx items
                               {:updated now :uuid uuid}
                               ["id = ?" id])))]
-      (let [workflow (postgres/get-table tx items)]
+      (if-let [workflow (postgres/get-table tx items)]
```
ehigham, line 50: I feel that the if-expression requires some handling for the case when workflow is `nil`. Is there a situation when `nil` is valid at this point? If not, should we just let it fail rather than program defensively? Can we handle this case at this point rather than fail silently?

ehigham: Note this also applies to the other call sites where there's an `if-let` now :)

src/zero/service/postgres.clj
```diff
@@ -52,10 +52,19 @@
...
+  (->> ["SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = ?" (str/lower-case table)]
+       (jdbc/query tx)
+       (count)
+       (not= 0)))
+
 (defn get-table
   "Return TABLE using transaction TX."
   [tx table]
-  (jdbc/query tx (format "SELECT * FROM %s" table)))
+  (if (table-exists? tx table)
```
ehigham, line 66: You've changed the API here and now it seems one needs to `nil` check the result of this case.
Personally, I'd just throw if it doesn't exist because it's exceptional in the case of `start`. You can handle that higher up the stack in the case of `get-workload` where perhaps you can compare the exception identifier to a known string (say `wfl::no-such-table`) and return `[]` in that case.

src/zero/module/copyfile.clj
samanehsan: That's a good point! Since the workload "items" table gets created when the workload is created, this call to `get-table` should not return `nil` unless something is wrong with the database

src/zero/api/handlers.clj
```diff
@@ -101,16 +102,23 @@
...
+        (doseq [[i uuid] (map-indexed vector uuids)] (.setString ps (+ i 1) (:uuid uuid)))
+        (jdbc/with-db-transaction [tx db-config]
+          (try
+            (->> ps
+               (jdbc/query tx)
+               (run! (partial start-workload! tx)))
+            (->> uuids
+               (mapv (partial postgres/get-workload-for-uuid tx))
+               succeed)
+            (catch Exception e
```
ehigham, line 120: I think there's error handling further up the stack in the http request handling library that reponds to a request with the exception message.
As a side note, we'll probably do something like this later when we add logging

src/zero/module/copyfile.clj
```diff
@@ -39,13 +40,15 @@
...
                 (jdbc/update! tx items
                               {:updated now :uuid uuid}
                               ["id = ?" id])))]
-      (let [workflow (postgres/get-table tx items)]
-        (run! (comp (partial update! tx) submit!) workflow)))))
+      (try
+        (let [workflow (postgres/get-table tx items)]
+          (jdbc/update! tx :workload {:started now} ["uuid = ?" uuid])
+          (run! (comp (partial update! tx) submit!) workflow))
+        (catch ExceptionInfo e (throw e))))))
```
ehigham, line 54: `catch` then `throw`???

src/zero/module/wl.clj
```diff
@@ -32,9 +33,10 @@
   "Use transaction TX to update _WORKLOAD statuses."
   [tx {:keys [cromwell items] :as _workload}]
   (let [env (@get-cromwell-wgs-environment cromwell)]
-    (->> items
-         (postgres/get-table tx)
-         (run! (partial maybe-update-workflow-status! tx env items)))))
+    (try
+      (let [workflows (postgres/get-table tx items)]
+        (run! (partial maybe-update-workflow-status! tx env items) workflows))
+      (catch ExceptionInfo (throw (ex-info "Error updating workload status" {:cause "no-workflows-found"}))))))
```
ehigham: What if the database is inaccessible? What if something goes wrong in `maybe-update-workflow-status!`? The `cause` might not be "no workflows found"
This message might be a red herring in the future - maybe we should wrap the underlying exception and add context:
```
(try
  (fetch-table-and-update)
  (catch Exception cause
    (throw (exec-info "Error updating worklflow status" {} cause))))

File: src/zero/module/wl.clj [hunk 6]
```diff
@@ -98,6 +99,9 @@
...
                               ["id = ?" id])))]
-      (let [workflows (postgres/get-table tx items)
-            ids-uuids (map submit! workflows)]
-        (run! (partial update! tx) ids-uuids)))))
+      (try
+        (let [workflows (postgres/get-table tx items)
+              ids-uuids (map submit! workflows)]
+          (jdbc/update! tx :workload {:started now} ["uuid = ?" uuid])
+          (run! (partial update! tx) ids-uuids))
+        (catch ExceptionInfo e (throw e))))))
```
ehigham, line 108: `try` and `throw` again?

src/zero/api/handlers.clj
samanehsan: You're right! The default exception handler in the exception-middleware will catch this and throw a 500 error so I removed the try/catch

PR #66, opened 2020-07-21
GH-807 Avoid extra WFL clones

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-807
- WFL cloned itself to put its master HEAD commit hash into `resources/zero/version.edn`. Doing so is not necessary because the current branch's HEAD commit is already in `version.edn` and is more representative of the version of the software anyway.

### Changes

- Modify `util.clj` to have a function that will swallow errors, because...
- `boot.clj` no longer prints out an error message if `git diff-index --quiet` returns -1 (that exit code is the whole point of `--quiet`)
- `boot.clj` modified to exclude its own repo from what it cloned for `version.edn`

### Review Instructions

- On this branch, running `boot build` should exclude `"wfl" "<commit hash>"` from `resources/zero/version.edn`
- The testing behavior is unchanged from master (in other words, the tests run here the same as they will on master because removing that entry from `version.edn` is a no-op with regard to functionality)

### Notes

- Tom mentioned in a comment on the ticket that boot could probably be eliminated entirely. I don't disagree but I don't think that's a change to make before a big deadline (plus, this change is on the road to that anyway). Would require discussion, so I haven't attempted that here.

### DIFF
--- a/src/zero/boot.clj
+++ b/src/zero/boot.clj
@@ -64,18 +64,23 @@
 
 (defn clone-repos
   "Return a map of zero/the-github-repos clones in a :tmp directory.
-   Delete the :tmp directory at some point."
+   Delete the :tmp directory at some point.
+
+   Specifically omits [[zero/the-name]]'s repo since it isn't needed
+   for the version."
   []
-  (let [tmp (str "CLONE_" (UUID/randomUUID))]
+  (let [tmp (str "CLONE_" (UUID/randomUUID))
+        the-github-repos-no-wfl (dissoc zero/the-github-repos zero/the-name)
+        the-other-github-repos-no-wfl (dissoc zero/the-other-github-repos zero/the-name)]
     (letfn [(clone [url] (util/shell-io! "git" "-C" tmp "clone"
                                          "--config" "advice.detachedHead=false" url))]
       (io/make-parents (io/file tmp "Who cares, really?"))
       (try
-        (run! clone (vals zero/the-github-repos))
+        (run! clone (vals the-github-repos-no-wfl))
         (catch Exception _
-          (run! clone (vals zero/the-other-github-repos)))))
+          (run! clone (vals the-other-github-repos-no-wfl)))))
     (into {:tmp tmp}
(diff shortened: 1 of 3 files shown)


src/zero/boot.clj
```diff
@@ -64,18 +64,23 @@
...
   "Return a map of zero/the-github-repos clones in a :tmp directory.
-   Delete the :tmp directory at some point."
+   Delete the :tmp directory at some point.
+
+   Specifically omits [[zero/the-name]]'s repo since it isn't needed
+   for the version."
   []
-  (let [tmp (str "CLONE_" (UUID/randomUUID))]
+  (let [tmp (str "CLONE_" (UUID/randomUUID))
+        the-github-repos-no-wfl (dissoc zero/the-github-repos zero/the-name)
```
ehigham, line 73: are `the-github-repos` and `the-other-github-repos` used anywhere else? Maybe you can just remove wfl from them?

jack-r-warren: Yeah wfl is specifically pulled out of `the-github-repos` elsewhere in `boot.clj`. `the-other-github-repos` isn't used anywhere else but it has the exact same repos as `the-github-repos` with a different URL scheme used in case of a failure with the first. I suppose this means we're keeping a backup URL for wfl that we wouldn't ever use, but there's a part of me that thinks keeping the parallelism between the two lists is more helpful than molding the lists to the behavior of this function.

jack-r-warren: Update: did more research and the backup URLs don't work and aren't supposed to as far as I can tell from GitHub's documentation. Given that I think they'd only help in the case of weird DNS issues on GitHub's raw domain (?) I removed the entire list of backup URLs. I also removed the try/catch around it in the versioning function so in case GitHub does go completely down we won't swallow the errors (same behavior as before).

jack-r-warren: Okay, last update. The backup URLs *are* used, exclusively in GitHub actions it seems, for accessing the private repos without other overhead. I documented that and re-organized the data structure to better represent it.

PR #67, opened 2020-07-22
[GH-764] Add WFL logging

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-764
- WFL needs logging. This adds it and cleans up the existing logging systems.

### Changes

- Make `deps.edn` and `build.boot` reference the same logging dependencies
- Add dependencies on `clojure.tools.logging` and SLF4J
- Add smoke-ish unit test checking the runtime dependency loading
- Add `docs/logging.md` documenting logging standards, usage, and how the dependency resolution is working (and why we are using `clojure.tools.logging`, SLF4J, and Log4j 2 all at once--I promise there's good reason and it results in a clean interface for us)
- Make Log4j 2's default behavior explicit by adding a equivalent-to-default config file at `resources/log4j2.xml` (if we have to change something, this'll save a bit of time rather than figuring out how/where to override the defaults)

- The `println`ing to stderr/stdout throughout WFL has been replaced with proper logging statements
- Program arguments are specifically logged
- Endpoint server returning a 400 failure is logged
- Added `jdbc.clj` to be a wrapper around clojure.java.jdbc that logs necessary commands
  - Comments and argument structure copied from Clojure source so quickdocs etc still function
- Changes all imports of `clojure.java.jdbc` to instead use `zero.jdbc`
- Removes dependency on liquibase (and moves related comments into dev-process.md)
- Makes Jetty be quiet by excluding everything non-error
- DB arguments in logging is now scrubbed of any `:password` fields

### Review Instructions

- The new unit tests should work for folks, can run them specifically with `clojure -A:test unit --focus zero.unit.logging-test`
- You can test logging by just "using" WFL. For example, run a local Postgres server and run WFL against it. From the WFL's base directory:
```bash
boot build
pg_ctl -D /usr/local/var/postgresql@11 start
createdb wfl
liquibase --classpath=$(clojure -Spath) --url=jdbc:postgresql:wfl --changeLogFile=database/changelog.xml --username=$USER update
psql wfl
./ops/server.sh
```
```bash
# Curl endpoints or do whatever
curl "http://localhost:8080/api/v1/workload" -H 'Accept: application/json'
```
You'll notice some nice silence from Jetty and some of WFL's own logging of things like entrypoint arguments and JDBC calls.


resources/log4j2.xml
```diff
@@ -0,0 +1,19 @@
...
+        <Console name="Console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+        </Console>
+    </Appenders>
+    <Loggers>
+        <Root level="error">
+            <AppenderRef ref="Console"/>
+        </Root>
+    </Loggers>
+</Configuration>
```
rexwangcc, line 19:
```suggestion
</Configuration>

```

docs/md/logging.md
```diff
@@ -0,0 +1,68 @@
...
+An example of usage is in `test/zero/unit/logging_test.clj`. 
+That file doesn't actually test our code, but rather our dependencies:
+it helps catch issues that might arise from clojure.tools.logging (or SLF4J, which
+it currently delegates to) having trouble finding a logging implementation via
+service loaders and the like. It isn't bulletproof--it is theoretically possible
+for tests to use a different classpath than other execution--but it is a good smoke
+test that in that any failures there would require remediation in our own codebase.
+
+## Under the Hood
+WFL's logging works as follows:
```
ehigham, line 44: This a bit verbose and repetitive. I wouldn't go into so much detail and try to avoid repeating yourself.

docs/md/logging.md
```diff
@@ -0,0 +1,68 @@
...
+
+1. clojure.tools.logging is imported and used directly
+   - Why clojure.tools.logging? It is Clojure-native so it has intuitive syntax
+   - Why not wrap or delegate to it? It is written in macros so delegation
+     is messier, the macros themselves just delegate to other logging
+     implementations anyway, and its interface is so generic that we'd be
+     repeating it very closely if we were to wrap it
```
ehigham, line 51: This is what makes wrapping such an easy task.
But not wrapping, you've injected a external dependency into almost every single file in workflow launcher. It's quite possible that we may want to change this in the future.
As a general software engineering tip, try to minimise how much of your code has to know about the dependencies - it's much easier to change code in one place than several.

resources/log4j2.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+This file configures Log4j 2.
+Its syntax is given by https://logging.apache.org/log4j/2.x/manual/configuration.html
+-->
+<Configuration status="WARN">
+    <Appenders>
+        <Console name="Console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
```
ehigham, line 9: Just use ISO 8601 long format.

src/zero/jdbc.clj
```diff
@@ -0,0 +1,270 @@
+(ns zero.jdbc
+  "clojure.tools.logging wrapping for clojure.java.jdbc"
+  (:require [clojure.java.jdbc :as jdbc]
+            [clojure.tools.logging.readable :as log]))
+
+(defmacro query
+  "Logged alias for [[clojure.java.jdbc/query]]:
+
+  Given a database connection and a vector containing SQL and optional parameters,
```
ehigham, line 9: Do not copy and paste from the clojure doc. Regardless of how i feel about copying work without accreditation, this is yet another thing we have to maintain. IT WILL GO OUT OF DATE.

src/zero/jdbc.clj
```diff
@@ -0,0 +1,270 @@
...
+  a single row, perform an insert.
+  When inserting a row as a map, the result is the database-specific form of the
+  generated keys, if available (note: PostgreSQL returns the whole row).
+  When inserting a row as a list of column values, the result is the count of
+  rows affected (1), if available (from getUpdateCount after executeBatch).
+  The row map or column value vector may be followed by a map of options:
+  The :transaction? option specifies whether to run in a transaction or not.
+  The default is true (use a transaction). The :entities option specifies how
+  to convert the table name and column names to SQL entities."
+  ([db table row] `(do
```
ehigham, line 142: Can you just foward the arguments?
```
(defn foo [& args]
     (log/info "called foo with arguments " args)
     (apply foo_impl args))
```

docs/md/logging.md
jack-r-warren: I agree that wrapping is often a useful pattern, but there's a reason we don't bother wrapping `clojure.string` or `clojure.pprint`. The key thing to understand about `clojure.tools.logging` is it is literally just a wrapper itself that abstracts over more logging frameworks than even SLF4J supports.

Even if Broad decided to develop its own JVM logging library from scratch, a trivial SLF4J "impl" (like what Log4j2 uses right now) would make it work with `clojure.tools.logging`. I'm having trouble coming up with a scenario where the Clojure wrapper built for exactly this purpose doesn't offer enough flexibility and we want to wrap it again.

resources/log4j2.xml
jack-r-warren: Done

src/zero/jdbc.clj
jack-r-warren: The attribution aspect slipped my mind, deleted.

src/zero/jdbc.clj
jack-r-warren: Not in all cases, because macros can't be applied to. Capturing arguments manually also helps inform useful logging messages, especially for the `insert` and related functions. Capturing arguments manually also fills in useful tooltip information in IDEs.

A lot of what I did in this file was frankly to get around the fact that Clojure documentation can't link to other functions, which I feel is a really significant issue in the toolchain. Copying documentation and manually capturing arguments try to help get around the fact that a simple delegate pattern loses all context in Clojure. There's nothing I can do to change Clojure but putting in the work to manually capture arguments improves the tooltips and the logging.

I'm not worried about maintenance costs for this because clojure.java.jdbc changes would've broken our existing code, and here we at least now have a wrapper we can use to adapt.

src/zero/jdbc.clj
ehigham: I think it's fine if you do credit the original work, but I'd more likely use their summary and link to their doc for examples.

src/zero/jdbc.clj
ehigham: The ability to maintain a stable API is a really good point that i didn't think of. `clojure/jdbc` hasn't yet reached v1 yet so perhaps you're right.

docs/md/logging.md
ehigham: Yes indeed, you're not going to wrap `string` or any other built in data type. I was kinda thinking someone might raise this lol! I really wouldn't group logging capabilities with core language types like `string` though.

In this case, I meant switching to something like timbre. I guess this is unlikely.

docs/md/logging.md
jack-r-warren: Added a summary at the top of the document.

I think I might be unique in terms of my want of documentation in WFL. I've found WFL to be extremely hard to get up to speed with, and I think it is a shame that the most useful resource I have found for understanding it has been looking at Tom's complaint tickets in the backlog--he mentions things that are superfluous and that begins to knock on the "why" of how on earth this software is built.

I'm happy to remove what I've written here and put it on my desktop instead--I'm going to maintain it regardless.

ehigham: Yeah wfl does have a number of things that make me scratch my head too! There are a number of things I don't understand yet.
Feel free to document, by all means. I think stuff like this is really useful but it could be a bit more concise. If you're exposing `clojure.tools.logging`, I would state we use that and its configuration with a brief explanation for those choices.

docs/md/logging.md
jack-r-warren: Totally fair. Timbre supports [both clojure.tools.logging and SLF4J](https://github.com/ptaoussanis/timbre#a-pure-clojurescript-logging-library), so we could switch to it super easily.

docs/md/logging.md
jack-r-warren: (Updated the document to make it more brief after a conversation with Ed last week)

PR #68, opened 2020-07-23
always create a new workload before starting it

ehigham:
### Purpose
Fix https://broadinstitute.atlassian.net/browse/GH-843
I wrote a bug in the `test-start-workload` test point where I got the first-unstarted workload and tried to start it.
The correct behaviour should be to always create the workload before starting it.

### Changes
Always create a new workload before starting it. Note this is still failing due to https://broadinstitute.atlassian.net/browse/GH-845


ehigham:
Failure is known issue: https://broadinstitute.atlassian.net/jira/software/projects/GH/boards/530/backlog?selectedIssue=GH-847

PR #70, opened 2020-07-24
[GH-837] Add cloud function to call WFL

samanehsan:
### Purpose
Tell WFL to process arrays data that has been uploaded to the cloud by PTC: https://broadinstitute.atlassian.net/browse/GH-837

### Changes
This function will:
1) Download the `ptc.json` file for the sample
2) Check that all gs files in `ptc.json` have been uploaded
3) Get the arrays workload uuid from WFL
4) Append the sample to WFL (use the contents of ptc.json as the payload, replacing "uuid:null" with the workload uuid from the previous step)

### Review Instructions

Questions:
- ~The google metadata API is only accessible from within a compute VM, so I included another auth method that I was using for local testing (controlled by the "DEFAULT_CREDENTIALS" flag). Would it be useful to leave that in?~ I removed this to simplify the code!

### DIFF
--- a/cloud_function/README.md
+++ b/cloud_function/README.md
@@ -13,5 +13,5 @@ Run `bash deploy.sh $GCLOUD_PROJECT> $TRIGGER_BUCKET`
 
 Testing
 -------
-1) Create a virtual python3 environment and install `dev-requirements.txt`
+1) Create a virtual python3 environment and install with `pip install -r dev-requirements.txt`
 2) Run the unit tests: `pytest tests/unit_tests.py`
--- a/cloud_function/README.md
+++ b/cloud_function/README.md
@@ -1,0 +1,17 @@
+Overview
+--------
+When the push to cloud service uploads arrays data to a GCS bucket, it will trigger execution of this cloud function.
+The function checks the parent directory of the uploaded file `"gs://{bucket}/{chipwell_barcode}/{analysis_version}/"`
+for a manifest file named `ptc.json`. If all of the files listed in the manifest have been uploaded, a request is sent 
+to the WFL API to start processing this sample.
+
+
+Deployment
+---------
+Run `bash deploy.sh $GCLOUD_PROJECT> $TRIGGER_BUCKET`
+
+
+Testing
+-------
+1) Create a virtual python3 environment and install `dev-requirements.txt`
+2) Run the unit tests: `pytest tests/unit_tests.py`
--- a/cloud_function/README.md
+++ b/cloud_function/README.md
@@ -13,5 +13,10 @@ Run `bash deploy.sh $GCLOUD_PROJECT> $TRIGGER_BUCKET`
 
 Testing
 -------
-1) Create a virtual python3 environment and install with `pip install -r dev-requirements.txt`
-2) Run the unit tests: `pytest tests/unit_tests.py`
+1) Create a virtual python3 environment
+2) Install requirements with `pip install -r dev-requirements.txt`
+3) Run the unit tests:
+```
+WFL_URL="https://workflow-launcher.gotc-dev.broadinstitute.org"
+WFL_URL=${WFL_URL} pytest tests/unit_tests.py
+```
--- a/cloud_function/main.py
+++ b/cloud_function/main.py
@@ -5,8 +5,10 @@ from google.cloud import storage
 from google.cloud import exceptions
 
 
-service_account = os.environ.get('FUNCTION_IDENTITY')  # Set by the cloud fn
-wfl_url = os.environ.get("WFL_URL")
+_SERVICE_ACCOUNT = os.environ.get('FUNCTION_IDENTITY')  # Set by the cloud fn
+WFL_URL = os.environ.get("WFL_URL")
+
+assert WFL_URL is not None, "WFL_URL is not set"
 
 def get_auth_headers():
     scopes_list = ["https://www.googleapis.com/auth/cloud-platform",
@@ -30,7 +32,7 @@ def get_manifest_path(object_name):
 
 def get_or_create_workload(headers, cromwell_url):
     payload = {
-        "creator": service_account,
+        "creator": _SERVICE_ACCOUNT,
         "cromwell": cromwell_url,
         "input": "aou-inputs-placeholder",
         "output": "aou-ouputs-placeholder",
@@ -39,7 +41,7 @@ def get_or_create_workload(headers, cromwell_url):
         "items": [{}]
     }
     response = requests.post(
-        url=f"{wfl_url}/api/v1/exec",
+        url=f"{WFL_URL}/api/v1/exec",
         headers=headers,
         json=payload
     )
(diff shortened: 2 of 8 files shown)


cloud_function/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+
+Deployment
+---------
+Run `bash deploy.sh $GCLOUD_PROJECT> $TRIGGER_BUCKET`
+
+
+Testing
+-------
+1) Create a virtual python3 environment and install `dev-requirements.txt`
```
rexwangcc, line 16:
```suggestion
1) Create a virtual python3 environment and install with `pip install -r dev-requirements.txt`
```

cloud_function/main.py
```diff
@@ -0,0 +1,127 @@
+import os
+import json
+import requests
+from google.cloud import storage
+from google.cloud import exceptions
+
+project = os.environ.get("GCP_PROJECT", "broad-gotc-dev")
+wfl_url = os.environ.get("WFL_URL", "https://workflow-launcher.gotc-dev.broadinstitute.org")
+default_credentials = os.environ.get("DEFAULT_CREDENTIALS", False)
```
rexwangcc, line 9: Since this function is going to run automatically, would it make more sense to fail the function instead of using dev by default?  e.g. don't set a default value for above and have something like `assert project is not None`?

cloud_function/main.py
```diff
@@ -0,0 +1,127 @@
...
+    access_token = r.json()['access_token']
+    headers = {
+        'Authorization': 'Bearer {}'.format(access_token)
+    }
+    return headers
+
+def get_manifest_path(object_name):
+    parts = object_name.split("/")
+    chipwell_barcode = parts[0]
+    analysis_version = parts[1]
```
rexwangcc, line 42:
```suggestion
    chipwell_barcode, analysis_version = object_name.split("/")
```

cloud_function/main.py
```diff
@@ -0,0 +1,127 @@
...
+    return "/".join([chipwell_barcode, analysis_version, "ptc.json"])
+
+def get_or_create_workload(headers, cromwell_url):
+    payload = {
+        "creator": "wfl-non-prod@broad-gotc-dev.iam.gserviceaccount.com",
```
rexwangcc, line 47: I assume once the deploy script of this function supports prod, this payload will also needs to be parameterized? (use different creator values for dev and prod?)

cloud_function/main.py
```diff
@@ -0,0 +1,127 @@
...
+            print(f"File not found: {gs_url}")
+            return
+
+    if upload_complete:
+        sample_alias = notification.get('sample_alias')
+        print(f"Sample upload completed for {sample_alias}")
+        cromwell_url = input_data.get('cromwell')
+        workload_uuid = get_or_create_workload(headers, cromwell_url)
+        workflow_ids = update_workload(headers, workload_uuid, input_data)
+        print(f"Started cromwell workflows: {workflow_ids}")
```
rexwangcc, line 127: I'm not sure about Cloud Functions logging features, but I can see we/hOps will be interested in being able to filter different information this cloud function logs, such as: what are the successfully submitted workflows (UUID), their chipwell-barcodes, which samples are throwing `File not found Erros`, etc. Not sure if we should do it here or as a follow up.

jack-r-warren:
> LGTM in general. I have a question about how this function will be triggered in the cloud. Does it get triggered whenever a new object is deposited to the bucket? If so, a lot of file not found errors will be thrown, how can we make sure to distinguish them from the real exceptions?

I'll look at it with fresh eyes tomorrow, but I think as written right now if it can't find a file (manifest or something referred to in a manifest) it prints and just exits rather than actually throwing an exception. If by nothing other than the exit code (?) we should be able to distinguish that normal operation from a real exception (sorry I'm a bit out of practice with Python)

cloud_function/main.py
samanehsan: There would be more than two items in the "parts" array so this would throw an error. 
(e.g. if the object name is "chipwell_barcode/analysis_version/some_data.json")

samanehsan: Also I realized there is a "/" in front of the file name in the event payload so I need to fix this parsing 🤦‍♀️

rexwangcc: > There would be more than two items in the "parts" array so this would throw an error.
> (e.g. if the object name is "chipwell_barcode/analysis_version/some_data.json")

Oh I see, my bad

cloud_function/main.py
samanehsan: Ah good catch! Turns out the cloud function VM comes with the "FUNCTION_IDENTITY" env variable set to the service account email so we don't have to pass it in

cloud_function/main.py
```diff
@@ -0,0 +1,127 @@
...
+    return "/".join([chipwell_barcode, analysis_version, "ptc.json"])
+
+def get_or_create_workload(headers, cromwell_url):
+    payload = {
+        "creator": service_account,
+        "cromwell": cromwell_url,
+        "input": "aou-inputs-placeholder",
+        "output": "aou-ouputs-placeholder",
+        "pipeline": "AllOfUsArrays",
+        "project": "gotc-dev",
```
samanehsan, line 53: @tbl3rd @rexwangcc would this field be in the "ptc.json" file or should we set it as an environment variable when deploying the function? 🤔

cloud_function/main.py
samanehsan: That's a good point! Instead of `"Sample upload completed for {sample_alias}"` it seems like `"Sample upload completed for {chipwell_barcode}-{analysis_version}" would be more useful? Also we can improve this in a follow-up PR too based on feedback from hOps!

samanehsan:
> > LGTM in general. I have a question about how this function will be triggered in the cloud. Does it get triggered whenever a new object is deposited to the bucket? If so, a lot of file not found errors will be thrown, how can we make sure to distinguish them from the real exceptions?
> 
> I'll look at it with fresh eyes tomorrow, but I think as written right now if it can't find a file (manifest or something referred to in a manifest) it prints and just exits rather than actually throwing an exception. If by nothing other than the exit code (?) we should be able to distinguish that normal operation from a real exception (sorry I'm a bit out of practice with Python)

+1 to what @jack-r-warren said! This function will be trigged each time an object is created or overridden (the [object finalize event](https://cloud.google.com/functions/docs/calling/storage#object_finalize)). We're assuming that if the function doesn't find a file that it needs to submit the sample, PTC hasn't uploaded it yet so it doesn't throw an error.

cloud_function/main.py
tbl3rd: This should be able to use the ptc.json as the request itself.

samanehsan: I thought the contents of `ptc.json` were for the `/api/v1/append_to_aou` endpoint? This is to create/get the workload to update.

tbl3rd: You're right.

samanehsan: Ok! So I replaced the hard-coded project here with the "environment" defined in ptc.json

PR #72, opened 2020-07-28
include copyfile.wdl in jar. Extract to a temporary file before sending

ehigham:
### Purpose
`copyfile.wdl` was not being deployed. This was causing a `FileNotFoundException` from being thrown in the `test-copyfile-workflow` test point.

### Changes
Ship `copyfile.wdl` in the wfl JAR
When the `start-workload!` function is called, extract the bundled wdl to a temporary file and send that to Cromwell.

### Review Instructions
It seems weird to extract this to a temporary file when this is read almost immediately afterwards by `zero.service.cromwell/post-workflow`. `zero.service.cromwell/submit-workflow` requires the wdl to be a file (see `zero.service.cromwell/partify-workflow`). Not sure how to get around this. Suggestions welcome.


src/zero/module/copyfile.clj
```diff
@@ -18,12 +17,12 @@
...
   [environment workflow]
   (cromwell/submit-workflow
-   environment
-   (io/file (:top workflow-wdl))
-   nil
-   (-> workflow (select-keys [:src :dst]) (util/prefix-keys pipeline))
-   (util/make-options environment)
-   {}))
+    environment
+    (util/extract-resource (:top workflow-wdl))
```
jack-r-warren, line 21:
```suggestion
    (clojure.java.io/file (.getFile (clojure.java.io/resource (:top workflow-wdl))))
```
That'll create a file out of something on the classpath (so "wdl/copyfile.wdl" here) without copying it

ehigham: This test is a bit rubbish and isn't really testing the real runtime behaviour - when this file is bundled in the jar, your code won't work unfortunately (i tried it myself :) ).

jack-r-warren: Fair enough. It is weird that the resources file isn't available (other things from resources load fine in the jar, like log4j2.xml in my logging branch) but I think what we have here works perfectly fine and we have little reason to spend more time tweaking it.

ehigham: Right, but I don't think you're reading from that file yourself? I guess log4j looks for that file on the classpath and handles it?

PR #77, opened 2020-07-29
[GH-764] Add better endpoint logging

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-764
- This is something of a followup to https://github.com/broadinstitute/wfl/pull/67
- @ehigham mentioned that we should've clarified "endpoint" instead of "entrypoint" on GH-764. I went ahead and added a few lines to clean up the logging and record endpoint activity.

### Changes

- Memoize the JDBC db argument printing to avoid clutter
- Record the parameters given to endpoints in zero.api.handlers

### Review Instructions

- To see it for yourself:
```bash
boot build
pg_ctl -D /usr/local/var/postgresql@11 start
createdb wfl
liquibase --classpath=$(clojure -Spath) --url=jdbc:postgresql:wfl --changeLogFile=database/changelog.xml --username=$USER update
./ops/server.sh
```

```bash
# Curl endpoints or do whatever
curl "http://localhost:8080/api/v1/workload" -H 'Accept: application/json'
```

- Main thing to review here is if I'm recording the right things about the endpoints parameters or if I should record less. That can be done by just looking at the changes to `src/zero/api/handlers.clj`

### DIFF
--- a/src/zero/api/handlers.clj
+++ b/src/zero/api/handlers.clj
@@ -1,17 +1,26 @@
-*(ns zero.api.handlers
-  "Define handlers for API endpoints"
-  (:require [clojure.string :as str]
-            [clojure.tools.logging :as log]
-            [clojure.tools.logging.readable :as logr]
-            [ring.util.response :as response]
-            [zero.module.aou :as aou]
-            [zero.module.copyfile :as cp]
-            [zero.jdbc :as jdbc]
-            [zero.module.wgs :as wgs]
-            [zero.module.wl :as wl]
-            [zero.service.cromwell :as cromwell]
-            [zero.service.postgres :as postgres]
-            [zero.zero :as zero]))
+* (ns zero.api.handlers
+    "Define handlers for API endpoints"
+    (:require [clojure.string :as str]
+              [clojure.tools.logging :as log]
+              [clojure.tools.logging.readable :as logr]
+              [ring.util.response :as response]
+              [zero.module.aou :as aou]
+              [zero.module.copyfile :as cp]
+              [zero.jdbc :as jdbc]
+              [zero.module.wgs :as wgs]
+              [zero.module.wl :as wl]
+              [zero.service.cromwell :as cromwell]
+              [zero.service.postgres :as postgres]
+              [zero.zero :as zero]))
+
+(defmacro ^:private print-handler-error
+  "Run BODY, printing any exception rises through it."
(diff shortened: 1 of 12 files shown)


src/zero/api/handlers.clj
```diff
@@ -43,14 +46,16 @@
 
 (defn status-counts
   "Get status counts for environment in REQUEST."
-  [{:keys [parameters] :as _request}]
+  [{:keys [parameters] :as _request}]0
```
rexwangcc, line 49:
```suggestion
  [{:keys [parameters] :as _request}]
```

src/zero/api/handlers.clj
```diff
@@ -125,6 +136,7 @@
 (defn post-start
   "Start the workloads with UUIDs in REQUEST."
   [request]
+  (log/info "post-start endpoint parameters:" (:parameters request))
```
rexwangcc, line 139: We might want to double check there are no credentials in the parameters, for example, auth tokens.

src/zero/jdbc.clj
```diff
@@ -3,80 +3,90 @@
...
                                   (jdbc/insert! ~db ~table ~cols ~values ~opts))))
 
 (defmacro with-db-transaction
   "Logged alias for [[clojure.java.jdbc/with-db-transaction]]"
   [binding & body]
   `(let [id# (rand-int 10000)]
-     (log/info "JDBC SQL transaction" id# "started to" (clean-db ~(second binding)))
+     (log/info "JDBC SQL transaction" id# "started to" (format-db ~(second binding)))
      (jdbc/with-db-transaction ~binding ~@body)
      (log/info "JDBC SQL transaction" id# "ended")))
```
rexwangcc, line 91: See #78

src/zero/api/handlers.clj
jack-r-warren: I wish my linting setup worked better, goodness 🙁

src/zero/jdbc.clj
jack-r-warren: Merged your branch into this one to apply the fix

src/zero/api/handlers.clj
jack-r-warren: I think I may have resolved this, would love another quick pass through to see if more narrowly logging is sufficient

PR #84, opened 2020-07-31
[GH-847] Remove unstable message test

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-847
- This test fails sometimes and we don't know why but also we don't care that much.

### Changes

- Since the pub/sub code no longer has a planned use, comment what it was originally for
- Comment out the test that sometimes fails

### Review Instructions

- The PR will run the unit tests. There's more info in the ticket, we know that this is the test that causes transient failures.
- **Big question:** should we delete the entire test file and source code file? That seems like a bit of a waste but it might also be a bit cleaner, I don't know what folks' thoughts on that are.


src/zero/service/pubsub.clj
```diff
@@ -7,6 +7,15 @@
             [zero.once         :as once])
   (:import [java.util Base64]))
 
+(comment
```
rexwangcc: I'd make this block of comments a multi-line `;;` comment simply because most IDEs can grey out `;;` to make them less distractive, but this is totally personal preference, so it's up to you!

rexwangcc: NVM, this is on an old commit.

PR #87, opened 2020-08-04
[GH-875] Cloud SQL Instance name inference

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-875
- We need cli.py to be smarter about how it finds the name of the Cloud SQL instance (as in, it needs to be not hardcoded)

### Changes

- Add a parameter to the `deploy` command to specify the instance (there was already one for `connect`)
- Remove the existing default (it references a soon-to-be-deleted orphaned resource)
- If the parameter isn't passed, look for a single Cloud SQL instance in the project with a [present and truthy 'wfl' tag](https://github.com/broadinstitute/terraform-shared/pull/122)
   - Bonus: this inference happens early during `deploy` execution so `deploy --dry-run` will print out the inferred name without doing anything to it

### Review Instructions

-  If you specify the name of the old orphaned Cloud SQL instance ("zero-postgresql") you should be able to deploy right now to gotc-dev.
- Blocked by [this PR](https://github.com/broadinstitute/terraform-shared/pull/122), you can `deploy --dry-run -e aou` to see it infer the AoU DB instance
   - I recommend we use this as the main test before merging, rather than waiting for the below. The below ones are likely to fail for reasons relating to the blocking tickets, not the task at hand.
- Blocked by the above referenced PR and [also this other one](https://github.com/broadinstitute/gotc-deploy/pull/111), you can `deploy --dry-run` to see it infer the gotc-dev DB instance or you can `deploy` to see it deploy to gotc-dev
- Blocked by the above referenced PR and [also this other (different) one](https://github.com/broadinstitute/gotc-deploy/pull/109), you can run `deploy -e aou` to see it deploy to AoU

### DIFF
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -148,21 +148,8 @@ def set_up_k8s(project: str, namespace: str, cluster: str, zone: str):
     success(f"[✔] Set namespace to {namespace}.")
 
 
-def run_cloudsql_proxy(project: str, cloudsql_instance_name: str = None):
+def run_cloudsql_proxy(project: str, cloudsql_instance_name):
     """Connect to a google cloud sql instance using the cloud sql proxy."""
-    if cloudsql_instance_name is None:
-        info("=> Finding cloud_sql_proxy instance")
-        gcloud_command = f"gcloud --format=json --project {project} sql instances list"
-        instances = json.loads(subprocess.check_output(gcloud_command, shell=True, encoding='utf-8').strip())
-        instance_names = instances\
-            .filter(lambda i: i.get("settings", {}).get("labels", {}).get("wfl", False))\
-            .map(lambda i: i["name"])
-        if len(instance_names) == 1:
-            cloudsql_instance_name = instance_names[0]
-            info(f"   Found instance: {cloudsql_instance_name}")
-        else:
-            error(f"   Didn't find one instance: {instance_names}")
-            exit(1)
     info("=> Running cloud_sql_proxy")
     token = subprocess.check_output("gcloud auth print-access-token", shell=True, encoding='utf-8').strip()
     instance_command = " ".join([f"gcloud --format=json sql --project {project}",
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -84,7 +84,7 @@ def register(func: Callable) -> Callable:
 def shell(command: str):
     """Run COMMAND in a subprocess."""
     info(f"Running: {command}")
-    return subprocess.check_call(command, shell=True)
+    return subprocess.check_output(command, shell=True, encoding='utf-8').strip()
 
 
 def shell_unchecked(command: str):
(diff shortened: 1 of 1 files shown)


ops/cli.py
```diff
@@ -160,6 +160,21 @@ def run_cloudsql_proxy(project: str, cloudsql_instance_name):
...
     return subprocess.check_output(docker_command, shell=True, encoding='utf-8').strip()
 
+def infer_cloudsql_name(project: str):
+    """Try to find a single cloudsql instance with a truthy 'wfl' tag."""
+    info("=> Finding cloud_sql_proxy instance")
+    gcloud_command = f"gcloud --format=json --project {project} sql instances list"
+    instances = json.loads(subprocess.check_output(gcloud_command, shell=True, encoding='utf-8').strip())
+    instance_names = instances \
+        .filter(lambda i: i.get("settings", {}).get("labels", {}).get("wfl", False)) \
+        .map(lambda i: i["name"])
```
rexwangcc, line 170: This looks like Java-ish dot chain style, and I really doubt if this would work (assuming `instance` is a `Union[List, Dict]` type) 🤔 

In Python, `filter` and `map` takes the `func` as the 1st argument and the collection as the second one, and return lazy generators. If `map(filter(func, coll))` is too ugly here, I'd use a custom generator that composes the 2 operations, or look into `functools` (for something like `partialmethod`)

ops/cli.py
```diff
@@ -160,6 +160,21 @@ def run_cloudsql_proxy(project: str, cloudsql_instance_name):
                         f'-token="{token}" -instances="{instance}=tcp:0.0.0.0:5432"'])
     return subprocess.check_output(docker_command, shell=True, encoding='utf-8').strip()
 
+def infer_cloudsql_name(project: str):
+    """Try to find a single cloudsql instance with a truthy 'wfl' tag."""
+    info("=> Finding cloud_sql_proxy instance")
+    gcloud_command = f"gcloud --format=json --project {project} sql instances list"
```
samanehsan, line 166: You can add `--filter="settings.userLabels.wfl=false"` to filter by label so that you don't have to do that with python: https://cloud.google.com/sdk/gcloud/reference/topic/filters

ops/cli.py
jack-r-warren: Yeah this doesn't work at all, I really wish IntelliJ didn't autocomplete it for me, feels like I'm fighting against the computer.

jack-r-warren: Okay I tried to use list comprehension to do it, hopefully that works

ops/cli.py
jack-r-warren: Welp. Thank you.

jack-r-warren:
As a heads up to folks that were kind enough to review this, there's a superseding PR that I'll make Monday morning that couples these changes with a bit more of a rework to fit with the wfl-instance changes on the way from the terraform-shared and gotc-dev repos. That new PR is also built on top of Ed's makefile work so we won't have incompatibilities there.

jack-r-warren:
Closing in favor of #93

PR #93, opened 2020-08-10
[GH-877] New cli.py for new wfl-instance Terraform module

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-877
- Supersedes #87 

### Changes

- New usage:
```
usage: cli.py [-h] [-d] COMMAND ENV INSTANCE [...]

deploy or connect to WFL infrastructure

positional arguments:
  COMMAND               one of [info, connect, deploy]
  ENV                   specify 'gotc-deploy/deploy/{ENV}' with one of [gotc-
                        dev, gotc-prod, aou]
  INSTANCE              specify the ID of the instance to use in the
                        environment

optional arguments:
  -h, --help            show this help message and exit
  -d, --dry-run         exit before COMMAND makes any remote changes
  -v VERSION, --version VERSION
                        specify the version to use instead of the 'version'
                        file
  --project PROJECT     specify the GCP project name instead of 'broad-{ENV}'
  --cloud-sql-name CLOUD_SQL_NAME
                        specify the Cloud SQL name instead of inferring from
                        labels
  --cluster-zone CLUSTER_ZONE
                        specify the zone for the GKE cluster instead of
                        inferring from CLUSTER_NAME
  --cluster-name CLUSTER_NAME
                        specify the GKE cluster name instead of inferring from
                        labels
  --cluster-namespace CLUSTER_NAMESPACE
                        specify the K8s namespace instead of '{INSTANCE}-wfl'
```
- Rendering can be directly accessed through its own script:
```
usage: render_ctmpl.py [-h] [-v VERSION] file

positional arguments:
  file                  CTMPL file to be rendered

optional arguments:
  -h, --help            show this help message and exit
  -v VERSION, --version VERSION
                        A specific version to use other than the root's
                        `version` file
```
- Made a util folder with misc.py to hold common logic between both scripts
- Used functions instead of classes

### Review Instructions

- On my branch you can:
  - Run the info command to see it infer arguments
  - Run the connect command (`connect gotc-dev dev`) to be connected to the brand new cloud sql instance in gotc-dev
  - Run deploy dry run (`deploy gotc-dev dev --dry-run`) to see it do everything that doesn't have remote side effects

### DIFF
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -102,7 +102,8 @@ class Config:
         """Look in the cluster list for the cluster's zone."""
         clusters = json.loads(shell(f"gcloud --project {self.project} --format=json "
                                     "container clusters list "
-                                    f"--filter='name={self.cluster_name}'"))
+                                    f"--filter='name={self.cluster_name}'",
+                                    quiet=True))
         if len(clusters) == 1:
             return clusters[0]["zone"]
         else:
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -121,7 +121,7 @@ def configure_kubectl(config: WflInstanceConfig) -> None:
     ctx = f"gke_{config.project}_{config.cluster_zone}_{config.cluster_name}"
     shell(f"kubectl config use-context {ctx}")
     try:
-        namespaces = json.loads(shell("kubectl get namespace -o json", timeout=5))
+        namespaces = json.loads(shell("kubectl get namespace -o json", timeout=10))
         if config.cluster_namespace not in [item["metadata"]["name"] for item in namespaces["items"]]:
             info(f"    Available namespaces in {config.cluster_name}:", plain=True)
             info(shell("kubectl get namespace", quiet=True), plain=True)
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -54,9 +54,9 @@ def validate_cloud_sql_name(config: WflInstanceConfig) -> None:
         instances = json.loads(shell(f"gcloud --project {config.project} --format=json "
                                      "sql instances list"))
         if config.cloud_sql_name not in [i["name"] for i in instances]:
+            error(f"Cloud SQL instance {config.cloud_sql_name} not found")
             info(f"    Available Cloud SQL instances in {config.project}:", plain=True)
             info(shell(f"gcloud --project {config.project} sql instances list", quiet=True), plain=True)
-            raise RuntimeError(f"Cloud SQL instance {config.cloud_sql_name} not found")
         else:
             success(f"Cloud SQL instance {config.cloud_sql_name} exists")
 
@@ -68,9 +68,9 @@ def validate_cluster_name(config: WflInstanceConfig) -> None:
         clusters = json.loads(shell(f"gcloud --project {config.project} --format=json "
                                     f"container clusters list"))
         if config.cluster_name not in [c["name"] for c in clusters]:
+            error(f"GKE cluster {config.cluster_name} not found")
             info(f"    Available clusters in {config.project}:", plain=True)
             info(shell(f"gcloud --project {config.project} container clusters list", quiet=True), plain=True)
-            raise RuntimeError(f"GKE cluster {config.cluster_name} not found")
         else:
             success(f"GKE cluster {config.cluster_name} exists")
 
@@ -123,9 +123,9 @@ def configure_kubectl(config: WflInstanceConfig) -> None:
     try:
         namespaces = json.loads(shell("kubectl get namespace -o json", timeout=10))
         if config.cluster_namespace not in [item["metadata"]["name"] for item in namespaces["items"]]:
+            error(f"Namespace {config.cluster_namespace} not found in {config.cluster_name}")
             info(f"    Available namespaces in {config.cluster_name}:", plain=True)
             info(shell("kubectl get namespace", quiet=True), plain=True)
-            raise RuntimeError(f"Namespace {config.cluster_namespace} not found in {config.cluster_name}")
         else:
             shell(f"kubectl config set-context --current --namespace={config.cluster_namespace}")
             success("Kubernetes configured")
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -282,6 +282,12 @@ class DeployCommand(ConnectCommand):
 
 
 class CLI:
+    command_mapping = {
+        "deploy": DeployCommand,
+        "connect": ConnectCommand,
+        "info": InfoCommand
+    }
+
     def __init__(self):
         """Set up the parser."""
         parser = argparse.ArgumentParser(description="Deploy or connect to WFL infrastructure instances",
@@ -292,9 +298,9 @@ class CLI:
                             help="Force COMMAND to run even if validation fails")
         commands = parser.add_argument_group("COMMAND")
         commands.add_argument("command",
-                              type=lambda s: getattr(sys.modules[__name__], f"{s.title()}Command"),
-                              choices=[DeployCommand, ConnectCommand, InfoCommand],
-                              metavar="{deploy, connect, info}",
+                              type=lambda s: self.command_mapping.get(s),
+                              choices=self.command_mapping.values(),
+                              metavar=f"{{{', '.join(self.command_mapping.keys())}}}",
                               help="Deploy the local build to the instance, "
                                    "connect to the instance's Cloud SQL, "
                                    "or display instance config")
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -177,9 +177,8 @@ class InfoCommand(Config):
     def __init__(self, parsed_args: argparse.Namespace):
         super().__init__(parsed_args)
 
-    # This can't be static if we're to override it in subclasses
-    # noinspection PyMethodMayBeStatic
-    def execute(self) -> int:
+    @staticmethod
+    def execute() -> int:
         return 0
 
 
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -1,26 +1,28 @@
 #!/usr/bin/env python3
-"""WFL Deployment Script
-@rex
+"""
+WFL Deployment Script
 
-requirements: pip3 install gitpython pyyaml
+requirements: pip3 install pyyaml
 
 usage: python3 cli.py -h
 """
-from enum import Enum
-from pathlib import Path
-from typing import Callable
 import argparse
 import json
 import os
 import shutil
 import subprocess
 import sys
-import tempfile
+from enum import Enum
+from typing import Callable, Optional
+
 import yaml
+
+from .render_ctmpl import render_ctmpl
 
 # In case this will need to run on Windows systems
 if sys.platform.lower() == "win32":
     os.system("color")
+
 
 class AvailableColors(Enum):
     GRAY = 90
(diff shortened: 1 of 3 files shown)


ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+
+        success("Deployment is done!")
+        info(shell("kubectl get pods"), plain=True)
 
-        success("[✔] Deployment is done!")
-        shell("kubectl get pods")
         return 0
 
+    @staticmethod
+    def __set_up_helm():
```
ehigham, line 252: Why is this a static method? What's the benefit of this class owning this behaviour rather than just making it a free function?

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+
+    def __helm_deploy_wfl(self, values: str):
+        """Helm deploy using gotc-charts/wfl."""
+        info(f"=>  Deploying to {self.cluster_name} in {self.cluster_namespace} namespace")
+        info("    This must run on a non-split VPN", plain=True)
+        shell(f"helm upgrade wfl-k8s gotc-charts/wfl -f {values} --install")
+        success("WFL deployed")
+
+    @staticmethod
+    def __run_liquibase_migration(db_username: str, db_password: str):
```
ehigham, line 274: Again there's no reason for these to be static methods. I have no idea why they exist in python.

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+        args = self.parser.parse_args()
+        for cmd in ["docker", "gcloud", "git", "helm", "java", "kubectl", "jq"]:
+            if not shutil.which(cmd):
+                error(f"{cmd} is not in PATH. Please install it!")
+        commands = {
+            "info": InfoCommand,
+            "connect": ConnectCommand,
+            "deploy": DeployCommand
+        }
+        config = commands[args.command](args)
```
ehigham, line 333: I look forward to trying to understand the error this generates.

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+                              help="Specify the exact K8s namespace instead of '{INSTANCE}-wfl'")
+        self.parser = parser
+
+    def execute(self) -> int:
+        """Run command."""
+        args = self.parser.parse_args()
+        for cmd in ["docker", "gcloud", "git", "helm", "java", "kubectl", "jq"]:
+            if not shutil.which(cmd):
+                error(f"{cmd} is not in PATH. Please install it!")
+        commands = {
```
ehigham, line 328: Why can't you manage this mapping in the parser? You've got two lists of commands to maintain now.

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+        parser.add_argument("-d", "--dry-run", action="store_true",
+                            help="Prevent COMMAND from enacting changes")
+        parser.add_argument("-f", "--force", action="store_true",
+                            help="Force COMMAND to run even if validation fails")
+        commands = parser.add_argument_group("COMMAND")
+        commands.add_argument("command", choices=["deploy", "connect", "info"],
+                              help="Deploy the local build to the instance, "
+                                   "connect to the instance's Cloud SQL, "
+                                   "or display instance config")
+        typical = parser.add_argument_group("typical arguments",
```
ehigham, line 299: Why are there so-called `typical` and `specific` arguments? I'm not sure this distinction is helpful. Either an option/switch is optional or it is not.

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+        """Use labels to identify the Cloud SQL name."""
+        instances = json.loads(shell(f"gcloud --project {self.project} --format=json "
+                                     "sql instances list "
+                                     f"--filter='labels.app_name=wfl AND labels.instance_id={self.instance_id}'",
+                                     quiet=True))
+        if len(instances) == 1:
+            return instances[0]["name"]
+        else:
+            error("No CloudSQL instance could be inferred")
+            info("    Couldn't find a single match for "
```
ehigham, line 78: just throw an exception with all these messages? 

pseudo-python
```python
if __name__ == "__main__":
    try:
        main()
        exit(0)
     except exception:
         print_message( exception ) 
         exit(1)
```

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+                                    f"--filter='name={self.cluster_name}'"))
+        if len(clusters) == 1:
+            return clusters[0]["zone"]
+        else:
+            error(f"No cluster zone could be found for cluster named {self.cluster_name}")
+            info("    Couldn't find a single match for "
+                 f"`--filter='name={self.cluster_name}'`")
+            return None if self.force else exit(1)
+
+    def validate(self):
```
ehigham, line 114: I don't understand why you've got a separate validation step.
Error when you encounter errors - don't save them for later and duplicate logic. 

In the olden days of mainframes and time sharing it was useful for your compiler to keep going and catch as many errors as possible. Nowadays, this doesn't actually help because later errors often depend on earlier ones. TLDR: don't accumulate errors.

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+        shell(f"kubectl config use-context {ctx}")
+        try:
+            namespaces = json.loads(shell("kubectl get namespace -o json", timeout=5))
+            return self.cluster_namespace in [item["metadata"]["name"] for item in namespaces["items"]]
+        except subprocess.TimeoutExpired:
+            error("Namespace query took too long--maybe you need to be on non-split VPN?")
+            return False
+
+
+class InfoCommand(Config):
```
ehigham, line 174: Inheritance is the source of all evil in OOP. If you read any modern best practices on it, they'll all say avoid it.

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
 
-        version = shell("cat version")
-        publish_docker_images(version=version)
+    def __init__(self, parsed_args: argparse.Namespace):
+        super().__init__(parsed_args)
+
+    def execute(self) -> int:
+        """Set up helm, publish docker, render CTMPL, helm deploy, migrate liquibase."""
+
+        if not input(f"Are you sure you want to deploy version {self.version}? [N/y]").lower().startswith("y"):
```
ehigham, line 220: Accessing properties of the base class like this will cause problems down the line if the base class changes.

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+
+class InfoCommand(Config):
+    """Class for info command that doesn't do anything special, just resolves/validates arguments."""
+
+    def __init__(self, parsed_args: argparse.Namespace):
+        super().__init__(parsed_args)
+
+    # This can't be static if we're to override it in subclasses
+    # noinspection PyMethodMayBeStatic
+    def execute(self) -> int:
```
ehigham, line 182: This is broken - you've had to work around your design already.

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
-    BLUE = 94
-    PURPLE = 95
-    WHITE = 97
-    BLACK = 30
-    DEFAULT = 39
+
+class Config:
+    """Base class for storing/resolving/validating command line input."""
+
+    def __init__(self, parsed_args: argparse.Namespace):
```
ehigham, line 30: No other `Config` seems to override this. This whole class can be replaced with a `Dictionary` returned by a `parse_input` function, whose keys were the field names of this class. Then:

1. You no longer need to override `__init__`
2. You no longer need inheritance - can use strategy pattern instead if this was Java, BUT
3. You no longer need the command classes because functions are objects in python

ops/cli.py
ehigham: For clarity, I'm taking about the `commands` mapping.

ops/cli.py
ehigham: As a final thought - this is a deployment script. Does it really need to be this engineered?

ops/cli.py
jack-r-warren: 1. It is private for better encapsulation (no need to pollute the namespace)
2. Editors can "fold" easier by this existing inside a class, making it quicker to get where you need in the file

ops/cli.py
jack-r-warren: `choices` doesn't operate on a mapping, you can give it a list of type T and a function from String to T. The closest I was able to get is this:
```
commands.add_argument("command",
                              type=lambda s: getattr(sys.modules[__name__], f"{s.title()}Command"),
                              choices=[DeployCommand, ConnectCommand, InfoCommand],
                              metavar="{deploy, connect, info}",
                              help="Deploy the local build to the instance, "
                                   "connect to the instance's Cloud SQL, "
                                   "or display instance config")
```
This does make the code you're referencing here cleaner: `config = args.command(args)`

ops/cli.py
ehigham: > It is private for better encapsulation (no need to pollute the namespace)
That's not what encapsulation is. Encapsulation is the union of data and operations so that one manipulates the essence of the data without exposing the data itself. There's no encapsulation here as none of `self`'s properties are used.
Why is polluting the global namespace here such an issue?

> Editors can "fold" easier
I'm using VSCode which handles functions just fine? If anything this funciton is one that shouldn't be folded - everything else is just boilerplate.

ops/cli.py
jack-r-warren: I think printing my own messages gives better output. Exceptions certainly still happen, if the shell commands actually run into errors those get passed up. In cases where the command worked and it is a human issue, I can at least say that.

ops/cli.py
ehigham: Alternatively, you could provid this function a `Dictionary<String, Callable>`, you could use that same dictionary to populate this parser as for dispatch. Then there's only one list to maintain.

ops/cli.py
jack-r-warren: Validating as a secondary step is a simple way to always validate even if you supply overrides for certain arguments. Making it one step wouldn't actually abstract away much code because all the same things still need to happen--whether it is "specific or calculate default, then validate" (as here) or "specific and validate or calculate default" (as you suggest) there isn't much of a change in the complexity. I think the order I chose gives a good demarcation for splitting it up.

The reason for having a `force` flag and allowing all these to be bypassed is so that the `connect` command can be used even if some parts of the wfl-instance aren't online or set up correctly. @rexwangcc said he didn't see a problem with running the same validation everywhere so long as it could be skipped if need be

ops/cli.py
ehigham: Or you can just put the error message into the exception and print that message?

ops/cli.py
jack-r-warren: We face the exact same issue if we pass the result of `parser.parse_args()` around like we currently do, if we so much as change an argument name we have to make changes everywhere. The benefit here is that every IDE and linter perfectly understands the usages here and will identify issues (or do the refactoring for you) whereas if we just pass an untyped object around we have to do the fix manually.

ops/cli.py
ehigham: The `force` flag is weird. You've added nulls when they're not necessary. You've got functions like `__find_cluster_name` and `__cluster_name_exists` where you're validating twice. It makes no sense.

ops/cli.py
jack-r-warren: If Python was properly typed I'd need this, but since it isn't who cares. I swapped out the lint ignore with `@staticmethod` since that's more clear, it doesn't cause any issues because Python is none the wiser

ops/cli.py
ehigham: "The IDE helps me" is not a good excuse

ehigham: Just because the IDE helps you deosn't mean you should.
Since this is a weakly typed language you are right - objects are pretty much dictionaries. I'm trying to help you in terms of software engineering principles, however.

ops/cli.py
ehigham: > since it isn't who cares

Why have you made this change then??

ops/cli.py
jack-r-warren: Fair enough. Abstracted out the mapping to the CLI class.

ops/cli.py
jack-r-warren: It is a discussion of
```
error("my message")
info("my new line")
exit(1)
```
versus
```
raise RuntimeError("my message\nmy new line")
```

I don't think that using exceptions for control flow grants us much benefit in terms of keeping the code clear or easy to maintain, especially when we consider cases where we don't actually want to exit all the way in the case of `--force`. If we had less console output and were comfortable just letting the raw CallProcessExceptions from shell commands rise up to the user, then we can absolutely get rid of all the error text and do that. I firmly believe that helpful error text will save us time and that we shouldn't make code worse in the pursuit of making it only marginally shorter.

ops/cli.py
jack-r-warren: The former function is used as a default while the latter is meant to check supplied values. I agree that there's a common case where we'll validate twice--we generate the default and then pass it through the validator. Alternatively we can validate only passed inputs, and that's a totally fair--I mentioned above that taking that route wouldn't actually reduce much code since we still need both pieces of functionality.

ops/cli.py
jack-r-warren: I see very very little benefit to using a dictionary here. If we want to go that route, that's okay, but we need to be aware that in doing so we are making the script less checked.

My entire thinking about deployment scripts is this: when it is 2am and one of us is awake because our prod went down, a fully-fledged script with helpful error text, argument validation, and sane but overridable defaults is what I want to be working with. Engineering it in this way means that our tooling can help check our work so we can catch mistakes earlier and fix them quicker. Production isn't going to break because of this script, but this script is what will help fix it, and I think it is worth it to put the time in to prepare this script for that possibility.

ops/cli.py
```diff
@@ -1,351 +1,343 @@
...
+            success("Exiting due to '--dry-run'")
+            return 0
+        else:
+            return config.execute()
+
 
 if __name__ == "__main__":
-    c = CLI()
-    exit(c())
+    exit(CLI().execute())
```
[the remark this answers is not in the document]
jack-r-warren: (Resolving because I discuss this below--TL;DR: dictionaries are less checked and classes help our tooling help us)

ops/cli.py
jack-r-warren: Re folding I mean that editors can fold classes down, just another layer of organization that some folks still might not use.
Polluting the namespace might not be a big deal for us, I dunno. I see it as a code quality thing but maybe that's just my out-of-touch academia talking.

ops/cli.py
jack-r-warren: Fair enough. Updated this to make environment and instance required. Can't do a whole lot about the way that argument groups work: every option that isn't in a group is an "optional argument" and comes first. Renamed this "required arguments" and the specific one "miscellaneous arguments" so they still fall in order of how most folks would execute the command

ops/cli.py
jack-r-warren: Resolved in new version, no methods exist in the file.

ops/cli.py
jack-r-warren: Resolved in new version, no argument groups exist in the file.

ops/cli.py
jack-r-warren: Resolved in the new version, eliminated `--force` and most error messages to instead use exceptions to stop program flow.

ops/cli.py
jack-r-warren: Solved in the new version. Individual functions handle inference and validation so that validation only occurs on manually customized arguments. Any errors cause immediate exits.

ops/cli.py
jack-r-warren: Resolved in new version. No methods exist in the file and `typing` has been important to type the core components (config and command_mapping).

ops/cli.py
jack-r-warren: Resolved in new version. No methods exist in the file, no inheritance exists in the file, commands are not classes. "Commands" are now done via a higher-order function such that individual steps can be very easily commented out.

ops/cli.py
jack-r-warren: Resolved in new version, no methods exist in the file.

ops/cli.py
jack-r-warren: Resolved in new version. No inheritance exists in the file and the `typing` library has been imported to more strongly type the mapping from command names to composite functions.

ops/cli.py
```diff
@@ -1,351 +1,309 @@
...
-    info("=> Setting up Helm charts")
-    ADD = " ".join(["helm repo add gotc-charts",
-                    "https://broadinstitute.github.io/gotc-helm-repo/"])
-    shell(ADD)
+@dataclass
+class WflInstanceConfig:
+    """Data class representing the 'state' of this script's configuration as it interacts with infrastructure."""
+    command: str
+    environment: str
+    instance_id: str
```
jack-r-warren, line 34: Without a default these become required by `@dataclass` for construction (but the CLI parsing also enforces that these are required anyway)

ops/cli.py
```diff
@@ -1,351 +1,309 @@
...
+    shell(f"kubectl config use-context {ctx}")
+    try:
+        namespaces = json.loads(shell("kubectl get namespace -o json", timeout=5))
+        if config.cluster_namespace not in [item["metadata"]["name"] for item in namespaces["items"]]:
+            info(f"    Available namespaces in {config.cluster_name}:", plain=True)
+            info(shell("kubectl get namespace", quiet=True), plain=True)
+            raise RuntimeError(f"Namespace {config.cluster_namespace} not found in {config.cluster_name}")
+        else:
+            shell(f"kubectl config set-context --current --namespace={config.cluster_namespace}")
+            success("Kubernetes configured")
```
jack-r-warren, line 127: This used to be a validation step, but no longer. Configuring kubectl now uses a reincarnation of some of the same logic to just print very helpful error text if the namespace doesn't exist.

ops/cli.py
```diff
@@ -1,351 +1,309 @@
...
+
+
+def render_values_file(config: WflInstanceConfig) -> None:
+    """Render the values file and store pertinent info from it in the config."""
+    info("=>  Rendering Helm values file")
+    deploy = os.path.join("derived", "helm", "deploy")
+    if not os.path.exists(deploy):
+        os.makedirs(deploy)
+    values = "wfl-values.yaml"
+    config.rendered_values_file = os.path.join(deploy, values)
```
jack-r-warren, line 148: We store the path in the config so that there's no longer a tight coupling between this function and the helm deploy (this means that the deploy --dry-run can actually make the values file)

ops/cli.py
```diff
@@ -1,351 +1,309 @@
...
+    ctmpl = f"derived/2p/{config.environment}/helm/{values}.ctmpl"
+    shutil.copy(ctmpl, deploy)
+    render_ctmpl(ctmpl_file=f"{config.rendered_values_file}.ctmpl", WFL_VERSION=config.version)
+    with open(config.rendered_values_file) as values_file:
+        helm_values = yaml.safe_load(values_file)
+        env = helm_values["api"]["env"]
+        config.db_username = env["WFL_POSTGRES_USERNAME"] or env["ZERO_POSTGRES_USERNAME"]
+        config.db_password = env["WFL_POSTGRES_PASSWORD"] or env["ZERO_POSTGRES_PASSWORD"]
```
jack-r-warren, line 156: Add a preference for WFL_POSTGRESS_... fields that don't currently exist so that getting rid of ZERO at some point in the future becomes a more tractable task.

ops/cli.py
```diff
@@ -1,351 +1,309 @@
...
+        render_values_file,
+        exit_if_dry_run,
+        prompt_deploy_version,
+        configure_cloud_sql_proxy,
+        publish_docker_images,
+        helm_deploy_wfl,
+        run_liquibase_migration,
+        stop_cloud_sql_proxy,
+        print_deployment_success
+    ]
```
jack-r-warren, line 266: @ehigham and @rexwangcc both mentioned they like to comment out parts of this script in order to make it do whatever specific thing they want. I think this part will work well for that: there's obviously some interdependencies but other than that things can be removed or reordered with impunity. Also becomes very easy to make new commands in the same vein.

ops/cli.py
```diff
@@ -1,351 +1,309 @@
...
+                        help="specify the GCP project name instead of 'broad-{ENV}'")
+    parser.add_argument("--cloud-sql-name",
+                        help="specify the Cloud SQL name instead of inferring from labels")
+    parser.add_argument("--cluster-zone",
+                        help="specify the zone for the GKE cluster instead of inferring from CLUSTER_NAME")
+    parser.add_argument("--cluster-name",
+                        help="specify the GKE cluster name instead of inferring from labels")
+    parser.add_argument("--cluster-namespace",
+                        help="specify the K8s namespace instead of '{INSTANCE}-wfl'")
+    return WflInstanceConfig(**vars(parser.parse_args()))
```
jack-r-warren, line 294: Match argparse keys to `@dataclass` constructor keys (which are the same as the WflInstanceConfig fields)

ops/cli.py
```diff
@@ -1,351 +1,310 @@
...
+from typing import Dict, Callable, List
+
 import yaml
 
+from render_ctmpl import render_ctmpl
+from util.misc import info, success, error, warn, shell
+
 # In case this will need to run on Windows systems
 if sys.platform.lower() == "win32":
     os.system("color")
```
jack-r-warren, line 26: I think we should maybe remove this because anyone running WFL stuff on Windows will need to be using some sort of Linux shell because of `make`. I use Windows for all of my work outside of Broad and I can't imagine someone would be using a terminal that wouldn't support Unix colors (all modern things do).

ops/cli.py
jack-r-warren: Also, I specifically configure the Cloud SQL proxy before any of the tasks with remote side effects because it is the most likely to fail. Now if you get a port error you can just clear the port and deploy again without needing to comment anything out.

ops/cli.py
jack-r-warren: (Outdated because I updated this bit of code to actually work but conceptually it is the same)

ops/cli.py
rexwangcc: I like this change!

ops/cli.py
```diff
@@ -1,351 +1,319 @@
...
+
+def main() -> int:
+    """Call `cli` and run each function in the command mapping in order."""
+    config = cli()
+    try:
+        for func in command_mapping[config.command]:
+            func(config)
         return 0
+    except Exception as err:
+        error(f"{err}")
```
ehigham, line 307: You don't need the formatted string here

ops/cli.py
```diff
@@ -1,351 +1,319 @@
...
+    parser.add_argument("--cluster-name",
+                        help="specify the GKE cluster name instead of inferring from labels")
+    parser.add_argument("--cluster-namespace",
+                        help="specify the K8s namespace instead of '{INSTANCE}-wfl'")
+    return WflInstanceConfig(**vars(parser.parse_args()))
+
+
+def main() -> int:
+    """Call `cli` and run each function in the command mapping in order."""
+    config = cli()
```
ehigham, line 301: you can probably move this into the `try` block

ops/cli.py
```diff
@@ -1,351 +1,319 @@
...
+def validate_cloud_sql_name(config: WflInstanceConfig) -> None:
+    """Validate that the stored Cloud SQL name exists."""
+    if config.cloud_sql_name:
+        info("=>  Validating Cloud SQL name")
+        instances = json.loads(shell(f"gcloud --project {config.project} --format=json "
+                                     "sql instances list"))
+        if config.cloud_sql_name not in [i["name"] for i in instances]:
+            info(f"    Available Cloud SQL instances in {config.project}:", plain=True)
+            info(shell(f"gcloud --project {config.project} sql instances list", quiet=True), plain=True)
+            raise RuntimeError(f"Cloud SQL instance {config.cloud_sql_name} not found")
```
ehigham, line 59: Just thinking what the user will see there (and in other places where yu've `info`ed before `raise`)- will we see diagnostics before the error message? That might be a bit confusing if so.
Maybe bundle the diagnostics with the error message?

ops/cli.py
jack-r-warren: Well we need to convert it to a string somehow to avoid linting issues. Changed it to `str(err)`

ops/cli.py
jack-r-warren: Having it outside of the try block will make the stacktrace get printed if there's an issue parsing arguments into the config. Since the config's constructor is generated by `@dataclass`, the extra info is helpful in diagnosing what actual argument didn't get matched right (can't just look at the source code).

ops/cli.py
jack-r-warren: This is an issue latent in using exceptions for control flow that I pointed out earlier--that doing the messages myself would provide better output. I'd like to avoid printing `gcloud`'s entire output in red text to stderr, but it is either that or print this first if we choose to use exceptions.

jack-r-warren:
> I think this is great simplification. Thank you.
> There are a couple of minor things/things I've commented along with this
> general comment.
> 
> Your idea of a processing sequence is quite fun! It's reminiscent of Haskell's
> `sequence` function :). There is a bit of a problem design wise in that every
> single function here has to know about a `WflInstanceConfig`. The TD;LR (detail
> below) is to give procedures as much information as they need to and no more.
> The naming is also quite hard to get right I think - these aren't really
> functions but really a processing sequence.
> 
> As such, I think it would be more obvious to just implement these as normal
> functions and pass the arguments from the `config` that each function needs.
> Running the bits that we need will be just as easy i think.
> 
> [TL;DR]
> While this isn't a big deal in python, it is a big deal for languages that
> support separate compilation - every time I want to change my data structure,
> I have to recompile everything. This isn't particularly elegant and in
> languages like C++, compilation can take an increasingly large amount of time
> and resources. Of course there is a trade off here with respect the length of
> the formal parameter lists. While compilation isn't an issue here, I think this
> approach to designing and writing procedures helps create simple and testable
> code.

I'll take functions knowing more than they strictly need in favor of loose coupling, especially in a deploy script.

ops/cli.py
ehigham: Then use an exception type that has some "details" field?

jack-r-warren: Going to back to printing it myself.

PR #96, opened 2020-08-11
[GH-883] Update test sample upload paths

samanehsan:
### Purpose
Update the helper script to manually test the cloud fn (follow-up to https://github.com/broadinstitute/wfl/pull/95)

### Changes
Add "chipname" to the upload path to match PTC behavior. 

### Review Instructions

- No instructions.


samanehsan:
> FWIW our bucket doesn't have paths like this, but that's probably an inconsistency with the bucket rather than the other way around. https://console.cloud.google.com/storage/browser/dev-aou-arrays-input;tab=objects?forceOnBucketsSortingFiltering=false&project=broad-gotc-dev-storage&prefix=

Yup those objects were created by this script!

PR #105, opened 2020-08-19
[GH-819] First draft of baits and targets editing

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-819

### Changes

- There was already a `baits-and-targets` map. Now, you can override any of its keys with
```
-B:bait_set_name=my-new-value
```
- The documentation for these options is also automatically generated from the defaults map itself, so we don't have to maintain it:
```
...
Override the baits and targets configuration by specifying arguments prefixed with -B
Example: -B:bait_set_name=my-new-value
Configuration defaults:
:bait_set_name=whole_exome_illumina_coding_v1
:bait_interval_list=gs://broad-references-private/HybSelOligos/whole_exome_illumina_coding_v1/whole_exome_illumina_coding_v1.Homo_sapiens_assembly38.baits.interval_list
:target_interval_list=gs://broad-references-private/HybSelOligos/whole_exome_illumina_coding_v1/whole_exome_illumina_coding_v1.Homo_sapiens_assembly38.targets.interval_list
:cram_ref_fasta=gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta
:cram_ref_fasta_index=gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta.fai
```
- There's nothing hardcoded about this, adding something new to the defaults map makes it editable and appear in the documentation. Parsing happens by literally creating a map from any passed arguments with the `-B` prefix and merging it with the defaults.
- Added a `baits-and-targets` input to several existing functions to let the overridden config makes its way down from the CLI entrypoint
- Improved help text printing (print help text without wrapping it in `log/error`)

### Review Instructions

- Nothing yet, this is just a draft to see if I'm heading in the right direction.

### DIFF
--- a/api/src/zero/module/xx.clj
+++ b/api/src/zero/module/xx.clj
@@ -4,7 +4,6 @@
             [clojure.java.io :as io]
             [clojure.set :as set]
             [clojure.string :as str]
-            [clojure.tools.logging :as log]
             [zero.module.all :as all]
             [zero.references :as references]
             [zero.service.cromwell :as cromwell]
@@ -47,6 +46,19 @@
         (->> (str/join \newline))
         (format zero/the-name title))))
 
+(defn describe-overrides
+  "Describe how to override OPTIONS that configure NAME via FLAG-PREFIX"
+  [name flag-prefix options]
+  (when (seq options)
+    (-> [""
+         (format "Override the %s configuration by specifying arguments prefixed with %s"
+                 name flag-prefix)
+         (format "Example: %s%s=my-new-value"
+                 flag-prefix (first (keys options)))
+         "Configuration defaults:"]
+        (concat (map (fn [[k v]] (str k "=" v)) options))
+        (->> (str/join \newline)))))
+
 (def workflow-wdl
   "The top-level WDL file and its version."
   {:release "ExternalExomeReprocessing_v1.1"
@@ -74,8 +86,8 @@
             :contamination_sites_mu  (str subset ".mu")
             :contamination_sites_ud  (str subset ".UD")})))
 
-(def baits-and-targets
-  "Baits and targets inputs."
+(def default-baits-and-targets
+  "Default baits and targets inputs."
   (let [private              "gs://broad-references-private/"
         bait_set_name        "whole_exome_illumina_coding_v1"
         interval_list        (str private "HybSelOligos/" bait_set_name "/" bait_set_name ".Homo_sapiens_assembly38")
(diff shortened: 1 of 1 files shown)


api/src/zero/module/xx.clj
```diff
@@ -88,9 +100,18 @@
...
      :cram_ref_fasta_index (str fasta ".fai")}))
 
+(defn apply-overrides
+  "Return an map like DEFAULTS but overridden with any RAW-ARGS that have FLAG-PREFIX."
+  [defaults flag-prefix raw-args]
+  (merge defaults (->> raw-args
+                       (filter #(str/starts-with? % (str flag-prefix ":")))
+                       (map #(str/split (subs % 3) #"="))
+                       (map (fn [[k v]] [(keyword k) v]))
+                       (into {}))))
```
jack-r-warren, line 110: I can't really tell unreadable clojure from readable clojure, happy to do this some other way if this is confusing for folks

api/src/zero/module/xx.clj
```diff
@@ -88,9 +100,18 @@
...
+
 (defn make-inputs
   "Return inputs for reprocessing IN-GS into OUT-GS in ENVIRONMENT."
-  [environment out-gs in-gs]
+  [baits-and-targets environment out-gs in-gs]
```
jack-r-warren, line 114: Added this as the first argument whenever a function needs to accept it because there are two cases where we use `partial` to slot this in (in other words, it being first is consistent and makes programmatic sense in some of the cases). Happy to put it elsewhere if we think that makes more sense. See below for further comment.

api/src/zero/module/xx.clj
```diff
@@ -180,18 +201,26 @@
...
+          env               (zero/throw-or-environment-keyword! (first args-no-overrides))]
+      (apply (case (count args-no-overrides)
+               4 (partial on-hold-some-workflows baits-and-targets)
                3 (partial all/report-status cromwell-label)
                2 release-some-on-hold-workflows
+               1 print-help
                (throw (IllegalArgumentException.
                        "Must specify 2 to 4 arguments.")))
-             env (rest args)))
+             env (rest args-no-overrides)))
```
jack-r-warren, line 223: This is why I think having `baits-and-targets` generally be the first argument makes sense. We already do the same thing for `all/report-status` and this lets us very easily `partial` it into the `on-hold-some-workflows` function. This the only place where it really needs to be first, but I went with then being consistent as a default.

api/src/zero/module/xx.clj
```diff
@@ -180,18 +201,26 @@
            (take max)
            (run! release!)))))
 
+(defn print-help
+  [& _]
+  (println description)
+  (println (describe-overrides "baits and targets" "-B" default-baits-and-targets)))
```
jack-r-warren, line 207: - This being a function means that `describe-overrides` can help automate documenting this for users (I think)
- This accepting any arguments means we can slot it into the `(apply (case ...` below to simply print help text if the user supplies only a single argument (I think that's useful since I think this already needs to be a function)

jack-r-warren:
Talked with Meaghan and Rex: since users can provide any bucket path they want, it makes sense to put a note in the help text reminding them that Cromwell still needs to access the buckets (via the picard-prod SA).

Rex agreed with me that hardcoding this in the help text is better than trying to use Vault and environments.clj to try to interpolate the true SA. Doing so would be error prone and we rarely read from Vault in WFL so any errors are loud and obnoxious. It is very unlikely that the SA will change suddenly during the lifetime of this CLI.

For context, I actually wrote code to interpolate the SA and it wasn't exactly concise or all that robust. I took it out in favor of hardcoding because I honestly think that'll be the lowest-maintenance-cost option.

PR #108, opened 2020-08-20
[GH-901] More deployment commands

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-901

### Changes

- Make the script a bit more eager to exit (I thought I had done this the first time through but maybe it got lost in me reverting back how the script exited overall)
- When deploying, print later so that more info is printed
- New `ops-deploy` checks that the tag is correct and doesn't push to docker
- New `hornet-pre-deploy` makes the tag and pushes to docker and does nothing else

### Review Instructions

- Looking for code review at this point. Discussion in Slack was specific in recommending against testing this without group consensus since some folks might want to talk about the tagging process.


ops/cli.py
```diff
@@ -275,6 +295,32 @@ def print_deployment_success(config: WflInstanceConfig) -> None:
         run_liquibase_migration,
         stop_cloud_sql_proxy,
         print_deployment_success
+    ],
+    "ops-deploy": [
```
samanehsan, line 299: How about changing this to something like "deploy-tag" or "deploy-from-tag"? It's not immediately clear what "ops-deploy" means

ops/cli.py
```diff
@@ -275,6 +295,32 @@ def print_deployment_success(config: WflInstanceConfig) -> None:
...
+        render_values_file,
+        print_config,
+        exit_if_dry_run,
+        configure_cloud_sql_proxy,
+        helm_deploy_wfl,
+        run_liquibase_migration,
+        stop_cloud_sql_proxy,
+        print_deployment_success
+    ],
+    "hornet-pre-deploy": [
```
samanehsan, line 316: Similar to above, how about naming this to something more descriptive like "tag-and-push-images"?

ops/render_ctmpl.py
```diff
@@ -5,13 +5,13 @@
...
 
-def render_ctmpl(ctmpl_file: str, **kwargs) -> int:
+def render_ctmpl(ctmpl_file: str, vault_token_path: str, **kwargs) -> int:
     """Render a ctmpl file."""
     info(f"=>  Rendering ctmpl file {ctmpl_file}")
     envs = " ".join([f"-e {k}={v}" for k, v in kwargs.items()]) if kwargs else ""
     info(f"=>  Feeding variables: {envs}")
     shell_unchecked(" ".join(['docker run -i --rm -v "$(pwd)":/working',
-                              '-v "$HOME"/.vault-token:/root/.vault-token',
+                              f'-v ${vault_token_path}:/root/.vault-token',
```
samanehsan, line 14: I don't think you need the `$` here

jack-r-warren: Thanks, good catch!

PR #116, opened 2020-08-24
[GH-841] Fix some endpoint specs and add Swagger OAuth2 button

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-841

### Changes

- Fix a mis-configured spec for `/workload`.
- Add partial support for Swagger OAuth2. 

### Review Instructions

**Note:**
The OAuth support for Swagger is not fully-functional. I haven't found a decent way to make it work. We might want to keep an eye on how the `reitit` folks respond to https://github.com/metosin/reitit/issues/430

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -372,11 +372,13 @@ dmypy.json
 
 ### VisualStudioCode ###
 .vscode/*
-!.vscode/settings.json
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
 *.code-workspace
+
+### Calva ###
+.calva/
 
 ### VisualStudioCode Patch ###
 # Ignore all local history of files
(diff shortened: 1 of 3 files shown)


rexwangcc:
> LGTM though I'm not totally sure how to test this (I don't know what the "partially working" means in context). Since the ui isn't really part of what's needing to work (or even accessible) in prod anyway I have a lower bar.

Good point. "Partially works" means that if you go to https://dev-wfl.gotc-dev.broadinstitute.org/swagger/index.html right now, there's no way you could actually use the swagger since there's no "Authorize" button. This PR adds that button as well as the scopes for you to authorize with Google. However, due to the issue I mentioned in the description, the client-id is not pass-able to the Google, so we have to manually grab and pass the client-id in the dialog. I think we can leave it as it is since the original ticket for this PR was just fixing the spec.

PR #117, opened 2020-08-24
[GH-905] [GH-908] Makefile fixes

ehigham:
### Purpose
@jack-r-warren noted that using `.ts` for timestamps clobbers tools that register the extension `.ts` with TypeScript and suggested we use `.timestamp` instead.
This PR addresses the following issues with our `make` infrastructure:

[GH-905](https://broadinstitute.atlassian.net/browse/GH-905): use `.timestamp` instead of `.ts`
[GH-908](https://broadinstitute.atlassian.net/browse/GH-908): update the UI semantic version


ui/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ui",
-  "version": "0.1.0",
+  "version": "%VERSION%",
```
rexwangcc, line 3: It seems we should avoid manually change [`package-lock.json` file](https://docs.npmjs.com/configuring-npm/package-lock-json.html) for it's generated by `npm install` from `package.json` automatically, changes `such as `"%VERSION%"` might be overwritten. But since the edit of `package.json` happens way later than `npm install`, I cannot come up with a better solution 🤔

ehigham: Yeah I was wondering about this. Do we need to commit `package-lock.json`?

jack-r-warren: Most integrations of `npm` into a multi-language build tool that I've seen end up generating the `package.json` itself to get around this and other issues. Kotlin uses Gradle's dependency graph to create a package.json programmatically just before it is needed

ehigham: We probably want to regenerate this every time we bump the version right?

rexwangcc: > We probably want to regenerate this every time we bump the version right?
I think so.

>Do we need to commit package-lock.json?

https://stackoverflow.com/a/44210813/7247312 they say so.

ehigham: ugh why do they make you commit derived files :'(

ehigham: OK so my solution is to leave these files untouched. I'm gonna change the version strings in the copies that get put into `derived`. This way, `npm install` in the `ui` directory still works and you don't need to use `make`.

PR #124, opened 2020-08-27
[GH-911] configure pytest environment with `pytest-env` plugin

ehigham:
### Purpose
`make cloud_function TARGET=check` currently fails as the environment is misconfigured.
This is a symptom of a larger problem in that to run the tests, we have to manually set the environment.

[pytest-env](https://github.com/MobileDynasty/pytest-env) is a nice little plugin that reads the environment variables out of `pytest.ini`. I propose we use this to manage and configure the environment.

This PR fixes GH-911.

related SO article:
https://stackoverflow.com/questions/36141024/how-to-pass-environment-variables-to-pytest

### DIFF
--- a/cloud_function/pytest.ini
+++ b/cloud_function/pytest.ini
@@ -1,6 +1,6 @@
 [pytest]
 env =
-    WFL_URL="https://workflow-launcher.gotc-dev.broadinstitute.org"
+    WFL_URL="https://dev-wfl.gotc-dev.broadinstitute.org"
     CROMWELL_URL="https://cromwell-gotc-auth.gotc-dev.broadinstitute.org/"
     WFL_ENVIRONMENT="aou-dev"
     OUTPUT_BUCKET="gs://fake-bucket"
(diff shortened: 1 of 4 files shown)


cloud_function/pytest.ini
```diff
@@ -0,0 +1,6 @@
+[pytest]
+env =
+    WFL_URL="https://workflow-launcher.gotc-dev.broadinstitute.org"
```
jack-r-warren, line 3:
```suggestion
    WFL_URL="https://dev-wfl.gotc-dev.broadinstitute.org"
```

jack-r-warren: workflow-launcher, if it still exists, is using orphaned resources. `dev-wfl` is what is currently managed in gotc-dev.

ehigham: @samanehsan - this is lifted from your documentation. Can you confirm?

samanehsan: My mistake! The correct url is `WFL_URL="https://dev-wfl.gotc-dev.broadinstitute.org"`. The tests currently mock everything that use these values, so that's why they didn't fail

PR #125, opened 2020-08-27
[GH-914] Release on version bump

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-914

### Changes

- Make cli.py work better if you call it from a weird spot (ended up not needing it but you can now call it from anywhere and it'll still find the `derived` folder)
- Make cli.py not require ENV and INSTANCE if you're just using `tag-and-push-images`
- Rename and rearrange the docker release action to instead use `cli.py`

### Review Instructions

- Test cli.py relative paths by doing something like `./cli.py deploy gotc-dev dev -d` (so you'd be in the ops folder and it would still figure out the paths)
- Test cli.py arguments by running `./ops/cli.py info` (see the error message) and `./ops/cli.py tag-and-push-images -d` (no error message, just exits because of `-d`)
- Action can't be tested really until we merge it :/ it is using stuff @ehigham wrote though, I didn't actually add much to it


.github/workflows/release_latest_version.yml
```diff
@@ -0,0 +1,68 @@
...
+        run: |
+          docker login \
+            --username "${{ secrets.dockerhub_username }}" \
+            --password "${{ secrets.dockerhub_password }}"
+
+      - name: Pre-Build
+        run: make MODULES="${MODULES}" TARGET=prebuild
+
+      - name: Build
+        run: make MODULES="${MODULES}" TARGET=build
```
ehigham, line 62: You probably want to run check as well:

```suggestion
        run: make MODULES="${MODULES}" TARGET="build check"
```

ehigham (commented):
Maybe you could tag wfl in this workflow too (similar to PTC's release workflow)?

ops/cli.py
```diff
@@ -248,30 +261,35 @@ def print_deployment_success(config: WflInstanceConfig) -> None:
 
 def make_git_tag(config: WflInstanceConfig) -> None:
     info("=>  Tagging current commit with version")
-    shell(f"git tag -a v{config.version} -m 'Created by cli.py {config.command}'")
-    shell(f"git push origin v{config.version}")
+    shell(f"git tag -a v{config.version} -m 'Created by cli.py {config.command}'", cwd=config.wfl_root_folder)
+    shell(f"git push origin v{config.version}", cwd=config.wfl_root_folder)
     success(f"Tag 'v{config.version}' created and pushed")
```
ehigham, line 266: I really don't think this script should be doing this.
It should be using a release, not creating one.

jack-r-warren:
> Maybe you could tag wfl in this workflow too (similar to PTC's release workflow)?

WFL gets tagged by the `tag-and-push-images` part of cli.py. The reason I'm having cli.py do it is because cli.py also has to read the tags, so this way it is self-consistent and is the source of truth.

ops/cli.py
jack-r-warren: The part that uses a release is over in the Jenkins job that actually deploys WFL. Just like you mentioned in your comment, we should tag WFL in this job and this is the part that does that

.github/workflows/release_latest_version.yml
jack-r-warren: Good idea, added (just added it as its own step)

PR #131, opened 2020-09-03
[GH-939] Fix issues with Cloud Function

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/jira/software/projects/GH/boards/530?selectedIssue=GH-939

#### Cloud Function
Looking at the Google Cloud Bucket implementation, which only implements `__repr__` but `__str__`

```python
def __repr__(self):
  return "<Bucket: %s>" % (self.name,)
```
we cannot simply use `str(Bucket())` to stringify the bucket name, which is why the test is failing. 

#### Tests
We should run Cloud Function Unit Tests per PR as well, which should have caught this bug earlier.

### Review Instructions

- Run `pytest tests/` again.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -112,6 +112,7 @@ jobs:
           make api TARGET=check
 
       - name: Cloud Function Unit Test
+        env:
+          GOOGLE_APPLICATION_CREDENTIALS: sa.json
         run: |
-          gcloud auth application-default login
           make cloud_function TARGET=check
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -107,5 +107,8 @@ jobs:
           make cloud_function TARGET=check
 
       - name: Cloud Function Unit Test
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8' # Version range or exact version of a Python version to use, using SemVer's version range syntax
         run: |
           make api TARGET=check
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -106,9 +106,11 @@ jobs:
         run: |
           make cloud_function TARGET=check
 
-      - name: Cloud Function Unit Test
+      - name: Setup Python3
         uses: actions/setup-python@v2
         with:
           python-version: '3.8' # Version range or exact version of a Python version to use, using SemVer's version range syntax
+
+      - name: Cloud Function Unit Test
         run: |
           make api TARGET=check
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -113,5 +113,5 @@ jobs:
 
       - name: Cloud Function Unit Test
         run: |
-          gcloud auth activate-service-account --key-file=sa.json
+          gcloud auth application-default login
           make cloud_function TARGET=check
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -24,6 +24,11 @@ jobs:
           tools-deps: '1.10.1.469'
           boot: 2.8.3
 
+      - name: Setup Python3
+        uses: actions/setup-python@v2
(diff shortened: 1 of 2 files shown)


jack-r-warren (approved):
Tests passed for me on this branch. LGTM!

jack-r-warren (changes requested):
Okay on checking the failed PR test, the job doesn't seem to set up `gcloud` so it'll fail in PR. We may need to add code in to authorize `gcloud` as a service account.

samanehsan:
I believe when the tests call `client = mock.create_autospec(storage.Client())` to create a mock client object, `storage.Client()` tries to access the default google account credentials. We don't even need real credentials for the test, so I wonder if there's a different way to mock this functionality 🤔

rexwangcc:
> I believe when the tests call `client = mock.create_autospec(storage.Client())` to create a mock client object, `storage.Client()` tries to access the default google account credentials. We don't even need real credentials for the test, so I wonder if there's a different way to mock this functionality 🤔

I did some googling and it turns out there are no google cloud api client emulator at the moment... Mocking like this does not prevent `Client()` to access the credentials, and it could have been easier if we apply dependency injection on `submit_aou_workload`. I will try to make the credentials access work for now since we already have that setup in the actions.

rexwangcc:
@samanehsan @jack-r-warren Could you take another look? I managed to fix the GCloud Client Auth mock issue.

cloud_function/main.py
```diff
@@ -100,7 +100,7 @@ def submit_aou_workload(event, context):
     input_data = json.loads(manifest_file_content)
     notification = input_data['notifications'][0]
     input_values = notification.values()
-    input_files = set([f for f in input_values if str(f).startswith(f"gs://{bucket}")])
+    input_files = set([f for f in input_values if str(f).startswith(f"gs://{bucket.name}")])
```
rexwangcc, line 103: Based on my local testing, even this is not enough, see:
```python
>>> b = client.bucket("gs://broad-exomes-dev1")
>>> str(b)
'<Bucket: gs://broad-exomes-dev1>'
>>> b.name
'gs://broad-exomes-dev1'
>>>
```

We probably need something like:
```suggestion
    input_files = set([f for f in input_values if str(f).startswith(bucket.name)])
```

.github/workflows/pr-test.yml
```diff
@@ -101,7 +106,13 @@ jobs:
...
-
-      - name: Unit Test
+          echo ::set-env name=GOOGLE_APPLICATION_CREDENTIALS::$(find `pwd` -maxdepth 1 -name sa.json)
+      - name: WFL Unit Test
         run: |
           make api TARGET=check
+
+      - name: Cloud Function Unit Test
+        # env:
+        #   GOOGLE_APPLICATION_CREDENTIALS: sa.json
```
rexwangcc, line 116:
```suggestion
```

cloud_function/main.py
rexwangcc: Actually never mind, the `name` indeed depends on the `event['bucket']` within this context and it's the bucket name without `gs://`. @samanehsan correct me if I'm wrong here 😄

samanehsan: Yes, that's right!

.github/workflows/pr-test.yml
```diff
@@ -101,7 +106,11 @@ jobs:
           fi
           echo ::add-mask::${GOOGLE_OAUTH_ACCESS_TOKEN}
           echo ::set-env name=GOOGLE_OAUTH_ACCESS_TOKEN::${GOOGLE_OAUTH_ACCESS_TOKEN}
-
-      - name: Unit Test
+          echo ::set-env name=GOOGLE_APPLICATION_CREDENTIALS::$(find `pwd` -maxdepth 1 -name sa.json)
```
jack-r-warren, line 109: Excellent fix

jack-r-warren (approved):
LGTM

PR #132, opened 2020-09-03
[GH-818] Make WGS db changes CASCADE

samanehsan:
### Purpose
Apply the db changes to existing tables as well, otherwise liquibase will throw an error when trying to apply this changeset.

### Changes

- No changes.

### Review Instructions

- No instructions.


rexwangcc (approved):
Good catch!

PR #133, opened 2020-09-03
Release 0.1.9 before I deprecate the WGS CLI.

rexwangcc:
### Purpose

- Release 0.1.9 before I deprecate the WGS CLI in another PR.

### Changes

- No changes.

### Review Instructions

- No instructions.


jack-r-warren (approved):
Full support because it'll QA the job that should build WFL docker images upon this being merged

rexwangcc:
>Full support because it'll QA the job that should build WFL docker images upon this being merged

That's a good point! So I guess we don't have to run `cli.py` to build and push docker images from local envs anymore?

PR #137, opened 2020-09-09
[GH-914] Add Python and Git configurations

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-914

### Changes

- Locks Python version to what works for me locally. This will solve the issue of the default Python being too old to use `dataclass`
- Use pip to install pyyaml, a dependency I haven't thought about in a while but that I suppose we'll need here
- Competing information online about if we need to do configuration to Git, but I'd rather just play it safe. Git user and email inferred from the commit.

### Review Instructions

- Wait for a new release, I suppose.
- I wish I had set up better tooling to test this, I'm sorry that getting this working has been so rocky. Since our releases and tags actually have meaning even if I were to do this locally I would still have to wait for real releases.

### DIFF
--- a/.github/workflows/release_latest_version.yml
+++ b/.github/workflows/release_latest_version.yml
@@ -15,11 +15,24 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      
+      - name: Configure Git
+        run: |
+          git config --local user.email "$(git log --format='%ae' HEAD^!)"
+          git config --local user.name "$(git log --format='%an' HEAD^!)"
 
       - name: Setup Java
         uses: actions/setup-java@v1
         with:
           java-version: '11'
+          
+      - name: Setup Python
+        uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


ehigham (commented):
Will the list of python dependencies ever change? Might it be better to encode this information into a `requirements.txt` under the `ops/` directory then `pip install -r requirements.txt`?

jack-r-warren:
> Will the list of python dependencies ever change? Might it be better to encode this information into a `requirements.txt` under the `ops/` directory then `pip install -r requirements.txt`?

I really like this idea. I've put a [ticket in the backlog](https://broadinstitute.atlassian.net/browse/GH-946) to track it since there's a lot of moving parts. I'd prefer to get this working to save us time on releases and we can circle back to improve how the scripts are structured.

rexwangcc (approved):
LGTM

PR #141, opened 2020-09-11
[GH-953] get creator from auth token instead of via the workload-request

ehigham:
### Purpose
https://broadinstitute.atlassian.net/browse/GH-954

Interestingly, the keys in the header map are lower case, hence I'm checking for Authorization as well as authorization. Not sure whats cauing this... Ideas welcome.

### Changes
Remove `creator` as a field from the `workload-request`
Get this information from the `userinfo` gcs oauth api.

Added automated test.

### Review Instructions
@samanehsan  Is this change going to cause problems for the cloud function?

### DIFF
--- a/api/src/wfl/service/gcs.clj
+++ b/api/src/wfl/service/gcs.clj
@@ -271,3 +271,13 @@
            :headers      (once/get-auth-header)
            :body         (json/write-str metadata :escape-slash false)}
           http/request :body (json/read-str :key-fn keyword)))))
+
+(defn userinfo
+  "Query Google Cloud services for who made the http REQUEST"
+  [request]
+  (if-let [auth-token (or (get-in request [:headers "Authorization"])
+                          (get-in request [:headers "authorization"]))]
+    (let [response (http/get (str api-url "oauth2/v3/userinfo")
+                     {:headers {"Authorization" auth-token}})]
+      (json/read-str (:body response) :key-fn keyword))
+    (throw (IllegalArgumentException. "no auth header in request"))))
--- a/ops/server.sh
+++ b/ops/server.sh
@@ -9,6 +9,7 @@ test "$1" && export WFL_DEPLOY_ENVIRONMENT="$1"
 npm run serve --prefix=derived/ui -- --port 8080 &
 
 pushd api
+export _JAVA_OPTIONS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"
 "./wfl" server 3000 &
 popd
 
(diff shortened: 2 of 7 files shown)


api/src/wfl/service/gcs.clj
```diff
@@ -271,3 +271,13 @@
            :headers      (once/get-auth-header)
            :body         (json/write-str metadata :escape-slash false)}
           http/request :body (json/read-str :key-fn keyword)))))
+
+(defn userinfo
```
ehigham, line 275: Maybe split this into two functions, `who-made` taking a request, and `userinfo` taking an authorisation header?

ops/server.sh
```diff
@@ -9,6 +9,7 @@ test "$1" && export WFL_DEPLOY_ENVIRONMENT="$1"
 npm run serve --prefix=derived/ui -- --port 8080 &
 
 pushd api
+export _JAVA_OPTIONS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"
```
jack-r-warren, line 12: I've never used server.sh before, I think it is just for testing--I'm not quite sure what's going on here, are these options we'll want to port over into our helm stuff for prod?

jack-r-warren (approved):

ops/server.sh
ehigham: You're right, it's just for testing and I wouldn't port this to a k8s deployment (except maybe for a sandbox). This line is so you can attach a debugger and we don't wanna do this for production deployments.

api/src/wfl/service/gcs.clj
ehigham: Won't do - I think it's unnecessary on reflection.

samanehsan (approved):
LGTM! The cloud function service account requests the [email scope](https://github.com/broadinstitute/wfl/blob/master/cloud_function/main.py#L21) so I don't think we'll have problems there.

ehigham:
> LGTM! The cloud function service account requests the [email scope](https://github.com/broadinstitute/wfl/blob/master/cloud_function/main.py#L21) so I don't think we'll have problems there.

I've probably misunderstood - does this give the cloud function service account a valid email address or does it make it query-able by the cloud? If so, should I remove this string?

samanehsan:
> Thanks, quick clarification please: does this give the cloud function service account a valid email address or does it make it query-able by the cloud? If so, should I remove this string?

So I believe the bearer token needs to be created with at least the email scope in order to get the associated user's email from the userinfo endpoint. So nothing needs to change!

rexwangcc (approved):

PR #142, opened 2020-09-11
[GH-957] Update docs

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-957

### Changes

- Removed outdated docs (such as those related to GAE)
- Re-organize the sections.
- Clean up readme and add the quickstart which covers most of the basics.

### Review Instructions

- Look at https://broadinstitute.github.io/wfl/


docs/md/README.md
```diff
@@ -13,68 +13,166 @@ is a workload manager.
...
 $ make [MODULE] TARGET={prebuild|build|check|images|clean|distclean}
 ```
 
-For most of the time, you would want to run:
+where currently available `MODULE`s are {api cloud_function docs helm ui}
+
+For most of the time, you would want to run something like:
+
+```bash
+$ make distclean
```
ehigham, line 87: Note that distclean also removes 2p and 3p libraries too. I don't think this is always needed for common dev work. clean should be sufficient as it removes all derived module files. 

Perhaps create a jira to better document make targets and assign it to me.

ehigham: Also cleaning in parallel has nice speedups if you're running off an ssd or ramdisk

docs/md/README.md
```diff
@@ -13,68 +13,166 @@ is a workload manager.
...
+
+```bash
+$ make ui api TARGET=images -j8
+```
+
+to **only build** the WFL and its docker images without running tests.
+
+### Test
+
+If you only want to run tests on specific modules, you could run:
```
ehigham, line 101: This will make all of check's prerequisites too

docs/md/README.md
```diff
@@ -13,68 +13,166 @@ is a workload manager.
...
+
+```shell
+./ops/server.sh
 ```
-to only build the WFL and its docker images without running tests.
 
-See the output of `make help` for more information.
+### Deploy
+
+Currently, we mainly deploy WFL to gotc-dev and gotc-prod projects.
```
jack-r-warren, line 145: We can be a bit more accurate here and say `broad-gotc-dev` and `broad-gotc-prod` (though we don't actually deploy to prod at all right now, I'm assuming this documentation is just forward-looking)

docs/md/README.md
```diff
@@ -13,68 +13,166 @@ is a workload manager.
...
-| Cromwell                    | x           | x        |
-| Google App Engine           | x           | x        |
-| Google Cloud Platform Admin | x           | x        |
-| Google Cloud Storage        | x           | x        |
-| Oracle DB                   | x           |          |
-| Vault                       | x           | x        |
+### Diagnosis
 
-Workflow Launcher has a diagnostic mode, `dx`,
+Workflow Launcher has a diagnostic command, `dx`,
```
ehigham, line 175: I wonder if we should make this undocumented as this really only makes sense in command like mode (aka zero). I'm unsure if we want to keep this for a server?

docs/md/dev-process.md
```diff
@@ -6,7 +6,55 @@ within the team,
...
+
+Find a local Cursive user for guidance if you like IntelliJ.
+[Rex Wang](mailto:chengche@broadinstitute.org) and
+[Saman Ehsan](mailto:sehsan@broadinstitute.org) know how to use it.
+Cursive licences are available
+[here](https://broadinstitute.atlassian.net/wiki/spaces/DSDE/pages/48234557/Software%2BLicenses%2B-%2BCursive).
+The steps for getting this project set up with very recent versions of IntelliJ
+differ from Cursive's docs:
+
+1. *Outside of IntelliJ*, `clone` the repo and run `boot` at the top-level to
```
ehigham, line 33: I'd recommend running prebuild before launching intellij as this sets up all libraries and derived resources and sources. 

```bash
$ make TARGET=prebuild - jN
```

jack-r-warren (approved):
Okay this is amazing. Literally learned a ton of new things about make and WFL just by reading through your changes.

ehigham (approved):
Really nice work, rex!

docs/md/README.md
rexwangcc: This is a really good point! I hesitated about this a bit, but now I think it makes sense to remove it from here. (At least if we wanna keep it, it should be in the dev section not readme.)

PR #144, opened 2020-09-15
[GH-946] Encode ops/cli deps into requirements file

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-946

### Changes

- Encode ops/cli deps into requirements file.
- Update the release Github Action accordingly
- Update the [Jenkins Job](https://gotc-jenkins.dsp-techops.broadinstitute.org/job/deploy-wfl) accordingly

### Review Instructions

- No instructions.


.github/workflows/release_latest_version.yml
```diff
@@ -11,11 +11,11 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      MODULES: api ui
+      MODULES: api ui ops
```
samanehsan, line 14: Does making the `ops` module create and activate the python environment?

rexwangcc: Yes! It's being handled by makerules here: https://github.com/broadinstitute/wfl/blob/6d1e2472f43ddf2baea932de0fcb8dcd4c7064dc/makerules/Python.mk#L24

samanehsan: Oh I see!

samanehsan (approved):
LGTM!

ehigham (commented):
I really don't think making this a makefile module is the right thing to do here - they're just deployment scripts.

rexwangcc:
> I really don't think making this a makefile module is the right thing to do here - they're just deployment scripts.

Do you think it's simpler to just add the `requirements.txt` file and update CIs to `pip install -r ` it? That was my original plan, but again, adding a make module is what the ticket instructed me to do, and I don't see it as a huge problem.

ehigham:
> > I really don't think making this a makefile module is the right thing to do here - they're just deployment scripts.
> 
> Do you think it's simpler to just add the `requirements.txt` file and update CIs to `pip install -r ` it? That was my original plan, but again, adding a make module is what the ticket instructed me to do, and I don't see it as a huge problem.

Absolutely. I think this would be the best solution.

samanehsan (approved):
LGTM! I like the simplification here!

PR #146, opened 2020-09-16
[GH-945] Add workload UUID as labels to workflows that WFL starts

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-945

### Changes

- Add workload UUID as labels to workflows that WFL starts. (I don't really like the top-down approach to pass down the labels as params between functions, but cannot think of a better way to deal with this while maintaining the simplicity).
- Some minor clean-ups, e.g. remove comment blocks, remove unused imports.

---
While cleaning up the code of the modules (aou, wgs), I came up with a relatively crazy idea that we might just compose a huge map of stuff (similar to `environments.clj`) that defines each of the data's model/schema, instead of defining functions like `fingerprint` or `per-sample-inputs`. This will turn the philosophy of WFL modules from FP to a more "data-driven" model, thus the modules will become more language/clj-agnostic and data-processors rather than just "one-off pipeline adaptors". Please let me know if you think this is not that crazy.

a short demo PR: https://github.com/broadinstitute/wfl/pull/149

---

### Review Instructions

- Take a look at the workflows started during my local integration testing: https://jmui.gotc-dev.broadinstitute.org/jobs/d1c4e65a-1e14-4aa3-a554-4788238592ee, the labels section now has `workload` field.
- Take a look at https://jmui.gotc-dev.broadinstitute.org/jobs, notice there's a new column called `WFL workload`.


samanehsan (approved):
LGTM! :octocat:

samanehsan:
> While cleaning up the code of the modules (aou, wgs), I came up with a relatively crazy idea that we might just compose a huge map of stuff (similar to environments.clj) that defines each of the data's model/schema, instead of defining functions like fingerprint or per-sample-inputs. This will turn the philosophy of WFL modules from FP to a more "data-driven" model, thus the modules will become more language/clj-agnostic and data-processors rather than just "one-off pipeline adaptors".

I think your idea sounds interesting! It could be nice to consolidate some of these default specifications.

api/src/wfl/module/aou.clj
```diff
@@ -168,7 +166,7 @@
         {:keys [commit version]} (wfl/get-the-version)
         probe-result (jdbc/query tx ["SELECT * FROM workload WHERE project = ? AND pipeline = ?::pipeline AND release = ? AND output = ?"
                                      project pipeline release output])
-        checked-output (gcs/parse-gs-url output)]
+        _            (gcs/parse-gs-url output)]
```
tbl3rd, line 169: Move this out of the bindings.

api/src/wfl/module/aou.clj
```diff
@@ -199,18 +197,10 @@
...
-                                        :cromwell "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org/"
-                                        :project  "gotc-dev"
-                                        :pipeline "AllOfUsArrays"}]
-                              (add-workload! tx body))))
-
 (defn start-workload!
   "Use transaction TX to start the WORKLOAD by UUID. This is simply updating the
    workload table to mark a workload as 'started' so it becomes append-able."
-  [tx {:keys [uuid] :as body}]
+  [tx {:keys [uuid]}]
```
tbl3rd, line 203: The other start-workload! methods take a workload.

api/src/wfl/module/aou.clj
```diff
@@ -240,34 +230,18 @@
...
-                                            :gender_cluster_file         "gs://broad-gotc-test-storage/arrays/metadata/HumanExome-12v1-1_A/HumanExomev1_1_gender.egt"
-                                            :extended_chip_manifest_file "gs://broad-gotc-test-storage/arrays/metadata/HumanExome-12v1-1_A/HumanExome-12v1-1_A.1.3.extended.csv"}]]
-                              (remove-existing-samples! tx samples "allofusarrays_000000001"))))
-
 (defn append-to-workload!
   "Use transaction TX to append the samples to a WORKLOAD identified by uuid.
    The workload needs to be marked as 'started' in order to be append-able, and
    any samples being added to the workload table will be submitted right away."
-  [tx {:keys [notifications uuid environment] :as body}]
+  [tx {:keys [notifications uuid environment]}]
```
tbl3rd, line 237: Where is the WORKLOAD?

api/src/wfl/module/aou.clj
```diff
@@ -297,43 +271,6 @@
                      :uuid                    uuid
                      :analysis_version_number analysis_version_number
                      :chip_well_barcode       chip_well_barcode}))]
-          (let [inserted            (jdbc/insert-multi! tx table (map sql-rize (vals to-be-submitted)))
+          (let [_                   (jdbc/insert-multi! tx table (map sql-rize (vals to-be-submitted)))
```
tbl3rd, line 274: Again, move this out of the let bindings.

api/src/wfl/module/wgs.clj
```diff
@@ -112,6 +112,12 @@
     (all/throw-when-output-exists-already! output)
     (util/prefix-keys inputs :ExternalWholeGenomeReprocessing)))
 
+(defn make-labels
+  "Return labels for wgs pipeline from OTHER-LABELS."
+  [other-labels]
+  (merge cromwell-label-map
+         other-labels))
```
tbl3rd, line 119: It looks like other-labels gets passed around a lot.
Maybe this can be factored to be more direct?
Start accreting labels where the UUID is generated?

tbl3rd (approved):
Suggest adding label to workload where UUID is generated.
Then augment the labels as necessary until input to Cromwell.

api/src/wfl/module/aou.clj
tbl3rd: Suggest changing `body` to `_workload` to parallel other `…-workload!` methods.

api/src/wfl/module/aou.clj
rexwangcc: thanks, will do 👍

api/src/wfl/module/aou.clj
rexwangcc: thanks, will do 👍

api/src/wfl/module/aou.clj
rexwangcc: will do 👍

rexwangcc:
Close in favor of #175

PR #148, opened 2020-09-16
[GH-948] Fix "output already exists" error in WGS tests

samanehsan:
### Purpose
Fixes: https://broadinstitute.atlassian.net/browse/GH-948

### Changes
- Make the WGS test workload output directory different for each test run. Otherwise WFL will not start the workload because there's an output already in that directory.
- Make the integrations tests use the api server when running locally

### Review Instructions

- No instructions.

### DIFF
--- a/api/test/wfl/tools/endpoints.clj
+++ b/api/test/wfl/tools/endpoints.clj
@@ -7,7 +7,7 @@
   "The WFL server URL to test."
   (if (System/getenv "WFL_DEPLOY_ENVIRONMENT")
     "https://dev-wfl.gotc-dev.broadinstitute.org"
-    "http://localhost:8080"))
+    "http://localhost:3000"))
 
 (defn parse-json-string
   "Parse the json string STR into a keyword-string map"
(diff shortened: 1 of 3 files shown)


api/test/wfl/tools/endpoints.clj
```diff
@@ -7,7 +7,7 @@
   "The WFL server URL to test."
   (if (System/getenv "WFL_DEPLOY_ENVIRONMENT")
     "https://dev-wfl.gotc-dev.broadinstitute.org"
-    "http://localhost:8080"))
+    "http://localhost:3000"))
```
rexwangcc, line 10: do we still need to do this since #147 is merged in?

samanehsan: I was thinking since these are API tests, it shouldn't require the UI server to be running so I changed it to use the API server's port directly. What do you think?

rexwangcc: I see, that totally makes sense!

rexwangcc (approved):

PR #150, opened 2020-09-18
[GH-968] 0.1.8 Hot-Fix (#150)

rexwangcc:
## ~DO NOT MERGE THIS PR~

- This is only for review purposes! 
- ~`virtual-0.1.8` should be delete after this PR gets reviewed.~
- ~`0.1.8-hotifx` should be deployed directly once this PR gets reviewed.~

The @jack-r-warren plan:

- [ ] Merge the PR, tag-and-push-images on the destination branch
- [ ] https://gotc-jenkins.dsp-techops.broadinstitute.org/job/deploy-wfl/build?delay=0sec deploy to gotc-dev
- [ ] https://gotc-jenkins.dsp-techops.broadinstitute.org/job/aou-ptc-wfl-end-to-end-dev/ run this test manually some
- [ ] Maybe we look at the container logs and/or connect gotc-dev dev to look at the DB to verify the changes?
- [ ] https://gotc-jenkins.dsp-techops.broadinstitute.org/job/deploy-wfl/build?delay=0sec deploy to aou
- [ ] Watch prod cloud function failures--we should see none

### Purpose

- https://broadinstitute.atlassian.net/browse/GH-968

### Changes

- Only try to update workflows statuses in `get-workload`. `start-workload` does not update them anymore, and only gets a snapshot of the current DB.
- Update workflows statuses in `get-workload` in a separate thread in a different transaction, so the request is not blocking.

### Review Instructions

- I have run tests on this branch. Most of the tests passed except one that was already fixed by a later PR: #132 , so it's not "fixed" again in this PR.


api/src/zero/service/postgres.clj
```diff
@@ -70,7 +70,7 @@
   [tx {:keys [cromwell id items] :as _workload}]
   (try
     (let [workflows (get-table tx items)]
-      (run! (partial update-workflow-status! tx cromwell items) workflows)
+      (pmap (partial update-workflow-status! tx cromwell items) workflows)
```
rexwangcc, line 73: I'm not sure about this:
- whether we should do this?
- whether this makes sense?

api/src/zero/api/handlers.clj
```diff
@@ -127,20 +128,25 @@
...
+                             (logr/infof "get-workload endpoint called: uuid=%s" uuid)
+                             [{:uuid uuid}])
+                           (do
+                             (logr/info "get-workload endpoint called with no uuid, querying all")
+                             (jdbc/query tx ["SELECT uuid FROM workload"])))
+                      (mapv (partial postgres/get-workload-for-uuid tx)))]
+       ;;sync the workflow statuses with cromwell in another thread
+       (future
+         (jdbc/with-db-transaction [another-tx (postgres/zero-db-config)]
+         (util/do-or-nil (postgres/update-workload-for-uuid! another-tx uuid-vec))))
```
rexwangcc, line 147: not sure if this is anti-pattern 🤔  also not sure if this is "clojure-diomatic"

ehigham: I'm not sure why you've moved the update into a future.... The whole point of this is to get a snapshot at the point of call of the `workload` endpoint.

api/src/zero/service/postgres.clj
ehigham: Is there any performance advantage to this?
I'd suggest not making threading changes for a hot-fix.

api/src/zero/api/handlers.clj
rexwangcc: > The whole point of this is to get a snapshot at the point of call of the workload endpoint.

OK, I thought this "snapshot" is a snapshot of the current DB, rather than the snapshot of the true statuses in Cromwell. The use of `future` is purely based on the observation that the update process took ~7min to finish, which would absolutely fail the request if this was synchronous.

rexwangcc:
Rebased off of `v0.1.10`.

samanehsan (approved):
LGTM! :octocat:

jack-r-warren (approved):
Looks good. Because we aren't merging this into master, the automatic release job won't run. You will need to build locally and run `./ops/cli.py tag-and-push-images` on the destination branch as soon as this is merged. That'll make it appear in the deployment Jenkins job

ehigham (approved):

PR #151, opened 2020-09-21
[GH-949] Consolidate test buckets

samanehsan:
### Purpose

Ticket: https://broadinstitute.atlassian.net/browse/GH-949

### Changes
- Make `with-temporary-gcs-folder` use the new test output bucket `broad-gotc-dev-wfl-ptc-test-outputs` that has an object lifecycle policy of 1 day
- Make the gcs-object test use `with-temporary-gcs-folder` instead of creating new buckets
- Make the data-repo test use `with-temporary-gcs-folder`instead of creating a new bucket
- Remove `bucket-test` for creating, listing & deleting buckets
- Remove gcs `make-bucket` and `delete-bucket` since they are no longer used after these changes (they were previously only used by the tests)
- Updated WGS and AOU test workloads inputs to use `broad-gotc-dev-wfl-ptc-test-inputs`
- Updated cloud function helper script to use AOU test inputs from `broad-gotc-dev-wfl-ptc-test-inputs`

### Review Instructions

The data-repo test makes more sense as an integration test, since it's not testing individual functions. Thoughts on moving it?

### DIFF
--- a/api/test/wfl/unit/gcs_test.clj
+++ b/api/test/wfl/unit/gcs_test.clj
@@ -3,8 +3,9 @@
   (:require [clojure.java.io :as io]
             [clojure.string :as str]
             [clojure.test :refer [deftest is testing]]
+            [wfl.once :as once]
             [wfl.service.gcs :as gcs]
-            [wfl.once :as once])
+            [wfl.tools.fixtures :refer [with-temporary-gcs-folder]])
   (:import [java.util UUID]))
 
 (def project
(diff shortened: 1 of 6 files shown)


api/test/wfl/unit/gcs_test.clj
```diff
@@ -57,32 +41,9 @@
...
-      (testing "list project prefix"
-        (let [result (gcs/list-buckets project prefix)]
-          (is (= 1 (count result)))
-          (is (= bucket (:name (first result))))))
-      (testing "delete"
-        (is (gcs/delete-bucket bucket))
-        (is (not (gcs/delete-bucket bucket)))))))
+(def blame
+  "Who is to blame?"
+  (or (System/getenv "USER") "wfl"))
```
rexwangcc, line 46: do we still need this function?

rexwangcc (approved):
I have been waiting for a test clean up for so long! 🎉  thanks for doing this!!!

> object lifecycle policy of 1 day
(I remember @pshapiro4broad  was asking this somewhere, are we sure we want 1 day over 1 week here?)

api/test/wfl/unit/gcs_test.clj
samanehsan: Good point! It's used [here](https://github.com/broadinstitute/wfl/blob/37a92e7abe9b93c000e5717515583df079843064/api/test/wfl/unit/gcs_test.clj#L50) but it's just for the local file name. I can change that file name prefix to "wfl" instead.

rexwangcc: Oh I see! I think it makes sense to keep it. I just didn't realized it's still being used!

samanehsan: I just removed it since it's only used for the local files in the test (previously it was also used as the bucket name so you could tell who created it)

rexwangcc: Sounds great!

samanehsan:
The tests are hanging again 🙀

rexwangcc:
🙀  I think it's safe to merge this given the previous run passed. But we should really look into why it hangs ...

PR #155, opened 2020-09-24
[GH-987] Split system and integration tests

samanehsan:
### Purpose
Before automating integration tests to run on PRs, move the time-consuming tests into a separate system test alias. 

Ticket: https://broadinstitute.atlassian.net/browse/GH-987

### Changes
Move every test that waits for a Cromwell workflow to complete under a new system test alias.

### Review Instructions

Notes: 
- The append to AoU test succeeds before the workflow actually finishes (looks like a bug) so even though it currently completes very quickly I'm not sure if it would once we fix it (created a ticket for this here: https://broadinstitute.atlassian.net/browse/GH-995)
- The WGS create test waits for a Cromwell workflow to succeed, but it actually completes very quickly due to call-caching so I wasn't sure where to put that one.

### DIFF
--- a/api/test/wfl/integration/v1_endpoint_test.clj
+++ b/api/test/wfl/integration/v1_endpoint_test.clj
@@ -1,13 +1,9 @@
 (ns wfl.integration.v1-endpoint-test
   (:require [clojure.string :as str]
             [clojure.test :refer [deftest testing is] :as clj-test]
-            [wfl.module.copyfile :as cp]
-            [wfl.service.cromwell :refer [wait-for-workflow-complete]]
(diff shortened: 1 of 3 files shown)


ehigham (approved):
looks good. Generally speaking, where do you recommend drawing the line between an integration and a system test?

jack-r-warren (approved):

samanehsan:
> Generally speaking, where do you recommend drawing the line between an integration and a system test?

I think a system test in WFL would be anything that launches a workload and then waits for the workflows to succeed in Cromwell. And an integration test would be anything that relies on an external service, but is not necessarily a test of the end-to-end functionality.

PR #156, opened 2020-09-25
[GH-886] Have OAuth2 Client ID be handled by the backend

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-886

### Changes

- Add an unauthenticated endpoint to WFL to return the OAuth2 Client ID correlating with WFL's environment (getting it from environment if possible, then from Vault, then just the dev one for backwards compatibility)
- Add a test for said endpoint
- Modify UI to call said endpoint to get the client ID before running the OAuth2 sequence
- Add references to the environment variable and endpoint to the Helm config

### Review Instructions

- Run integration tests
- Run a local server and see that you can still log in--this verifies that WFL's endpoint returned the OAuth2 Client ID it got from vault and that the login sequence still works
- Go to https://jwarren-wfl.gotc-dev.broadinstitute.org/oauth2id to verify that the endpoint works unauthenticated on a real deployment (that's my own sandbox running off of the exact same wfl-instance and deployment setup that dev and prod are on)
    - If you try to log in on my sandbox, it will fail _intentionally_: the old default of the dev ID (still used as a fallback) actually works on my sandbox, so for this deployment I manually edited the values file to use the prod ID. The fact that it fails with a "hey! you're on the wrong URL to be using this ID!" error is super important and actually means the whole thing is working, because it means that the prod ID got used during the log in process like I asked. If you want to see the log in process work, just run a local server or let me know.


jack-r-warren:
Also see [this gotc-deploy PR](https://github.com/broadinstitute/gotc-deploy/pull/176/files) that's necessary for this change to do anything more than the current "always use the dev ID" behavior. As stated in that PR, these can be merged out of order, both are fully backwards-compatible.

api/src/wfl/once.clj
```diff
@@ -52,3 +52,24 @@
     (if (util/getenv "WFL_DEPLOY_ENVIRONMENT")
       (service-account-token)
       (util/shell! "gcloud" "auth" "print-access-token"))))
+
+(defn- read-oauth-client-id
+  "Actively read the OAuth client ID from vault or the environment"
```
ehigham, line 57: As opposed to passively read?

api/src/wfl/once.clj
```diff
@@ -52,3 +52,24 @@
...
+                   wfl/error-or-environment-keyword
+                   env/stuff :server :vault
+                   util/vault-secrets :oauth2_client_id))]
+      id-from-vault
+      ;; Client ID for gotc-dev, the old hardcoded value for backwards-compatibility
+      "450819267403-n17keaafi8u1udtopauapv0ntjklmgrs.apps.googleusercontent.com")))
+
+(def return-oauth-client-id
+  "Memoize the reading of the OAuth client ID to be more responsive"
+  (memoize read-oauth-client-id))
```
ehigham, line 75: I was wondering when to use `memoize` vs `delay`. This link seems interesting:
https://stackoverflow.com/questions/60760896/clojure-when-to-use-memoize-and-when-to-use-delay-force

Out of interest, what was your reasoning for one over the other?

ehigham (approved):

api/src/wfl/once.clj
jack-r-warren: The reasoning is that I didn't know delay existed lol

api/src/wfl/once.clj
jack-r-warren: Just wanted to make it clear that this actually read stuff, guess it is moot now

jack-r-warren:
Rewrote using `delay` rather than `memoize`

api/src/wfl/once.clj
ehigham: Haha!

api/src/wfl/once.clj
tbl3rd: Let the JVM memoize for you. Try memoize if that fails.

tbl3rd: https://github.com/tbl3rd/clj-intro/blob/05508471656578630dd3f9b00671dbc4e2dfe527/src/017-deref.clj#L213-L239

PR #158, opened 2020-09-28
[GH-982] Move datarepo test to integration and mark it as pending

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-982

### Changes

- Move datarepo test to integration and mark it as pending

### Review Instructions

- Run `clojure -A:test unit` and make sure `datarepo` test is not run anymore.

### DIFF
--- a/api/tests.edn
+++ b/api/tests.edn
@@ -1,6 +1,6 @@
 #kaocha/v1
     {:tests [{:id           :integration
-              :skip-meta    [:pending]
+              :skip-meta    [:exclude]
               :source-paths ["../derived/api/src" "src"]
               :test-paths   ["test"]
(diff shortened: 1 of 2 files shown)


api/tests.edn
```diff
@@ -1,5 +1,6 @@
 #kaocha/v1
     {:tests [{:id           :integration
+              :skip-meta    [:pending]
```
ehigham, line 3: This is quite a nice feature. Can we use something like `exclude` or `filter` - `pending` doesn't quite capture the general case i think!

rexwangcc: Thanks! I think `exclude` is a better name, will update.

samanehsan (approved):
LGTM!

PR #166, opened 2020-10-05
[GH-1001] No workloads still succeeds

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1001

### Changes

- Don't fail if nothing was found

### Review Instructions

```bash
curl -X GET "https://jwarren-wfl.gotc-dev.broadinstitute.org/api/v1/workload" -H "accept: application/json" -H "Authorization: Bearer $(gcloud auth print-access-token)"
```


jack-r-warren:
A side note: I spent like 2 hours digging around Swagger stuff to try to get the OAuth thing working. Basically, because [OAuth configuration in the UI happens via a method, not via a parameter](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md#instance-methods), Reitit is [unable to configure it](https://github.com/metosin/reitit/blob/master/modules/reitit-swagger-ui/src/reitit/swagger_ui.cljc#L15). Specifically, the [`index.html` that Reitit serves when you use `ring-swagger-ui`](https://github.com/metosin/ring-swagger-ui/blob/master/resources/swagger-ui/index.html) does not do any oauth configuration as described [here](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/oauth2.md), even if literally everything else is configurable via the config.json it loads in: the method can't be called. The repo for `ring-swagger-ui` seems stale too. 

We can solve this problem by ditching `ring-swagger-ui` and just keep using [`ring-swagger`](https://github.com/metosin/ring-swagger) directly to still make the spec from the backend (nothing would change), and and change to use [`swagger` via NPM](https://www.npmjs.com/package/swagger-ui) in our frontend (where we can configure everything we like and make use of the OAuth2 endpoint we now have to grab the client ID).

@rexwangcc (Since I think you looked into this a while back)

api/src/wfl/api/handlers.clj
```diff
@@ -146,7 +146,7 @@
                            (jdbc/query tx ["SELECT uuid FROM workload"]))))]
        (util/do-or-nil (postgres/update-workload-for-uuid! tx uuid-vec))
        (let [result (mapv (partial postgres/get-workload-for-uuid tx) uuid-vec)]
-         (if (every? empty? result)
+         (if (and (not-empty result) (every? empty? result))
```
samanehsan, line 149: How about only calling `update-workload-for-uuid!` and `get-workload-for-uuid` if `uuid-vec` is not empty? That might make this logic a bit clearer 🤔

jack-r-warren: Good idea!

ehigham (approved):

PR #168, opened 2020-10-06
[GH-1007] Check WGS workload values before using defaults

samanehsan:
### Purpose

Ticket: https://broadinstitute.atlassian.net/browse/GH-1007

### Changes

Use the `base_file_name`, `final_gvcf_base_name` or `sample_name` if they are specified in the workload, otherwise use the default input cram name.

### Review Instructions

Should "sample_name" actually be an optional field?

### DIFF
--- a/api/src/wfl/api/spec.clj
+++ b/api/src/wfl/api/spec.clj
@@ -84,21 +84,21 @@
 (s/def ::items-wgs (s/keys :opt-un [::base_file_name
                                     ::final_gvcf_base_name
                                     ::reference_fasta_prefix
+                                    ::sample_name
                                     ::unmapped_bam_suffix]
-                           :req-un [::input_cram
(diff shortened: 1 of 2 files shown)


jack-r-warren (approved):
LGTM. Re the review question I think if WFL can handle things without a name then it shouldn't put additional arbitrary restrictions (for all we know the field might end up being used as an optional description at some point, it could just be poorly named)

samanehsan:
> If WFL can handle things without a name then it shouldn't put additional arbitrary restrictions

That makes sense! I changed the `sample_name` to optional to match the module behavior.

samanehsan:
@jack-r-warren would you mind taking another look? I made some more changes!

jack-r-warren (approved):
LGTM

PR #169, opened 2020-10-06
[GH-1009] Change to environment files

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1009

### Changes

- Use the new environment files

### Review Instructions

- This should result in no new test failures (this is a no-op change)


rexwangcc (approved):
Looks right to me. The test also passed. so 👍

PR #171, opened 2020-10-06
[GH-1005] fix api/v1/workload endpoint; simplify endpoints; do exec i…

ehigham:
### Purpose
Fixes: https://broadinstitute.atlassian.net/browse/GH-1005 - "Table null not found"

Included changes:
- `post-exec` now uses a single transaction
- simplified `post-start`

Exposed a number of code improvements we can make. Will commit to code in the exomes endpoint.

### DIFF
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -1,7 +1,6 @@
 * (ns wfl.api.handlers
     "Define handlers for API endpoints"
-    (:require [clojure.string :as str]
-              [clojure.tools.logging :as log]
+    (:require [clojure.tools.logging :as log]
               [clojure.tools.logging.readable :as logr]
               [ring.util.response :as response]
               [wfl.module.aou :as aou]
@@ -11,7 +10,6 @@
               [wfl.service.cromwell :as cromwell]
               [wfl.service.postgres :as postgres]
               [wfl.wfl :as wfl]
-              [wfl.util :as util]
               [wfl.service.gcs :as gcs]))
 
 (defn fail
@@ -29,23 +27,6 @@
      (catch Exception e#
        (log/error e#)
        (fail e#))))
-
-(defn fail-with-response
-  "A failure RESPONSE with BODY."
-  [response body]
-  (log/warn "Endpoint sent a failure response:")
-  (log/warn body)
-  (-> body response (response/content-type "application/json")))
-
-(defn unprocessable-entity
-  "Returns a 422 'Unprocessable Entity' response with BODY."
-  [body]
-  (let [response {:status  422
-                  :headers {}
-                  :body    body}]
-    (log/warn "Endpoint sent a 422 failure response:")
-    (log/warn body)
-    response))
 
 (defn succeed
   "A successful response with BODY."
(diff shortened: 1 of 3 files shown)


api/src/wfl/api/handlers.clj
```diff
@@ -118,65 +112,72 @@
...
+                (logr/infof "updating workload %s" uuid)
+                (postgres/update-workload! tx workload)
+                (postgres/load-workload-for-id tx id))
+              workload))]
+    (log-and-fail-on-error
+      (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+        (succeed
+          (mapv (partial go! tx)
+            (if-let [uuid (-> request :parameters :query :uuid)]
+              (if-let [workload (postgres/load-workload-for-uuid tx uuid)]
```
ehigham, line 142: @tbl3rd any suggestions for combining the nested if?

api/src/wfl/api/handlers.clj
```diff
@@ -118,65 +112,72 @@
...
   "Create and start workload described in BODY of REQUEST"
-  (let [pack   (fn [response] {:parameters {:body [(:body response)]}})
-        unpack (comp first :body)]
-    (comp succeed unpack post-start pack post-create)))
+  [request]
+  (letfn [(create! [tx workload-request]
+            (->>
+              (add-workload! tx workload-request)
+              :uuid
+              (postgres/load-workload-for-uuid tx)))
```
ehigham, line 170: I think add-workload should return the created workload rather than just its UUID.

api/src/wfl/api/handlers.clj
```diff
@@ -118,65 +112,72 @@
...
+          (start! [tx workload]
+            (start-workload! tx workload)
+            (postgres/load-workload-for-id tx (:id workload)))
```
ehigham, line 173: Likewise, `start-workload` should return its started workload

api/src/wfl/api/handlers.clj
tbl3rd: Works for me.

api/src/wfl/api/handlers.clj
tbl3rd: Given they have distinct alternates, nested `if-let` seems the way to go.

api/src/wfl/api/handlers.clj
```diff
@@ -118,65 +92,74 @@
...
-
-(def post-exec
+  (letfn [(go! [tx {:keys [uuid]}]
+            (if-let [workload (postgres/load-workload-for-uuid tx uuid)]
+              (if-not (:started workload)
+                (do
+                  (logr/infof "starting workload %s" uuid)
+                  (start-workload! tx workload)
+                  (postgres/load-workload-for-id tx (:id workload)))
+                workload)
```
tbl3rd, line 137: Why not?
```clojure
(if (:started workload) workload
    (do
      (logr/infof "starting workload %s" uuid)
      (start-workload! tx workload)
      (postgres/load-workload-for-id tx (:id workload))))
```

api/src/wfl/api/handlers.clj
```diff
@@ -118,65 +92,74 @@
...
-    (comp succeed unpack post-start pack post-create)))
+  [request]
+  (letfn [(create! [tx workload-request]
+            (let [{:keys [uuid]} (add-workload! tx workload-request)]
+              (log/infof "created workload %s" uuid)
+              (postgres/load-workload-for-uuid tx uuid)))
+          (start! [tx {:keys [id uuid] :as workload}]
+            (log/infof "starting workload %s" uuid)
+            (start-workload! tx workload)
+            (postgres/load-workload-for-id tx id))
```
tbl3rd, line 155: I think you want (at least) a `(defn start-workload …)`.

ops/server.sh
```diff
@@ -9,6 +9,7 @@ test "$1" && export WFL_DEPLOY_ENVIRONMENT="$1"
 npm run serve --prefix=derived/ui -- --port 8080 &
 
 pushd api
+clojure -A:liquibase
```
tbl3rd, line 12: Hope that works.

api/src/wfl/api/handlers.clj
tbl3rd: Why `mapv` instead of `map`?

tbl3rd (approved):

api/src/wfl/api/handlers.clj
ehigham: Because i need to force evaluation before the transaction exits

ops/server.sh
ehigham: It does :)

api/src/wfl/api/handlers.clj
ehigham: For posterity:
There are two `start!` letfns which look almost identical.
The plan is that this will be absorbed into the `workload` "class"

PR #172, opened 2020-10-07
GH-651: Automate integration test.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-651

### Changes

- No changes.

### Review Instructions

- No instructions.


api/module.mk
```diff
@@ -36,11 +36,19 @@ $(BUILD): $(SCM_SRC) $(SCM_RESOURCES)
 	@$(TOUCH) $@
 
 LOGFILE := $(DERIVED_MODULE_DIR)/test.log
-$(CHECK): $(SCM_SRC) $(SCM_RESOURCES) $(CLOJURE_PROJECT)
+$(UNIT): $(SCM_SRC) $(SCM_RESOURCES) $(CLOJURE_PROJECT)
```
ehigham, line 39: You need to rename this in `modules.mk`, both the timestamp and the phony target

api/module.mk
```diff
@@ -36,11 +36,19 @@ $(BUILD): $(SCM_SRC) $(SCM_RESOURCES)
...
 	$(EXPORT) CPCACHE=$(CPCACHE_DIR);                        \
 	$(CLOJURE) $(CLJFLAGS) -A:test unit | $(TEE) $(LOGFILE)
 	@$(TOUCH) $@
 
+LOGFILE := $(DERIVED_MODULE_DIR)/test.log
+$(INTEGRATION): $(SCM_SRC) $(SCM_RESOURCES) $(CLOJURE_PROJECT)
```
ehigham, line 45: You also need to add this to `modules.mk` and set up the dependencies. Probably depend on `BUILD`.

api/module.mk
tbl3rd: Thanks. Why do we create result files for phony targets?

ehigham: `make` will always try to make `phony` targets. By using a time stamp, we only need to make each target when its dependencies are newer than the time stamp.

api/module.mk
```diff
@@ -35,12 +35,20 @@ $(BUILD): $(SCM_SRC) $(SCM_RESOURCES)
...
+	$(TEE) $(DERIVED_MODULE_DIR)/unit.log
 	@$(TOUCH) $@
 
+$(INTEGRATION): $(SCM_SRC) $(SCM_RESOURCES) $(CLOJURE_PROJECT)
+	$(EXPORT) CPCACHE=$(CPCACHE_DIR);                        \
+	$(CLOJURE) $(CLJFLAGS) -A:test integration | \
+	$(TEE) $(DERIVED_MODULE_DIR)/integration.log
+	@$(TOUCH) $@
+
+$(CHECK): $(UNIT) $(INTEGRATION)
```
ehigham, line 50: I'd move this into  `modules.mk` to share with all modules

ehigham: If you don't indent to execute any steps in `check` (i.e. it's just an alias for `unit` and `integration`) you don't need to worry about the CHECK timestamp. Simply add the following after `makerules/modules.mk:35`:
```
check: unit integration
```

jack-r-warren (commented):
This looks good, but GH-1019 moved most of what I think of as integration tests to be under system tests. The only thing that is an integration test now is a test with the data repo (which doesn't relate to any code we currently run in WFL as I understand it). That's all good, it's just that if we merge this as is we need to be aware that this does not give us any extra test coverage over WFL's code.

ehigham:
> This looks good, but GH-1019 moved most of what I think of as integration tests to be under system tests. The only thing that is an integration test now is a test with the data repo (which doesn't relate to any code we currently run in WFL as I understand it). That's all good, it's just that if we merge this as is we need to be aware that this does not give us any extra test coverage over WFL's code.

It will, when we write integration tests. See the ones I added for the exomes module, for example.

ehigham (approved):

PR #175, opened 2020-10-07
[GH-945] label workload UUID to workflows wfl started again

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-945

### Changes

Re-apply #146

- Add workload UUID as labels to workflows that WFL starts. (I don't really like the top-down approach to pass down the labels as params between functions, but cannot think of a better way to deal with this while maintaining the simplicity).
- Some minor clean-ups, e.g. remove comment blocks, remove unused imports.

### Review Instructions

- No instructions.


tbl3rd (changes requested):
Let's pair on this.

jack-r-warren (approved):

tbl3rd (approved):

PR #177, opened 2020-10-08
[GH-1012] Make the Swagger UI work better

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1012

### Changes

- Swagger UI no longer configured in Clojure, it is now configured in the UI
- Swagger UI lives within a Vue view, not on its own, so it is styled like our other pages now
- Swagger UI OAuth2 prefills the correct ClientID to use
- Swagger UI OAuth2 selects all scopes for auth by default
- Swagger scheme now categorizes `/api` endpoints under "Authenticated" rather than "default" (so we have "Information" and "Authenticated", not "Information" and "default")
- Swagger scheme now marks `/api` endpoints as requiring authentication (so the header gets passed if you're authenticated--**this means that the "try it out" button works for all our endpoints now**)
- Vue's webpack config now copies Swagger UI's oauth redirect page during build time, so that `/oauth2_redirect.html` will always exist (it is basically transparent during normal usage and harmless otherwise, it just needs to be there as a callback URL)
- Documented that WFL deployments now have an extra redirect URI that needs to be authorized in GCP
- I already added that new URI to all of our WFL OAuth credentials (gotc-dev, gotc-prod, and aou), so the documentation just applies to future sandboxes

### Review Instructions

- I have this currently deployed on my sandbox. Head over to https://jwarren-wfl.gotc-dev.broadinstitute.org/ and you can play around with it.


api/src/wfl/api/routes.clj
```diff
@@ -90,31 +89,38 @@
...
+(defn endpoint-swagger-auth-processor
+  "Use the same security-info across all /api endpoints"
+  [endpoints]
+  (let [security-info {:swagger {:tags ["Authenticated"]
+                                 :security [{:googleoauth []}]}}]
+    (letfn [(needs-security? [endpoint] (str/starts-with? (first endpoint) "/api"))
+            (write-security-info [method-description] (merge-with merge security-info method-description))
+            (modify-method [method] (zipmap (keys method) (map write-security-info (vals method))))]
+      (vec (map #(if (needs-security? %) (apply vector (first %) (map modify-method (rest %))) %)
+                endpoints)))))
```
jack-r-warren, line 108: Super interested in feedback on this. Looks at the endpoints data structure above this might help.

ui/src/components/TopBar.vue
```diff
@@ -27,7 +27,7 @@
           <v-col cols="5">
             <v-row justify="end" align="center">
             <v-btn class="mr-2"  text >Status: {{ status.status }}</v-btn>
-            <v-btn class="mr-2"  outlined v-if="isAuthenticated" v-bind:href="swagger_link">Swagger API</v-btn>
+            <v-btn class="mr-2"  outlined v-if="isAuthenticated" v-bind:to="swagger_link">Swagger API</v-btn>
```
jack-r-warren, line 30: Router links are `to`, not `href`, and this is now handled by the router so it should change

ui/src/router/index.js
```diff
@@ -44,6 +44,14 @@ const routes = [
...
     component: () => import(/* webpackChunkName: "modules" */ '../views/Modules.vue')
   },
+  {
+    path: '/swagger',
+    name: 'swagger',
+    // route level code-splitting
+    // this generates a separate chunk (modules.[hash].js) for this route
+    // which is lazy-loaded when the route is visited.
+    component: () => import(/* webpackChunkName: "modules" */ '../views/Swagger.vue')
+  },
```
jack-r-warren, line 54: This is just copied from similar routes here and it works so I didn't touch it much more

ui/src/views/Swagger.vue
```diff
@@ -0,0 +1,56 @@
...
+
+    axios.get("/oauth2id").then((response) => {
+      swagger.initOAuth({
+        ...(response.status == 200 && {
+          clientId: response.data["oauth2-client-id"],
+        }),
+        appName: "Workflow Launcher",
+        scopes: "openid email profile",
+      });
+    });
```
jack-r-warren, line 26: If the response from `/oauth2id` is 200 then we spread the returned client ID into the parameter object.

If the backend was down I can't imagine that authenticating would be much help, but this is at least controlled behavior that doesn't just crash the swagger page if only the frontend is online.

ui/vue.config.js
```diff
@@ -16,5 +18,18 @@ module.exports = {
...
-    ]
+    ],
+    configureWebpack: {
+        plugins: [
+            new CopyPlugin({
+                patterns: [
+                    {
+                        from: 'node_modules/swagger-ui/dist/oauth2-redirect.html',
+                        flatten: true,
+                        force: true
```
jack-r-warren, line 29: At build time, the `oauth2-redirect.html` static file that is included in the swagger-ui `dist` is copied into our own `dist` so it gets hosted like everything else.

I argue this is the best solution because:
- It is very much recommended to `import` other things from their `dist`, like the .css files to style the UI
- We can't just `import` this HTML like Gatsby users etc can because we don't want to mess with our router to treat this like a Vue template (in other words, we'd have to mess with Webpack and our router to be able to `import` HTML without changing it)
- CopyPlugin will error the build if this file ever disappears
- Having an automated build-time copy avoids licensing issues and means they can update that file whenever they like

rexwangcc: makes sense to me

rexwangcc (approved):
LGTM, all look good to me. Thanks for doing this. Having a usable swagger is a big step closer to make WFL more user friendly!

PR #183, opened 2020-10-14
[GH-1026] Make the `/oauth2id` endpoint lazier

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1026

### Changes

- Makes the `/oauth2id` endpoint lazier so that it won't try to come up with its response during `prebuild`

### Review Instructions

- You can run a local server to see that the endpoint still works
- Tests here should still pass
- There should be no error text in the output of the `prebuild` step of the github action here

### DIFF
--- a/api/src/wfl/api/routes.clj
+++ b/api/src/wfl/api/routes.clj
@@ -39,7 +39,7 @@
             :swagger {:tags ["Informational"]}}}]
    ["/oauth2id"
     {:get {:summary   "Get the OAuth2 Client ID for this deployment of the server"
-           :handler   (handlers/success {:oauth2-client-id @once/oauth-client-id})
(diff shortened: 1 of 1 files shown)


api/src/wfl/api/routes.clj
```diff
@@ -39,7 +39,7 @@
             :swagger {:tags ["Informational"]}}}]
    ["/oauth2id"
     {:get {:summary   "Get the OAuth2 Client ID for this deployment of the server"
-           :handler   (handlers/success {:oauth2-client-id @once/oauth-client-id})
+           :handler   (fn [_] (handlers/succeed {:oauth2-client-id @once/oauth-client-id}))
```
jack-r-warren, line 42: `succeed` is the same as `success`, not just using `constantly`, so it is properly lazy

rexwangcc (approved):

PR #189, opened 2020-10-20
[GH-1031] Fix unknown pipeline version

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1031

### Changes

- Add the WARP WDLs back into what ends up in the version file
- Add a fast-running unit test that WDL versions are in the version.edn

### Review Instructions

The PR will run the unit test showing that the version.edn file contains the versions of the WDLs, and since the code that produces "Unknown" is

```clojure
(key-for :wdl-version) (or (the-version wdl-value) "Unknown")
```

we should be all set. Can also verify via logging in to `./ops/server.sh`, you'll see the WARP WDL version info on the right-hand side again which is just read straight out of `version.edn`.

### DIFF
--- a/api/test/wfl/unit/all_modules_test.clj
+++ b/api/test/wfl/unit/all_modules_test.clj
@@ -1,7 +1,13 @@
 (ns wfl.unit.all-modules-test
-  (:require [clojure.test :refer [deftest testing is]]
+  (:require [clojure.string :as str]
+            [clojure.test :refer [deftest testing is]]
             [wfl.environments :as env]
-            [wfl.module.all :as all]))
+            [wfl.module.all :as all]
+            [wfl.module.aou :as aou]
+            [wfl.module.ukb :as ukb]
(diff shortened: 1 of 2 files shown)


api/test/wfl/unit/all_modules_test.clj
```diff
@@ -1,7 +1,13 @@
 (ns wfl.unit.all-modules-test
-  (:require [clojure.test :refer [deftest testing is]]
+  (:require [clojure.string :as str]
+            [clojure.test :refer [deftest testing is]]
             [wfl.environments :as env]
-            [wfl.module.all :as all]))
+            [wfl.module.all :as all]
+            [wfl.module.aou :as aou]
+            [wfl.module.ukb :as ukb]
```
ehigham, line 7: think you can skip ukb! That's `zero` only!

ehigham (approved):

PR #190, opened 2020-10-20
[GH-1032] append-to-aou accepts one or more sample

ehigham:
### Purpose

Fixes GH-1032:
In spec, accept one or more samples.
Add test points to `test-append-to-aou-workload` to lock down behaviour.


jack-r-warren (approved):
+1 to the additional tests!

samanehsan (approved):
Thanks for adding more test cases! :octocat:

rexwangcc (approved):

PR #157, opened 2020-09-28
[GH-994] Fix WGS input preparation

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-994
- See: https://github.com/broadinstitute/dsde-pipelines/pull/1084/files?file-filters%5B%5D=.wdl#r470019578
- Also see: https://broadinstitute.slack.com/archives/CESEYJW9W/p1601306420061600

### Changes

- Fix WGS input preparation, see [here](https://broadinstitute.slack.com/archives/CESEYJW9W/p1600973191017200?thread_ts=1600968869.015600&cid=CESEYJW9W)

### Review Instructions

- No instructions.

PR #191, opened 2020-10-20
[GH-771] WGS updates

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-771

### Changes

- Bump the WDL to 1.1.1 per Jessica Way
- Remove the fingerprinting since that is a per-sample thing (also per Jessica Way)
- Roll in Rex's v52 changes from GH-994 (#157)

### Review Instructions

- Is GH-994 still blocked? Are we not wanting this to be in because of where Cromwell v52 is and isn't?

### DIFF
--- a/api/src/wfl/module/wgs.clj
+++ b/api/src/wfl/module/wgs.clj
@@ -64,10 +64,7 @@
          (-> {:disable_sanity_check true}
              (util/prefix-keys :CheckContamination)
              (util/prefix-keys :UnmappedBamToAlignedBam)
-             (util/prefix-keys :UnmappedBamToAlignedBam)
              (util/prefix-keys :WholeGenomeGermlineSingleSample)
-             (util/prefix-keys :WholeGenomeGermlineSingleSample)
-             (util/prefix-keys :WholeGenomeReprocessing)
              (util/prefix-keys :WholeGenomeReprocessing))))
 
 (defn env-inputs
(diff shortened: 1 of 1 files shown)


rexwangcc (commented):
Looks good, I'm not sure if this will work for Cromwell v52+ since https://github.com/broadinstitute/wfl/pull/157/files are not in yet.

api/src/wfl/module/wgs.clj
```diff
@@ -106,7 +106,7 @@
                    (assoc input-key in-gs)
                    (assoc :destination_cloud_path (str out-gs out-dir))
                    (assoc :references (make-references ref-prefix))
-                   #_(merge fingerprinting) ;; Uncomment to enable fingerprinting
+                   (merge fingerprinting) ;; Uncomment to enable fingerprinting
```
ehigham, line 109: remove the comment too!

jack-r-warren: Oops, my bad. Guess this is staying commented anyway for right now? I'm still trying to sort out this ticket, I reverted the commit that uncommented this

jack-r-warren:
Just went and pulled Rex's Cromwell v52 changes from GH-994 (#157) into this PR so this should work properly

rexwangcc (approved):
LGTM. (I moved GH-994 to the current sprint, we can close it together with this one once the PR gets in)

PR #194, opened 2020-10-21
[GH-1034] Fix reference_fasta function

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1034

### Changes

- Make `hg38-genome-references` smarter about not passing nil to `reference_fasta`
  - Can't use fnil because the default is in the function
- Add a test for prefix handling in the references file

### Review Instructions

- See the ticket for more info on what happened here

### DIFF
--- a/api/test/wfl/unit/references_test.clj
+++ b/api/test/wfl/unit/references_test.clj
@@ -5,8 +5,10 @@
 
 (deftest prefix-handling
   (testing "reference_fasta function"
-    (testing "treats nil as no arg"
-      (is (= (references/reference_fasta) (references/reference_fasta nil))))
+    (testing "treats nil/empty as no arg"
+      (is (= (references/reference_fasta)
+             (references/reference_fasta nil)
+             (references/reference_fasta ""))))
     (testing "supplies a bucket prefix if no arg"
       (run! #(is (str/starts-with? % "gs://")) (vals (references/reference_fasta))))
     (testing "uses a given prefix"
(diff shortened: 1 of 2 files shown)


api/test/wfl/unit/references_test.clj
```diff
@@ -0,0 +1,22 @@
...
+      (run! #(is (str/starts-with? % "gs://"))
+            (vals (references/reference_fasta))))
+    (testing "uses a given prefix"
+      (let [prefix "foo"]
+        (run! #(is (str/starts-with? % prefix))
+              (vals (references/reference_fasta prefix))))))
+  (testing "hg38-genome-references function"
+    (testing "supplies a reference_fasta default if nil"
+      (run! #(is (str/starts-with? % "gs://"))
+            (vals (:reference_fasta (references/hg38-genome-references nil)))))
```
ehigham, line 18: Appreciate that you've taken the time to lick down behaviour in tests.
For my money, passing `nil' is never really the right thing to do as it doesn't really capture what the programmer rmeans. They don't mean a nil thing, they mean give me the default.

ehigham (approved):

api/test/wfl/unit/references_test.clj
jack-r-warren: Thanks, and I agree. I did it this way because XX accesses `(reference_fasta)` to get the default, while WGS uses `(make-references nil)` which calls `(hg38-genome-references nil)` which now calls `(reference_fasta)`. If I were to remove nil further up the chain on the WGS side, it would make it awkward to have arg and no-arg versions of all of those functions. I think the code is at least readable as is and the testing should form a good set of guide rails if other people venture here in the future.

PR #195, opened 2020-10-21
GH-986: Remove dsde-pipelines as WFL's dependency and fix what breaks.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-986

### Changes

- No changes.

### Review Instructions

- No instructions.


ehigham:
Looking good so far! In case you weren't aware, be sure to remove the dependence on dsde-pipelines in api/module.mk!

tbl3rd:
Thanks. Might have missed that.

api/src/wfl/boot.clj
```diff
@@ -111,7 +95,7 @@
               (io/make-parents out)
               (io/copy file out)))]
     (let [environments (clone "pipeline-config" "wfl/environments.clj")]
-      (stage resources (clone "dsde-pipelines" "tasks/CopyFilesFromCloudToCloud.wdl"))
+      (stage resources (clone "warp" "tasks/broad/CopyFilesFromCloudToCloud.wdl"))
```
rexwangcc, line 98: Ah, good catch! Mb

api/src/wfl/module/wgs.clj
```diff
@@ -41,31 +41,30 @@
 
 (def cram-ref
   "Ref Fasta for CRAM."
-  {:cram_ref_fasta        "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta"
-   :cram_ref_fasta_index  "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta.fai"})
+  (let [hg38           "gs://gcp-public-data--broad-references/hg38/v0/"
+        cram_ref_fasta (str hg38 "Homo_sapiens_assembly38.fasta")]
+    {:cram_ref_fasta        cram_ref_fasta
+     :cram_ref_fasta_index  (str cram_ref_fasta ".fai")}))
```
rexwangcc, line 47: 2 parts of me have been struggling on "explicit hard-coded consts" v.s. "dynamically generate data" in Clojure for a year, I preferred the former simply because it's more friendly for full text search, but I guess this (latter) is a de facto in clj?

rexwangcc (approved):
LGTM. I left some comments. There are a few lines in the Github Actions that are also referencing dsde-pipelines, [this file](https://github.com/broadinstitute/wfl/blob/master/.github/workflows/pr-test.yml#L53) and [this file](https://github.com/broadinstitute/wfl/blob/master/.github/workflows/release_latest_version.yml#L62) just fyi :octocat:

api/src/wfl/module/wgs.clj
tbl3rd: I'd prefer these were just literals too now, but the code isn't consistent.

Wanted to just factor the data, but discovered wgs.clj now has a synthetic
`reference_fasta_prefix` parameter that necessitates something like a
`(defn reference_fasta [prefix] …)` to support it.  Tackling that in this PR
is a design change (and definitely a bridge too far).

tbl3rd: That said, for things like `foo` and `foo.index`, I think we should do something to emphasize that the index is just the root with a suffix added.

api/src/wfl/module/xx.clj
```diff
@@ -27,43 +27,33 @@
...
+      references/exome-scatter-settings
+      {:calling_interval_list (str hg38 "exome_calling_regions.v1.interval_list")
+       :haplotype_database_file (str hg38 hsa ".haplotype_database.txt")})))
 
 (def ^:private workflow-defaults
-  (let [hg38 (partial str "gs://gcp-public-data--broad-references/hg38/v0/")]
+  (let [hg38          "gs://gcp-public-data--broad-references/hg38/v0/"
+        hsa           "Homo_sapiens_assembly38"
+        bait_set_name "whole_exome_illumina_coding_v1"
+        iv1           (str "HybSelOligos/" bait_set_name "/" "." hsa)]
```
ehigham, line 46: I really don't like this - I wrote it this way to be explicit. I can look at the design doc and automatically recognise the defaults without juggling temporaries. Please revert.

api/src/wfl/module/xx.clj
```diff
@@ -27,43 +27,33 @@
 (def workflow-wdl
   "The top-level WDL file and its version."
   {:release "ExternalExomeReprocessing_v2.1.1"
-   :top     "pipelines/broad/reprocessing/external/exome/ExternalExomeReprocessing.wdl"})
+   :top     (str "pipelines/broad/reprocessing/external/exome/"
```
ehigham, line 30: why?

api/src/wfl/module/xx.clj
```diff
@@ -27,43 +27,33 @@
...
-     :bait_interval_list   (hg38 "HybSelOligos/whole_exome_illumina_coding_v1/whole_exome_illumina_coding_v1.Homo_sapiens_assembly38.baits.interval_list")
-     :target_interval_list (hg38 "HybSelOligos/whole_exome_illumina_coding_v1/whole_exome_illumina_coding_v1.Homo_sapiens_assembly38.targets.interval_list")
+     :cram_ref_fasta       (str hg38 hsa ".fasta")
+     :cram_ref_fasta_index (str hg38 hsa ".fasta.fai")
+     :bait_set_name        bait_set_name
+     :bait_interval_list   (str hg38 iv1 ".baits.interval_list")
+     :target_interval_list (str hg38 iv1 ".targets.interval_list")
      :references           references-defaults
-     :scatter_settings     scatter_settings-defaults
-     :papi_settings        papi_settings-defaults}))
```
ehigham, line 66: These are required inputs to the wdl. They're not part of the "references" structure. Why have you removed this?

ehigham (changes requested):

api/src/wfl/module/xx.clj
```diff
@@ -27,43 +27,33 @@
...
+     :scatter_settings     references/exome-scatter-settings
+     :papi_settings        {:agg_preemptible_tries 3
```
ehigham, line 55: Does inlining the papi settings improve clarity?

tbl3rd: I think so, but let's discuss it, because even I don't agree with some of these changes.

api/src/wfl/module/xx.clj
tbl3rd: I don't know what you mean about juggling temporaries.
I agree with you about the scatter-settings.
They are back where they belong.

api/src/wfl/module/xx.clj
tbl3rd: I moved them back.  (Just trying to make sense of stuff.)

api/src/wfl/module/xx.clj
tbl3rd: FWIW: Wanted to emphasize that inputs depend on the chosen bait set.

ehigham (approved):

api/src/wfl/util.clj
```diff
@@ -401,8 +388,7 @@ vault.client.http/http-client                               ; Keep :clint eastwo
 (defn to-quoted-comma-separated-list
   "Return the sequence `xs` composed into a comma-separated list string.
   Example:
-    (to-quoted-comma-separated-list ['x 'y 'z])
```
ehigham, line 404: FYI the extra line is what you get when evaluating in a repl

tbl3rd: Didn't even realize I'd closed that.  More interested in the simpler quoting.

PR #196, opened 2020-10-22
some code tidy-ups.

ehigham:
Changes:

`spec.clj`
1. Update spec to name things consistently (use a pipeline prefix and common suffix). 
2. Remove stub aou specs
3. Add copyfile spec

`aou.clj`
1. Simplify the `append-to-workload` implementation. 
2. Remove the need to specify `cromwell` and `environment` - we already know these from the workload.
3. Add integration tests
4. Stop marking the workload as `:finished` when all the workflows are done. 
5. Remove the need to specify the empty `items` list

`workloads.clj`:
1. Throw if the workload cannot be loaded. This removes the need for duplicate error handling logic in `handlers.clj`. 
2. Add error handling logic to default multimethod implementations to add context to exceptions and rethrow.

`postgres.clj`:
1. Mark the workload as `:finished` when all workflows are in a finished state. This would previously happen the next time `update-workload!` was called. Fixes https://broadinstitute.atlassian.net/browse/GH-1040
2. Only update the cromwell status of the workflows that are not already finished. 

`server.sh`:
1. Stop opening the workflow-launcher ui. This is rarely the purpose of starting a local development server and it stealing focus is annoying.

`module.mk`:
1. prebuild dependends on `build.boot` - changes here will cause prebuild to me run again. 
2. Make unit and integration tests depend on the test code. If the code changes, the tests will be re-run. 

`xx.clj`:
1. Removed input bucket `item` - this complicated the spec implementation and was a bit grim anyway.


jack-r-warren:
I'll take a look on my work laptop in the morning where I have a clj repl. There's some wonders I was able to pull with `s/describe` a few days ago, maybe I'll get lucky again

jack-r-warren (commented):
Generally looks good to me so far. I'm assuming this'll be merged before GH-1013 (#185) so I'll have to rework some stuff there. (Specifically the spec stuff and re-organizing the integration tests will conflict with that PR)

api/src/wfl/api/spec.clj
```diff
@@ -55,39 +57,38 @@
...
+                      :wgs      ::wgs-workflow-inputs
+                      :xx       ::xx-workflow-inputs))
 
-(s/def ::inputs (s/or :aou ::aou-workflow-inputs
-                      :wgs ::wgs-workflow-inputs
-                      :xx  ::xx-workflow-inputs))
+(s/def ::workflows (s/* ::workflow))
+(s/def ::workflow
+  (s/keys :req-un [::inputs]
+          :opt-un [::status ::updated ::uuid]))
```
jack-r-warren, line 70: I like this simplification

api/test/wfl/integration/wgs_pipeline_test.clj
```diff
@@ -1,43 +0,0 @@
-(ns wfl.integration.wgs-pipeline-test
```
jack-r-warren, line 1: How could I have written this test file better so we'd judge it worth keeping?

jack-r-warren (approved):

api/test/wfl/integration/wgs_pipeline_test.clj
ehigham: This test file has been really useful for me! We can add more tests as we go!

ehigham: It's actually been moved to `test/integration/modules/wgs_test.clj`

jack-r-warren: Ah I gotcha, GitHub's diff didn't recognize the move and you simplified the test enough that I didn't recognize it 😅

PR #197, opened 2020-10-22
[GH-1035] Add terra module to WFL

samanehsan:
### Purpose

Started as a part of https://broadinstitute.atlassian.net/browse/GH-1018

### Changes
- Adds a terra module with functions to: 
  - Create a submission
  - Get a submission by uuid

A submission is like Terra's concept of a workload. It includes what data need to get processed with what pipeline and when you make a request to get a submission it includes a list of the workflows that were started in Cromwell.

### Review Instructions

- No instructions.

### DIFF
--- a/api/src/wfl/service/terra.clj
+++ b/api/src/wfl/service/terra.clj
@@ -3,7 +3,8 @@
   (:require [clojure.data.json :as json]
             [clojure.string :as str]
             [clj-http.client :as http]
-            [wfl.once :as once]))
+            [wfl.once :as once]
+            [wfl.util :as util]))
 
 (defn workspace-api-url
   [terra-url workspace-namespace workspace-name]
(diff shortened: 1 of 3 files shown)


ehigham (changes requested):
Can you add integration tests for these?

api/src/wfl/service/terra.clj
```diff
@@ -0,0 +1,53 @@
...
+         :headers   (once/get-auth-header)
+         :body      (json/write-str {
+                    :methodConfigurationNamespace method-configuration-namespace
+                    :methodConfigurationName method-configuration-name
+                    :entityType entity-type
+                    :entityName entity-name
+                    :useCallCache true
+                   } :escape-slash false)}
+      http/request :body
+      (json/read-str :key-fn keyword)
```
ehigham, line 29:
```
util/parse-json
```
?

api/src/wfl/service/terra.clj
```diff
@@ -0,0 +1,53 @@
...
+      :submissionId)))
+
+(defn get-submission
+  "Get information about a Terra Cromwell submission."
+  [terra-url workspace-namespace workspace-name submission-id]
+  (let [workspace-url (workspace-api-url terra-url workspace-namespace workspace-name)]
```
ehigham, line 35: Why not call `http/get` directly?

api/src/wfl/service/terra.clj
```diff
@@ -0,0 +1,53 @@
...
+    (-> {:method  :get
+         :url     (str workspace-url "/submissions/" submission-id)
+         :headers (once/get-auth-header)}
+      http/request :body
+      (json/read-str :key-fn keyword))))
+
+(comment
```
ehigham, line 42: Move this into an integration test?

api/src/wfl/service/terra.clj
samanehsan: Added it!

api/src/wfl/service/terra.clj
jack-r-warren: I think there's some slight benefit to making the code here the same as that for create-submission. If there's a http/post we can call there then sure I guess but this is pretty readable to me as-is.

samanehsan: No particular reason...I referenced another module when writing this one and it used `http/request` but I think `http/get` is a bit clearer!

samanehsan: Oh I didn't see Jack's comment...I updated these functions to use `http/get` and `http/post`. Let me know if you all think this is more readable!

rexwangcc (commented):
I think this is a good start especially after today's eng discussion. It seems the test is failing and it's crying about the workspace does not exist.

samanehsan:
> It seems the test is failing and it's crying about the workspace does not exist. 

I just added the wfl service account to the Terra workspace and the test passes now! I also added the rest of the Hornet team.

rexwangcc (approved):
Cool, LGTM!

api/test/wfl/integration/terra_test.clj
```diff
@@ -0,0 +1,34 @@
...
+
+(def entity-type
+  "sample")
+
+(def entity-name
+  "NA12878_PLUMBING")
+
+(deftest test-terra-submission
+  (testing "A workflow is created for the entity"
+    (let [submission-id (terra/create-submission terra-url workspace-namespace workspace-name
```
ehigham, line 29: Thanks for adding the test. Makes it easier to see how it's meant to be used.

What's the difference between the namespace and the name? Does it make sense to combine these into a fully-qualified name? (same for method-configuration)

samanehsan: The `workspace-namespace` corresponds to a gcloud billing project and can contain multiple workspaces. Here's one I set up in dev using the "general-dev-billing-account" namespace: https://bvdp-saturn-dev.appspot.com/#workspaces. I like the idea of combining the two parameters into one since they're just used in the submission url.

samanehsan: Also it seems like the `method-configuration-namespace` will always be the same as the `workspace-namespace`. But they can be specified individually through the Terra API so I'm assuming there's a case where they can be different.

ehigham (approved):

PR #198, opened 2020-10-27
[GH-1014] Workflow Options

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1014

### Changes

- Modify database to add workflow_options columns to workloads table and workflow types
- Modify spec to declare usage of workload_options field
- Modify database workload-request writing to merge default options with given ones and write the result as json
- Modify database workload reading to read the options as json
- Modify each module to pass along options present in the workload
  - In other words, rather than calculating options when we submit to cromwell, we now calculate options right when we receive the request and save it to the database. This solves the problem of how to communicate this back to the user, because our "workload" "type" just always includes the exact options that would be / were used. 
- Rather than XX doing a "batch call" once, it now does one per set of workflows with a set of options so that we can swap in the real batch call handling in `wfl.service.cromwell` in the future and this'll work for it
- Add a new page to the docs explaining how this works and how to use it. Tried to make it abstract over all modules so we don't clutter individual module documentation with stuff about the options.
- Add a new integration test for every module (including copyfile) demonstrating:
  - Existing defaults are still used and are communicated to the user in responses for both per-workload and per-workflow
  - Per-workload options are used, seen by a mocked Cromwell, and are communicated to the user in responses in both per-workload and per-workflow fields
  - Per-workflow options can be varied per workflow, seen by a mocked Cromwell, and are communicated to the user in responses in the per-workload field

### Review Instructions

- I think running the integration tests gives really good coverage because the test literally mocks Cromwell and records what options it sees and this is done for every module
- You can also play with it and see that workflow options including WFL's defaults are communicated to you in the responses. You can see them recorded in the DB too
- If you're curious about the precedence, it might be helpful to look at [this image](https://github.com/broadinstitute/wfl/blob/jack-r-warren/GH-1014-workflow-options/docs/md/assets/option-precedence.svg) that I made that is shown in the doc page this PR adds (I figure 40KB isn't that bad to commit to git 🥴)


database/changesets/20201027_WorkflowOptions.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext
+                        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
+                        http://www.liquibase.org/xml/ns/dbchangelog
+                        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+    <changeSet author="jwarren@broadinstitute.org"
```
jack-r-warren, line 10: I'd certainly like jwarren@broadinstitute.org to be a long-lived email but in the case that it's not, should I use the non-Broad primary email on my github account here?

tbl3rd: Should `workflow_options` should be on the `workload` table instead?
(It is usually shared across multiple workflows in Cromwell.)

tbl3rd: Or … after discussing it with Ed, maybe not.  Carry on.

api/src/wfl/api/spec.clj
```diff
@@ -32,9 +32,11 @@
 (s/def ::uuid-query (s/keys :opt-un [::uuid]))
 (s/def ::version string?)
 (s/def ::wdl string?)
+(s/def ::workflow_options map?)
 (s/def ::workload-request (s/keys :opt-un [::input
                                            ::items
-                                           ::common_inputs]
+                                           ::common_inputs
+                                           ::workflow_options]
```
ehigham, line 39: Here's an idea! (Not sure if it's any good)
You could combine these into a "common" map that could look like: 

```
(s/def ::common (s/keys :opt-un [::inputs ::options]))
```

(This would also just require one call to `deep-merge` later in the implementation, just that's an extra)

api/src/wfl/module/copyfile.clj
```diff
@@ -14,31 +15,37 @@
...
-  (let [[uuid table] (all/add-workload-table! tx workflow-wdl _workload-request)
-        inputs (map :inputs items)]
+  (let [default-options   (-> (:cromwell _workload-request)
+                              all/cromwell-environments
+                              first
+                              util/make-options)
+        [uuid table opts] (all/add-workload-table! tx workflow-wdl _workload-request default-options)
+        to-row (fn [item] (assoc (:inputs item)
+                            :workflow_options
+                            (json/write-str (util/deep-merge opts (:workflow_options item)))))]
```
ehigham, line 38: Alternative?
```
(-> item 
  (select-keys [:inputs :workflow_options])
  (update :workflow_options #(json/write-str (util/deep-merge opts %)))))
```

api/src/wfl/module/xx.clj
```diff
@@ -95,30 +95,35 @@
...
-          (merge cromwell-labels {:workload uuid}))))))
+  (let [path                 (wdl/hack-unpack-resources-hack (:top workflow-wdl))
+        environment          (get-cromwell-environment workload)
+        ;; Batch calls have uniform options, so we must group by options to submit
+        workflows-by-options (seq (group-by :workflow_options workflows))]
+    (letfn [(update-workflow [workflow cromwell-uuid]
+              (assoc workflow :uuid cromwell-uuid
+                              :status "Submitted"
+                              :updated (OffsetDateTime/now)))
+            (submit-workflows-by-options [[options associated-workflows]]
```
ehigham, line 106: what's an `associated-workflow`?

api/src/wfl/module/xx.clj
```diff
@@ -95,30 +95,35 @@
...
+              (mapv update-workflow
+                    associated-workflows
+                    (cromwell/submit-workflows
+                      environment
+                      (io/file (:dir path) (path ".wdl"))
+                      (io/file (:dir path) (path ".zip"))
+                      (map (comp (partial cromwellify-inputs environment) :inputs) associated-workflows)
+                      options
+                      (merge cromwell-labels {:workload uuid}))))]
+      (apply conj (mapv submit-workflows-by-options workflows-by-options)))))
```
ehigham, line 116: huh? Why are you building a sequence here? why don't you just `run!`

api/src/wfl/module/xx.clj
```diff
@@ -136,12 +141,16 @@
...
+          (unpack-options [m]
+            (update m :workflow_options #(util/parse-json (or % "{}"))))]
     (->> (postgres/get-table tx items)
-      (mapv (comp #(update % :inputs load-inputs!) unnilify))
-      (assoc workload :workflows)
-      unnilify)))
+         (mapv (comp #(update % :inputs load-inputs!) unnilify unpack-options))
+         (assoc workload :workflows)
+         unnilify
+         unpack-options)))
```
ehigham, line 153: Aren't the workflow options per workflow? Why are you unpacking them at the workload level too?

database/changesets/20201027_WorkflowOptions.xml
jack-r-warren: It is on both the workload table and the table of individual workflows. The table storing workloads stores the options guaranteed across that entire workload, while the table storing the individual workflow stores the options on a per-workflow basis

api/src/wfl/api/spec.clj
jack-r-warren: I like that idea--there's a spec thing called `s/merge` that lets you define multiple map specs and merge them together so we could do the sort of abstraction you're talking about. I didn't use it here because the "common" stuff is optional on requests and required on response, and we wouldn't have a great way to represent that.

api/src/wfl/module/copyfile.clj
jack-r-warren: Good idea! Put this in WGS too

api/src/wfl/module/xx.clj
jack-r-warren: A collection of workflows that all have the same options. This is an xx-specific things to allow us to easily implement the batch endpoint in the future. (Ed and I talked about this offline for a bit but for other folks tuning in, the xx module is "batch endpoint ready" and this behavior of grouping workflows by their options before making batch calls keeps it that way)

jack-r-warren: At risk of being verbose I changed the signature here to be `(submit-workflows-by-options [[options workflows-with-those-options]]` which I think might be a bit more clear

api/src/wfl/module/xx.clj
jack-r-warren: The caller of the function expects the workflows to be returned and for them to have uuids etc added. This function doesn't touch the database, its the caller that does

api/src/wfl/module/xx.clj
jack-r-warren: This is something in the spec doc. workflow_options are indeed per workflow, but something Trevyn pointed out is that often they'd want the same options for a huge number of workflows. Thus we provide the option at the workload-level too to accommodate that. Of course, anything provided on an individual workflow will override. 

This is also why we store workflow_options in the database at both the per-workflow and per-workload level, too

api/src/wfl/module/copyfile.clj
jack-r-warren: I ended up undoing this because I didn't realize that this can't really deal with the input nesting well. I like the `select-keys` approach but what I really want is just "put JSON-ified workflow_options into the map of inputs"

api/src/wfl/module/xx.clj
tbl3rd: Is this repeated (comp unnilify unpack-options) to handle the "two levels"?

api/src/wfl/module/copyfile.clj
ehigham: I don't understand.

api/src/wfl/module/xx.clj
ehigham: I really don't think you need the verbose names in this function - it's pretty clear from the usage. If anything, introducing new names can just cause confusion.

api/src/wfl/module/copyfile.clj
ehigham: nevermind - i misunderstood your code initially.

ehigham: why are workflow_options part of the inputs?

api/src/wfl/api/spec.clj
jack-r-warren: Are you suggesting another layer of nesting for common inputs and workflow options in the actual API?

api/src/wfl/module/copyfile.clj
jack-r-warren: This is how the WGS and copyfile databases are. I know you're reworking WGS, but basically what I did was just add a column of text to be the json encoded options. So since this bit of code is just turning the workflow into a row that we can write into the db, we want to put the workflow options right next to the inputs

api/src/wfl/module/xx.clj
jack-r-warren: Alright, I've given it another shot and renamed it to `ws` since `workloads` is already defined by the outer scope, I didn't want to shadow.

api/src/wfl/module/xx.clj
jack-r-warren: Yeah, we unpack both for every workflow and for the workload as a whole

api/src/wfl/api/workloads.clj
```diff
@@ -110,14 +111,17 @@
...
-            (split-inputs [m]
-              (let [keep [:id :finished :status :updated :uuid]]
-                (assoc (select-keys m keep)
-                  :inputs (unnilify (apply dissoc m keep)))))]
+            (split-inputs [workflow]
+              (let [keep [:id :finished :status :updated :uuid :workflow_options]]
+                (assoc (select-keys workflow keep)
+                  :inputs (unnilify (apply dissoc workflow keep)))))
+            (unpack-options [m]
+              (update m :workflow_options #(util/parse-json (or % "{}"))))]
```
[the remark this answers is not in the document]
jack-r-warren: Okay. I'll update the spec and docs to reflect that the key won't always be included--I had done this as a readability thing.

api/src/wfl/module/all.clj
```diff
@@ -83,28 +84,32 @@
...
-                                    :project  project
-                                    :release  release
-                                    :uuid     (UUID/randomUUID)
-                                    :version  version
-                                    :wdl      top})
-        table (format "%s_%09d" pipeline id)
-        work  (format "CREATE TABLE %s OF %s (PRIMARY KEY (id))"
+  "Return UUID, TABLE, and WORKFLOW-OPTIONS for _WORKFLOW-WDL in BODY under transaction TX."
+  ([tx workflow-wdl body]
+   (add-workload-table! tx workflow-wdl body {}))
```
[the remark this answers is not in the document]
jack-r-warren: Good call, done

docs/md/workflow-options.md
```diff
@@ -0,0 +1,158 @@
...
+- "my_option" will have different strings for the different samples
+- "another_option" will be an object just set for the latter sample
+- "yet_another_option" will be the same string for all samples
+
+In other words, WFL will recursively merge the options objects together to
+resolve the options for individual workflows. You can see this in WFL's
+response, which includes all workflow options:
+```json
+{
+  // Some fields omitted for brevity!
```
jack-r-warren, line 97: What I would give for universal support of `.jsonc`. I feel like this is a helpful note but if everyone's parsers are going to throw a fit we can remove it

api/test/wfl/system/v1_endpoint_test.clj
```diff
@@ -45,6 +45,7 @@
                 (is uuid "workloads should be been assigned a uuid")
                 (is (util/absent? uuids uuid))
                 (is (:created workload) "should have a created timestamp")
+                (is (:workflow_options workload) "should have non-null options")
```
jack-r-warren, line 48: Options can technically be null, but for the modules we test here that will never be the case because there are always defaults that will be stored in this key

api/src/wfl/api/spec.clj
ehigham: Yeah sure - they're both "common" things. Maybe rename "workflow_options" to just "options".

database/changesets/20201031_WorkflowOptions.xml
```diff
@@ -0,0 +1,40 @@
...
+        </sql>
+        <sql>
+            ALTER TYPE CromwellWorkflow
+                ADD ATTRIBUTE workflow_options   text  CASCADE   -- workflow options to be passed to Cromwell
+        </sql>
+        <comment>
+            Add workflow options column to overall workload table
+        </comment>
+        <sql>
+            ALTER TABLE workload ADD workflow_options text
```
ehigham, line 37: Sorry I didn't spot this before.
I don't think adding workflow options to the workload table is the best. I'd just merge them with the per-workflow inputs and store them in the workflow table.
I realise that this means we'll be storing duplicate information, but I don't think this information should live in the workload table. 

Maybe in the future, we'd have 3 tables, one for the workload, one for the common features of the workflows, and one for the workflows itself. Who knows.

jack-r-warren: Ah, this was why I wanted to make sure my DB stuff was on the right path before I did everything else.

Why should we not store the workflow options at a per-workload level? If we don't, our API becomes lossy in that we forget about the nature of some options described by the user. As the documentation page describes, the workload-level options returned is also key for a user differentiating between WFL's per-workload defaults and per-workflow defaults.

api/src/wfl/api/spec.clj
jack-r-warren: I like keeping workflow_options because it helps differentiate itself from inputs and it the name is the same that Cromwell uses to help reference that authoritative documentation. I don't think we should invent a new name and add a layer of indirection, I think keeping it reinforces that WFL is transparent about them and just passes them along.

I'm indifferent about nesting this stuff inside a "common" key, I can change it if you really think it is important. I wish we'd fleshed that out during design rather than back-tracking now.

database/changesets/20201031_WorkflowOptions.xml
jack-r-warren: I could also add a new table just to associate options to workloads, if we didn't want to just add a column to the table. I don't know what that gets us with regard to database design but I don't see a reason why we shouldn't store the per-workload options. I suppose it might've lead to a simpler implementation, but the work here is done and tested so I don't know if that's moot.

tbl3rd: Why do we care "what is common"?  (Maybe we should talk about this.)

docs/md/workflow-options.md
rexwangcc: I like the comments, but it does not persuade me to switch to jsonc which means additional conversions before I can use any JSON. What we could do here in docs is moving this comment out of the JSON block and wrap it in a special block?

docs/md/assets/option-precedence.svg
```diff
@@ -0,0 +1 @@
...ink:href="#Q"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,4.876543209876543,0)" xlink:href="#N"/></g><g id="F"><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,0,0)" xlink:href="#T"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,7.345679012345679,0)" xlink:href="#W"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,19.691358024691358,0)" xlink:href="#K"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,30.80246913580247,0)" xlink:href="#O"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,43.148148148148145,0)" xlink:href="#L"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,55.49382716049382,0)" xlink:href="#N"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,67.8395061728395,0)" xlink:href="#K"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,78.95061728395062,0)" xlink:href="#W"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,91.2962962962963,0)" xlink:href="#J"/><use transform="matrix(0.06172839506172839,0,0,0.06172839506172839,95.49382716049384,0)" xlink:href="#K"/></g></defs></g></svg>
```
rexwangcc, line 1: Great chart!

jack-r-warren: Thanks! Took some inspiration from your module structure diagram

database/changesets/20201031_WorkflowOptions.xml
jack-r-warren: Users can provide options to be common, and WFL treats some options as being common across a workload (util/make-options, for the most part). Yes, those things end up being visible on the workflows, but they could've been overridden there. Keeping track of what is common allows us to give clear output to the user and also allows this change to work with modules that use an append system, like AoU.
(Here I'm using "common" == "per-workload" == "baseline for every workflow in the workload")

tbl3rd: OK. But make-options is just implementation.
Letting a user specify options to apply to all workflows is useful.
But that doesn't necessitate reflecting it in the database.
WFL could merge options into workflows at workload creation.

api/src/wfl/module/aou.clj
```diff
@@ -103,16 +104,18 @@
         (get-per-sample-inputs per-sample-inputs))
     (util/prefix-keys :Arrays)))
 
-(defn make-options
-  "Return options for aou arrays pipeline."
-  [sample-output-path]
```
tbl3rd, line 108: Yay.

api/src/wfl/module/aou.clj
```diff
@@ -103,16 +104,18 @@
...
   {; TODO: add :default_runtime_attributes {:maxRetries 3} here
-   :final_workflow_outputs_dir sample-output-path
    :use_relative_output_paths  true
    :read_from_cache            true
    :write_to_cache             true
    :default_runtime_attributes {:zones "us-central1-a us-central1-b us-central1-c us-central1-f"}})
 
+(defn ^:private per-workflow-default-options
+  [sample-output-path]
+  {:final_workflow_outputs_dir sample-output-path})
```
tbl3rd, line 117: Inline this below.

api/src/wfl/module/batch.clj
```diff
@@ -1,28 +1,37 @@
...
+                    :workflow_options (json/write-str workflow-options))
+             (->> (jdbc/insert! tx :workload)))
+         table (format "%s_%09d" pipeline id)]
+     (jdbc/execute! tx
+                    ["UPDATE workload SET pipeline = ?::pipeline WHERE id = ?" pipeline id])
+     (jdbc/db-do-commands tx
+                          (map #(format "CREATE TABLE %s OF CromwellWorkflow (PRIMARY KEY (id))" %)
+                               [table]))
+     (jdbc/update! tx :workload {:items table} ["id = ?" id])
+     [id table workflow-options])))
```
tbl3rd, line 37: workflow-options is here because we add it to the workload table?

tbl3rd (commented):
Simplify by pushing workflow_options to the workflow tables?
Factor the various test-workflow-options and re-incorporate baseline-options-across-workload parameterized somehow?

api/src/wfl/module/aou.clj
jack-r-warren: Done

api/src/wfl/module/batch.clj
jack-r-warren: Correct. Removing it now since it was mentioned that the caller might as well be responsible for this.

PR #203, opened 2020-10-29
[GH-1022] Repurpose external exome table for batch workloads

ehigham:
### Purpose
The table I wrote for external exome reprocessing is actually generic across the set of batch workloads in cromwell.
I intend to re-use this when extending the WGS module.

### Changes
Note that the database changes are breaking. Since the exomes module is not released yet, however, I don't think this is an issue.
I could have unified the all/create-workload-table and the batch version but decided against otherwise it would have caused a lot of merge conflicts for jack. I'll tidy this up later.

### DIFF
--- a/api/src/wfl/module/batch.clj
+++ b/api/src/wfl/module/batch.clj
@@ -7,7 +7,7 @@
 
 (defn add-workload-table!
   "Create row in workload table for `workload-request` using transaction `tx`.
-  Instantiate a BatchWorkload table for the workload.
+  Instantiate a CromwellWorkflow table for the workload.
   Returns: [id table-name]"
   [tx
    {:keys [release top] :as _workflow-wdl}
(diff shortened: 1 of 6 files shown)


jack-r-warren (approved):
Any WFL deployments that already used the deleted changelog will need to have their database destroyed before updating, at least I think. I'm sure liquibase will let folks know.

ehigham:
> Any WFL deployments that already used the deleted changelog will need to have their database destroyed before updating, at least I think. I'm sure liquibase will let folks know.

There are none. Exomes is not released, hence I re-purposed rather than migrated.

jack-r-warren:
> > Any WFL deployments that already used the deleted changelog will need to have their database destroyed before updating, at least I think. I'm sure liquibase will let folks know.
> 
> There are none. Exomes is not released, hence I re-purposed rather than migrated.

Yeah, I think my sandbox instance might need nuking but that's not a big deal

PR #204, opened 2020-10-29
[GH-1022] make wgs use `CromwellWorkflow` table.

ehigham:
### Purpose
In this change, I'm making the wgs workload store its workflow data in the new (repurposed) `CromwellWorkflow` table .
I've implemented the loader such that it's backwards compatible with our current deployments- we can load old WGS workflows as well as new ones (added tests).

> The only issue is that I need to bump the version without creating a release in order for the tests to pass in github.

I'm pretty sure that wgs now accepts `input_bam`s as well as `input_cram`s now. I'll add a test to verify this.


rexwangcc (approved):
LGTM.

PR #207, opened 2020-11-02
GH-1014 without a column in the workload table

jack-r-warren:
Like #198 but does not have a column in the workload table for storing workflow options. This has the effect of disabling AoU's `workload_options` functionality.

See #208 for the additional modifications that would eliminate the `workflow_options` column from the AoU table too (it is here to match with the other workflow tables to hook into the same system to send `workflow_options` back to the user).

This PR is code/test complete, though changes would be required to documentation files.


api/test/wfl/integration/modules/wgs_test.clj
```diff
@@ -36,21 +37,32 @@
...
-        uuid))))
+(defn ^:private old-create-wgs-workload!
+  ([] (old-create-wgs-workload! nil))
+  ([workflow-options]
+   (let [request (make-wgs-workload-request)]
+     (jdbc/with-db-transaction [tx (fixtures/testing-db-config)]
+       (let [[id table] (all/add-workload-table! tx wgs/workflow-wdl request)
+             to-row (fn [m id] (assoc (:inputs m)
+                                 :id id
+                                 :workflow_options (when workflow-options (json/write-str workflow-options))))]
```
ehigham, line 48: The old table didn't store workflow options (and I don't think it should be updated). Can you just add these to the loader?

database/changesets/20201031_WorkflowOptions.xml
```diff
@@ -0,0 +1,34 @@
...
+        <sql>
+            ALTER TYPE AllOfUsArrays
+                ADD ATTRIBUTE workflow_options   text  CASCADE   -- workflow options to be passed to Cromwell
+        </sql>
+        <sql>
+            ALTER TYPE copyfile
+                ADD ATTRIBUTE workflow_options   text  CASCADE   -- workflow options to be passed to Cromwell
+        </sql>
+        <sql>
+            ALTER TYPE ExternalWholeGenomeReprocessing
```
ehigham, line 26: As above - this type is deprecated and nothing will be stored using it anymore. We need to make sure we can still load it, however, and the only place we can handle this in the loader.
If anything, not updating this table will make the load previous wgs test more accurate.

api/test/wfl/integration/modules/wgs_test.clj
jack-r-warren: Fair enough. The default loader used for the old database already supports this out of the box (and doesn't complain if it is `nil`) so I've just removed the testing capability for it

database/changesets/20201031_WorkflowOptions.xml
jack-r-warren: Done--this is removed from liquibase

PR #209, opened 2020-11-02
GH-1014 but with workload-level options in their own table

jack-r-warren:
Another modification to #198.

This one basically:

- Adds a "common" table
- Puts workload-level options under a "common" key in the request spec (so `{:common {:workflow_options ...}}`)
- Adds another line to the various "add workload table" functions to also record anything common in the common table
- Adds another line to the lowest-level "load workload" function to basically take everything from the common table and `util/deep-merge` it into every workflow that was just loaded
- Removes some logic from the various "create workload" functions because they only have to worry about storing any workflow options that are actually unique in the database row (anything common will be merged in whenever the workload are loaded)
- Every other function behaves as normal

Functionally, this is all there, all tests pass. Implementation and naming might not be perfect, and docs would need updating

Fun fact: this stores less in the database overall than any other solution I've come up with because in the vast majority of cases, workflow's individual workflow_options would be `null` in the database because it would not need to repeat what was common. That does mean more computation, due to a lot of `deep-merge`ing, but whatever.


jack-r-warren:
I'll note here too that I think this would work for inputs too? As a scratch thing I absolutely did not concern myself with inputs or whatever Ed has done with common inputs at all, I just wanted to test the water. But in theory, if you added an "inputs" column to the table, anything you specified in .common.inputs would become an automatic baseline for all workflows whenever they were loaded.

database/changesets/20201101_CommonTable.xml
```diff
@@ -0,0 +1,27 @@
...
+        </comment>
+        <createTable tableName="common_within_workload">
+            <column name="workload_id" type="bigint">
+                <constraints nullable="false" primaryKey="true" unique="true"
+                             foreignKeyName="FK_workload_id" references="workload(id)"/>
+            </column>
+            <column name="workflow_options" type="text">
+                <constraints nullable="true"/>
+            </column>
+        </createTable>
```
ehigham, line 25: Wondering if this can be done in raw SQL like type definitions - I find XML a bit too cluttered...

database/changesets/20201101_CommonTable.xml
```diff
@@ -0,0 +1,27 @@
...
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext
+                        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
+                        http://www.liquibase.org/xml/ns/dbchangelog
+                        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+    <changeSet author="jwarren@broadinstitute.org"
+               dbms="postgresql"
+               id="20201101"
+               logicalFilePath="20201101_CommonTable.xml">
+        <comment>
+            A table containing content common within workflows in a workload.
```
ehigham, line 15: Am I correct in that you propose to only have one of these tables and all workloads *could* write into this table?
It's a nice idea, but I wonder if this will lead to greater database contention - hence why we're using separate tables for each workload.

Maybe you can discuss this in the meeting later.

database/changesets/20201101_CommonTable.xml
jack-r-warren: I'm sure it could. I don't know SQL well and was admittedly having trouble getting what I wanted working so IntelliJ's XML autocomplete was helping me out. If we went with this idea I could figure out how to get that working.

database/changesets/20201101_CommonTable.xml
jack-r-warren: This table is a is a sorta like 1-to-0..1 mapping from the workload table to this one. Each row in the workload table can have either 0 or 1 rows in this one. The "interface" I've written over this table in the `common.clj` file (poorly named, couldn't think of something better) treats these rows as immutable

jack-r-warren:
General consensus from talking with Rex, Saman, and Tom about this is to try to give this solution a shot. It improves over #198 by being kinder to the database (with more principles to boot), and the complexity is maybe in a bit better of a spot:
- rather than complicating the "add workload table" process itself, we just append an extra step to the list of things that happens after the table has been added
- rather than relying on the process of loading workflows to unpack and process options, we append an extra step to the overall workload process (this extra step is perfectly constant across all modules and variations because it operates specifically on a "fully loaded" workload)

I don't wanna put words in people's mouths but it seemed to be like the slightly better implementation and database stuff here tips the scales against just removing it from the DB and AoU module altogether (the implementation isn't so bad that it outweighs the loss of functionality).

Some tasks before I close the other PRs and propose this for actual merging:
- [ ] Rex mentioned making deletions in the workload table cascade to this one, I want to make sure I can figure out how to do that
- [ ] I'm going to check with @ehigham tomorrow to make sure it makes sense to leave `common_inputs` out of this--we could in theory very easily enable it here but that seems like something for a future PR

I'll probably put it up for review while I finish writing documentation. It is all there, just need to maybe update it a bit since the common stuff is in a different place.

A preview of review requests in that PR
- Names: could they be better? Any ideas? Table names, function names, etc
- SQL: Idk what I'm doing please check my work
- Once my documentation is done, does it make sense? I'll probably just show it to Trevyn or Nareh or Steve or someone

database/changesets/20201101_CommonTable.xml
```diff
@@ -0,0 +1,32 @@
...
+                workflow_options text NULL,
+
+                CONSTRAINT fk_workload_id
+                    FOREIGN KEY (workload_id)
+                    REFERENCES workload (id)
+                    ON DELETE CASCADE
+                    ON UPDATE CASCADE,
+                CONSTRAINT workload_id_unique
+                    UNIQUE (workload_id)
+            );
```
jack-r-warren, line 29: Tests pass but does this look right @rexwangcc?

rexwangcc: It looks right to me. Let me test on your branch. I figured delete cascade is good since we are making a 1 to 1 mapping between records in `workload` table and `common` table, so it makes sense to delete the corresponding record from common when its workload is removed.

jack-r-warren: Definitely, I literally didn't know that was a thing when I first wrote the liquibase stuff here

rexwangcc: OK, based on my local testing, both `ON UPDATE CASCADE` and `ON DELETE CASCADE` work.

jack-r-warren: Thank you very very much for your lending your SQL prowess to making sure I'm not off my rocker--much more confident about this code now

database/changesets/20201101_CommonTable.xml
ehigham: So exactly as I had said above. Did you decide in the meeting that this wasn't an issue?

PR #210, opened 2020-11-03
[GH-978] Reap obsolete analysis versions from AOU input bucket

samanehsan:
### Purpose
Addresses https://broadinstitute.atlassian.net/browse/GH-978

Requirements from talking to hOps:
- Inputs for any old analysis versions can be safely cleaned up.
- Inputs for the latest analysis versions should be left alone (hOps can decide if/when they need to remove those to save on additional storage costs
- There should be a way for hOps to specify particular inputs to keep for debugging [TODO]

### Changes

Add a script `aou_cleanup.py` that will move old analysis inputs into a separate "cleanup" bucket with a lifecycle policy. If an Arrays workflow is being run in Cromwell that is using any of those input files, they will not get moved. Note there may be a potential race-condition where the workflow hasn't appeared in Cromwell yet so the script thinks it is safe to clean-up. Since the "reaped" files won't get deleted immediately this seems low risk.

### Review Instructions

I wasn't sure about the best place to put this script...maybe it could go into an hOps repo instead?


jack-r-warren:
I love this script! I don't know if it should go here though because it is unrelated to WFL. I don't think adding python code to PTC is the right call either, I like your idea of potentially making an hOps repo on the assumption of us having other scripts for them in the future (ahem, deploying auth proxy or jmui)

ehigham (approved):
Looks brilliant!

rexwangcc (commented):
LGTM!

rexwangcc (approved):

jack-r-warren:
> AoU Arrays project
@ehigham what do you mean by project?

samanehsan:
Thanks! I think another option is if hOps already has a repo where they keep various scripts we could put it there.

PR #211, opened 2020-11-04
[GH-978] Give hOps a way to exclude certain AOU files from cleanup

samanehsan:
### Purpose

Ticket: https://broadinstitute.atlassian.net/browse/GH-978

### Changes

Give hOps a way to specify files that they want to keep for debugging that would otherwise get cleaned up. This file is version controlled instead of being passed in as a parameter because different people from hOps will likely need to run this script.

### Review Instructions

- No instructions.


tbl3rd (approved):
Maybe "keep_files" instead of "excluded_files"?
Excluded from what?

samanehsan:
The data repo integration test is failing but since that is unrelated to this change I'm going to go ahead and merge this :octocat:

PR #215, opened 2020-11-05
[GH-1037] Parallel system test runner

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1037

### Changes

- Make the system test no longer test that returned UUIDs are unique (the test seemed to have low utility and it prevented parallelization)
- Split system tests up into individual `deftest`s
- Mark the long-running system tests with `^:parallel`
- Add a new Clojure alias for a simple non-Kaocha runner that just `future`s all tests marked with `^:parallel` in the namespace
- Add a note to the docs about the parallelizing runner
- An important thing _not_ changed is how the system tests are defined (thanks to Tom for the idea). They are still `deftest`s that can be run normally through Kaocha: `clojure -A:test system` has the exact same behavior as before, and `--focus` can be used that way for debugging.

### Review Instructions

- You can call something like `clojure -A:parallel-test wfl.system.v1-endpoint-test` to run the system tests in parallel (you can look at JMUI to see everything started at once, or at your local server's logs)
- I wonder if there's additional ways I can clean up the Clojure


api/test/wfl/tools/parallel_runner.clj
```diff
@@ -0,0 +1,22 @@
...
+(defn -main
+  [& args]
+  (letfn [(go! [& tests]
+            (let [output (new java.io.StringWriter)]
+              (binding [test/*test-out* output
+                        test/*report-counters* (ref test/*initial-report-counters*)]
+                (doseq [t tests] (t))
+                [@test/*report-counters* (str output)])))]
+    (let [ns      (symbol (first args))
+          _       (use ns)
```
jack-r-warren, line 13: This is hideous but I need something to call `ns-interns` on later. Even `find-ns` doesn't work if it isn't required or something, so this seems like a fix? I don't know. I just want to point at a namespace by its name and say "please let me call `vals` on you"

api/test/wfl/tools/parallel_runner.clj
```diff
@@ -0,0 +1,22 @@
...
+          tests   (group-by (comp #(contains? % :parallel) meta)
+                            (filter (comp :test meta) (vals (ns-interns (the-ns ns)))))
+          futures (doall (map #(future (go! %)) (tests true)))
+          results (apply list (apply go! (tests false)) (map deref futures))
+          summary (assoc (apply merge-with + (map first results))
+                    :type :summary)]
```
jack-r-warren, line 19: The `:type :summary` thing is what lets the `clojure.test/do-report` function print out a nice summary. I could also roll that myself though

api/test/wfl/tools/parallel_runner.clj
```diff
@@ -0,0 +1,22 @@
...
+      (run! println (filter not-empty (map second results)))
+      (test/do-report summary)
+      (System/exit (if (test/successful? summary) 0 1)))))
```
jack-r-warren, line 22: I couldn't find any easy docs on how to properly return an exit code and literally just returning 0 or 1 didn't actually seem to work for some reason

jack-r-warren (commented):
Notes to myself on stuff I wanna come back to, this is just a sketch and I wanna see how I'd do it with make

api/test/wfl/tools/parallel_runner.clj
jack-r-warren: Or maybe I should just document/comment this?

api/test/wfl/tools/parallel_runner.clj
```diff
@@ -0,0 +1,25 @@
+(ns wfl.tools.parallel-runner
+  (:require [clojure.test :as test])
+  (:import (java.io OutputStreamWriter StringWriter OutputStream)))
+
+(defn -main
+  [& args]
+  (letfn [(go! [& tests]
+            (let [null   (new OutputStreamWriter (OutputStream/nullOutputStream)) ;; Java 11+
```
jack-r-warren, line 8: I don't think it is worth it to try to roll our own from scratch to back-port this and it is only available in Java 11 and higher. I think we try to keep compatibility with 8 so maybe I just put another stringwriter here?

api/module.mk
```diff
@@ -54,6 +57,13 @@ $(INTEGRATION): $(TEST_SCM_SRC) $(TEST_SCM_RESOURCES) $(CLOJURE_PROJECT)
 	$(TEE) $(DERIVED_MODULE_DIR)/integration.log
 	@$(TOUCH) $@
 
+.PHONY: SYSTEM_TESTS
+$(SYSTEM): $(SYSTEM_TESTS)
+	@$(TOUCH) $@
+$(SYSTEM_TESTS): $(TEST_SCM_SRC) $(TEST_SCM_RESOURCES) $(CLOJURE_PROJECT)
+	$(EXPORT) CPCACHE=$(CPCACHE_DIR);            \
+    $(CLOJURE) $(CLJFLAGS) -A:test system --focus wfl.system.v1-endpoint-test/$@
+
```
jack-r-warren, line 66: This works too, this just sorta forces us to keep a list of exact system test names outside of our code. Also it feels funky to me to have to separately stand up a server and liquibase and stuff to run this. I bet this also breaks the `all` thing, because it can't run straight through now. I think I could start a server from make but i dunno how I would stop it

api/module.mk
```diff
@@ -28,6 +28,9 @@ TEST_SCM_SRC       = $(shell $(FIND) $(TEST_DIR) -type f -name "*.$(CLJ)")
 JAR          := $(DERIVED_TARGET_DIR)/wfl-$(WFL_VERSION).jar
 JAR_LINK     := $(DERIVED_TARGET_DIR)/wfl.jar
 
+SYSTEM_TESTS := \
+	test-append-to-aou-workload
```
ehigham, line 32: I don't think the makefile should have these details and i'm not sure implementing the parallelism in the makefile is right either

api/module.mk
```diff
@@ -54,6 +57,13 @@ $(INTEGRATION): $(TEST_SCM_SRC) $(TEST_SCM_RESOURCES) $(CLOJURE_PROJECT)
 	$(TEE) $(DERIVED_MODULE_DIR)/integration.log
 	@$(TOUCH) $@
 
+.PHONY: SYSTEM_TESTS
```
ehigham, line 60: When are you using this target?

api/module.mk
jack-r-warren: Fair enough--I think the clojure route might be a bit more extensible

api/test/wfl/tools/parallel_runner.clj
```diff
@@ -0,0 +1,33 @@
...
+                  output (new StringWriter)]
+              (binding [*out*                  null
+                        test/*test-out*        output
+                        test/*report-counters* (ref test/*initial-report-counters*)
+                        log/*logger-factory*   disabled-logger-factory]
+                (doseq [t tests] (t))
+                [@test/*report-counters* (str output)])))]
+    (let [ns      (map symbol args)
+          _       (run! use ns)
+          tests   (filter (comp :test meta) (apply concat (map (comp vals ns-interns the-ns) ns)))
```
ehigham, line 25:
```
(apply concat (map f xs)) == (mapcat f xs)
```

I think?

api/test/wfl/tools/parallel_runner.clj
ehigham: Is this creating one thread for each test marked `^:parallel`? Using an executor here might be advantageous, but maybe for another change.

ehigham (approved):
Looks good. Out of interest, how much wall clock time have you saved?

api/test/wfl/tools/parallel_runner.clj
jack-r-warren: Yeah, I figure that's another change. Figured I'd leave it to the JVM or host to manage the threads since they just spend the whole time sleeping anyway

api/test/wfl/tools/parallel_runner.clj
jack-r-warren: Yeah Tom pointed this out--was a good catch, I literally didn't know that existed

jack-r-warren:
> Looks good. Out of interest, how much wall clock time have you saved?

With WGS not being parallelized I roughly estimate the tests to take 20 minutes--I've had the tests creep in close to an hour sometimes so about 40 minutes or so in terms of savings. If we can get rid of the skip thing or get around that somehow then it would go down to just the longest workflow of the bunch

jack-r-warren:
Ed said to go ahead and merge this, tests only currently failing due to the Vault approle stuff

PR #216, opened 2020-11-09
[GH-1049] Better handling of cram bam in xx and wgs

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1049

### Changes

- Enforce `input_cram` to only accept `*.cram` and `input_bam` to only accept `*.bam` through specs as GH-1022 suggested.
- Add a test (unhappy path) for the spec.
- De-slashify Cromwell URLs in `aou` and `xx` modules in case someone like myself passes in `https://cromwell-gotc-auth.gotc-dev.broadinstitute.org/` and has no clue why it failed.

### Review Instructions

- No instructions.


api/src/wfl/api/spec.clj
```diff
@@ -15,8 +16,8 @@
 (s/def ::environment string?)
 (s/def ::finished inst?)
 (s/def ::input string?)
-(s/def ::input_bam string?)
-(s/def ::input_cram string?)
+(s/def ::input_bam (s/and string? #(str/ends-with? % ".bam")))
```
ehigham, line 19: think this just needs `str/ends-with?` (i.e. I don't think the `and` is necessary)

api/src/wfl/module/aou.clj
```diff
@@ -134,7 +134,8 @@
       (make-labels per-sample-inputs other-labels))))
 
 (defn ^:private get-cromwell-environment! [{:keys [cromwell]}]
-  (let [envs (all/cromwell-environments #{:aou-dev :aou-prod} cromwell)]
+  (let [cromwell (all/de-slashify cromwell)
```
ehigham, line 137: do you need to generate a temporary here?

api/test/wfl/unit/spec_test.clj
```diff
@@ -11,4 +12,12 @@
       (run! valid? (map #(% (UUID/randomUUID))
                         [workloads/wgs-workload-request
                          workloads/aou-workload-request
-                         workloads/xx-workload-request])))))
+                         workloads/xx-workload-request]))))
+  (testing "Mismatched cram/bam inputs cannot pass spec validation"
```
ehigham, line 16: Maybe make this a different test point?

rexwangcc: Good idea! Will update.

api/src/wfl/api/spec.clj
rexwangcc: will update!

api/src/wfl/module/aou.clj
rexwangcc: The extra var here is only to make the error reporting accurate without inlining `(all/de-slashify cromwell)` twice. (for example, without it `https://unregistered-cromwell.broad.org/` throws `no unique environment matching Cromwell URL. {:cromwell https://unregistered-cromwell.broad.org/}`, the temporary makes it throw `no unique environment matching Cromwell URL. {:cromwell https://unregistered-cromwell.broad.org}` as the de-slashified url is indeed what WFL used for the filtering.)

I'm ok with either way, do you think moving `(all/de-slashify cromwell)` inline looks better? I can update it.

jack-r-warren: The error reporting is something I hadn't thought about at all here. I don't really have any thoughts on this particular case partly because I think this class of error is something we'd want to solve at-scale when we start thinking about Terra integration and FISMA compliance. In other words, I think a lot of models for getting WFL FISMA-compliant are in conflict with the concept of an environments file, and I think solutions to that problem could be designed to eliminate this sorta class of errors and error messages.

ehigham: Ah yes errors! didn't see that - thanks.

api/test/wfl/unit/spec_test.clj
```diff
@@ -12,3 +13,13 @@
...
                          workloads/aou-workload-request
                          workloads/xx-workload-request])))))
+
+(deftest request-spec-negative-test
+  (testing "Mismatched cram/bam inputs cannot pass spec validation"
+    (letfn [(invalid? [req] (is (not (s/valid? ::spec/workload-request req))))]
+      (let [requests (map #(% (UUID/randomUUID))
+                          [workloads/wgs-workload-request
+                           workloads/xx-workload-request])
+            mismatched-requests (map #(walk/postwalk-replace {:input_cram :input_bam} %)
```
ehigham, line 23: `postwalk-replace`?? Looks shiny!

ehigham (approved):

api/test/wfl/unit/spec_test.clj
rexwangcc: thanks 😄  I realized the combination of `rename-keys` and `update-in` is too cumbersome here, thus `postwalk-replace` is a good candidate!

PR #218, opened 2020-11-09
[GH-1045] Arbitrary WGS Inputs

ehigham:
In this change:

# WGS
- specify arbitrary inputs
- update `skip-workflow?` to handle bams as well!

# Database changes:
- CromwellWorkflow is unreleased so we can just shove the options in (don't need a change log)
- copyfile can just use the `CromwellWorkflow` table

# API changes
- combine `workflowOptions` and `common_inputs` into a "common" map `:common { :inputs {} :options {}}`, where `options` are `workflowOptions` renamed.
- implement (extending Jack's work) for xx, wgs and copyfile

# Serialisation change
- we can't store and just load workflow options owing to compatibility concerns (previous workflows didn't store these). I've opted not to store the default environment workflow-options and simply merge these when we talk to cromwell. I'm happy to move this to the loader if this is a point of contention. Thoughts/ideas/opinions welcome.

# Puzzles:
- unsure what to do with artificial inputs like `reference_fasta_prefix` and how that should tie in with being able to specify any input.

# Not included
- doc changes will not be included in this PR. They'll be updated in a subsequent change


jack-r-warren:
Maybe when we do that followup documentation PR we should write down what the default options are for the different modules since this API no longer surfaces that information. Could also address the loading-old-workloads thing by merging the defaults in twice--once when we write to the DB and once when we submit to Cromwell. Would generally be a no-op the second time but it would provide the backwards compatibility from merging at submit-time.

ehigham:
> Maybe when we do that followup documentation PR we should write down what the default options are for the different modules since this API no longer surfaces that information. Could also address the loading-old-workloads thing by merging the defaults in twice--once when we write to the DB and once when we submit to Cromwell. Would generally be a no-op the second time but it would provide the backwards compatibility from merging at submit-time.

Do we want all these options to be made public? This includes service account stuff that's sort of a hack/implementation detail?
I can make all options visible, if you prefer! I'll move the behaviour into the serialisation impl if you prefer - not a big issue for me.

api/src/wfl/module/copyfile.clj
```diff
@@ -14,57 +15,68 @@
...
+  (letfn [(merge-to-json [shared specific]
+            (json/write-str (util/deep-merge shared specific)))
+          (serialize [workflow id]
+            (-> workflow
+              (assoc :id id)
+              (update :inputs #(merge-to-json (:inputs common) %))
+              (update :options #(merge-to-json (:options common) %))))]
+    (let [[id table] (batch/add-workload-table! tx workflow-wdl request)]
+      (jdbc/insert-multi! tx table (map serialize items (range)))
+      (workloads/load-workload-for-id tx id))))
```
ehigham, line 49: The `create-workload!` implementations for copyfile, wgs and xx are almost identical. In a subsequent change, we could explore reducing this technical debt. This means (in theory) we need to test only once for the "batch" workflows (or instantiate a test suite).

api/src/wfl/module/copyfile.clj
```diff
@@ -14,57 +15,68 @@
...
+                    (util/deep-merge default-options options)
+                    {:workload uuid})])
+            (update! [tx [id uuid]]
               (jdbc/update! tx items
-                {:updated (OffsetDateTime/now) :uuid uuid :status status}
+                {:updated (OffsetDateTime/now) :uuid uuid :status "Submitted"}
                 ["id = ?" id]))]
       (run! (comp (partial update! tx) submit!) (:workflows workload))
       (jdbc/update! tx :workload
         {:started (OffsetDateTime/now)} ["uuid = ?" uuid]))))
```
ehigham, line 66: The implementations for `start-workload!` still differ somewhat. This is due to the xx module using (a faked) batch submit to cromwell. In a subsequent change, we could explore moving the other modules onto that endpoint and actually implement it. Given that, maybe we can extract common functionality for these modules (c.f. `create-workload!`)

jack-r-warren:
> Do we want all these options to be made public? This includes service account stuff that's sort of a hack/implementation detail?
> I can make all options visible, if you prefer! I'll move the behaviour into the serialisation impl if you prefer - not a big issue for me.

I think it would be helpful for ops to know in order for them to override stuff. Honestly I think if we are doing hacky stuff I think that's all the more reason to surface what on earth we are doing. And again we don't gotta actually return it from the endpoint I guess -- if there's stuff we really don't want to show them then maybe just documenting what we do want to show them is fine. But since they'd be able to override our hacky stuff I don't see much of a harm.
A bit of my reasoning here is that Cromwell doesn't document any way to get the existing workflow options -- you can get them if you pass `/metadata` through `jq '.submittedFiles.options | fromjson'`, but that submittedFiles field isn't documented and that makes me worry that we don't have guarantees around it.

api/src/wfl/api/workloads.clj
```diff
@@ -110,18 +111,25 @@
...
 
 (defoverload load-workload-impl :default default-load-workload-impl)
+
+;; Common workload operations
+(defn saved-before?
+  "Test if the `_workload` was saved before the `reference` version string."
+  [reference {:keys [version] :as _workload}]
+  (letfn [(decode [v] (map util/parse-int (str/split v #"\.")))
+          (lt? [[x & xs] [y & ys]]
+            (or (< x y) (and (== x y) (every? some? [xs ys]) (lt? xs ys))))]
```
ehigham, line 134: I don't like this implementation and I think I'm over-complicating it. Suggestions welcome.

api/src/wfl/api/workloads.clj
```diff
@@ -110,18 +111,25 @@
...
+    (lt? (decode version) (decode reference))))
```
ehigham, line 135: I think this implementation is over-complicated by I'm not sure how to simplify. Suggestions welcome.

ehigham:
> > Do we want all these options to be made public? This includes service account stuff that's sort of a hack/implementation detail?
> > I can make all options visible, if you prefer! I'll move the behaviour into the serialisation impl if you prefer - not a big issue for me.
> 
> I think it would be helpful for ops to know in order for them to override stuff. Honestly I think if we are doing hacky stuff I think that's all the more reason to surface what on earth we are doing. And again we don't gotta actually return it from the endpoint I guess -- if there's stuff we really don't want to show them then maybe just documenting what we do want to show them is fine. But since they'd be able to override our hacky stuff I don't see much of a harm.
> A bit of my reasoning here is that Cromwell doesn't document any way to get the existing workflow options -- you can get them if you pass `/metadata` through `jq '.submittedFiles.options | fromjson'`, but that submittedFiles field isn't documented and that makes me worry that we don't have guarantees around it.

As an aside - if they don't document if, I don't think we should.

api/src/wfl/module/wgs.clj
```diff
@@ -79,121 +79,113 @@
...
+              (cromwell/query :gotc-dev)
+              (filter #(= pipeline (:name %)))
+              (map #(->> % :id (cromwell/metadata env) :inputs))
+              (map #(some % [:input_bam :input_cram]))
+              (filter #{in-gs})
+              seq))]
+    (let [in-gs  (some inputs [:input_bam :input_cram])
+          [_ object] (gcs/parse-gs-url in-gs)
+          out-gs (str (all/slashify output) object)]
+      (or (exists? out-gs) (processing? in-gs)))))
```
tbl3rd, line 164: Noice.

api/test/wfl/integration/modules/copyfile_test.clj
```diff
@@ -1,58 +1,82 @@
...
+      (fn []
+        (->
+          (make-copyfile-workload-request "gs://fake/foo" "gs://fake/bar")
+          (assoc-in [:common :inputs]
+            {:supports_common_inputs true :overwritten false})
+          (update :items
+            (partial map
+              #(update % :inputs
+                 (fn [xs] (merge xs {:supports_inputs true :overwritten true})))))
+          workloads/execute-workload!)))))
```
tbl3rd, line 82: Noice again.

api/test/wfl/integration/modules/wgs_test.clj
```diff
@@ -79,37 +102,50 @@
...
+        (->
+          (make-wgs-workload-request)
+          (assoc-in [:common :options]
+            {:supports_common_options true :overwritten false})
+          (update :items
+            (partial map
+              #(assoc % :options {:supports_options true :overwritten true})))
+          workloads/execute-workload!
+          :workflows
+          (->> (map (comp verify-workflow-options :options))))))))
```
tbl3rd, line 151: Thanks.

api/test/wfl/tools/workloads.clj
```diff
@@ -102,12 +103,6 @@
     (done! (endpoints/get-workload-status (:uuid workload)))
     nil))
 
-(defn baseline-options-across-workload
-  "True if OPTIONS are present across the entire WORKLOAD."
-  [options workload]
-  (let [unique-options (distinct (map :workflow_options (:workflows workload)))]
-    (every? #(= % (util/deep-merge % options)) unique-options)))
-
```
tbl3rd, line 110: Thanks.

api/test/wfl/unit/modules/xx_test.clj
```diff
@@ -2,38 +2,42 @@
...
-                  :bait_set_name "geoff"}]
-      (is (= "geoff" (:bait_set_name (xx/make-combined-inputs-to-save output common items)))))))
+(deftest test-make-inputs-from-bam
+  (let [sample "gs://fake-input-bucket/folder/sample.bam"
+        inputs (xx/make-inputs-to-save output-url {:input_bam sample})]
+    (is (= sample (:input_bam inputs)))
+    (is (= "sample" (:sample_name inputs)))
+    (is (= "sample.bam" (:base_file_name inputs)))
+    (is (= "sample.bam" (:final_gvcf_base_name inputs)))
+    (is (= (str output-url "folder") (:destination_cloud_path inputs)))))
```
tbl3rd, line 23: Better.

database/changesets/20201029_CromwellWorkflow.xml
```diff
@@ -16,12 +16,14 @@
...
-                uuid       text,        -- Cromwell workflow UUID
-                inputs     text         -- JSON string of non-default workflow inputs
+            CREATE Type CromwellWorkflow AS
+            (
+                id      bigint      -- primary key
+            ,   status  text        -- Cromwell workflow status
+            ,   updated timestamptz -- status update time
+            ,   uuid    text        -- Cromwell workflow UUID
+            ,   inputs  text        -- JSON string of non-default workflow inputs
+            ,   options text        -- JSON string of Cromwell workflow options
```
tbl3rd, line 26: Weirdo.

api/src/wfl/api/workloads.clj
tbl3rd: Will try.

tbl3rd (approved):

api/src/wfl/api/workloads.clj
```diff
@@ -110,18 +111,25 @@
...
   (letfn [(unnilify [m] (into {} (filter second m)))
           (split-inputs [m]
-            (let [keep [:id :finished :status :updated :uuid :workflow_options]]
-              (assoc (select-keys m keep)
-                :inputs (unnilify (apply dissoc m keep)))))
-          (unpack-options [m]
-            (update m :workflow_options #(when % (util/parse-json %))))]
+            (let [keep [:id :finished :status :updated :uuid :options]]
+              (assoc (select-keys m keep) :inputs (apply dissoc m keep))))
+          (load-options [m] (update m :options (fnil util/parse-json "null")))]
```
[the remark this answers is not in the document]
ehigham: There may be uses where passing `nil` into this is an error - building the `fnil`'ness into `parse-json` would make that harder to detect.

PR #219, opened 2020-11-10
[GH-1045] update docs for wgs + `common` block changes

ehigham:
Documentation changes to go with https://github.com/broadinstitute/wfl/pull/218. See that PR for full details.


tbl3rd (approved):
JSON comments?

rexwangcc (approved):
Nice documentation updates!!

Re: JSON comments, I personally am happy to see as many helpful comments as we make, even if they make the code blocks illegal JSONs since I don't expect users copy-paste the examples without editing them. (jsonc might be a better choice but AFAIK it's not supported by mkdocs's highlighter)

PR #220, opened 2020-11-10
[GH-1044] Set default maxRetries workflow option

samanehsan:
### Purpose

Ticket: https://broadinstitute.atlassian.net/browse/GH-1044

### Changes

Set the `maxRetries` workflow option to 1 by default to automatically retry transient failures in Cromwell: https://cromwell.readthedocs.io/en/stable/RuntimeAttributes/#maxretries. 

This includes the `aou`, `wgs`, `xx` and `copyfile` modules.

### Review Instructions

- No instructions.

### DIFF
--- a/api/src/wfl/module/aou.clj
+++ b/api/src/wfl/module/aou.clj
@@ -106,11 +106,11 @@
 
 ;; visible for testing
 (def default-options
-  {; TODO: add :default_runtime_attributes {:maxRetries 3} here
-   :use_relative_output_paths  true
+  {:use_relative_output_paths  true
    :read_from_cache            true
    :write_to_cache             true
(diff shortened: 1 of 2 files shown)


rexwangcc (approved):
LGTM! Thanks for doing this, it's also interesting to see how we have spent much more time talking about it 😅 

One question I have is how can we test/prove this works? I'm not sure what kind of artificial errors we could craft to make workflow fail while persuading Cromwell to retry it 🤔

samanehsan:
@rexwangcc for QA I started a workflow for each module and checked that the workflow options included "maxRetries: 1". I don't think we need to test the actual retry functionality, just that it's getting set correctly!

PR #221, opened 2020-11-12
[GH-1061] don't send back defaults when user specifies reference_fasta_prefix

ehigham:
### Purpose
Using the reference_fasta_prefix to dictate all the references is weird. We should only use it to compute the reference_fasta.

## Changes
- merge the reference_fasta into references if the specifies them
- don't make default values part of the API. This makes back-porting fixes + updated input values easier.

### DIFF
--- a/api/src/wfl/module/wgs.clj
+++ b/api/src/wfl/module/wgs.clj
@@ -47,12 +47,11 @@
     {:cram_ref_fasta       cram_ref_fasta
      :cram_ref_fasta_index (str cram_ref_fasta ".fai")}))
 
-(defn make-references
+(def ^:private default-references
   "HG38 reference, calling interval, and contamination files."
-  [prefix]
   (let [hg38 "gs://gcp-public-data--broad-references/hg38/v0/"]
     (merge references/contamination-sites
-      (references/hg38-genome-references prefix)
+      (references/hg38-genome-references nil)
       {:calling_interval_list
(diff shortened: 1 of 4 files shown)


rexwangcc (approved):

api/src/wfl/module/wgs.clj
```diff
@@ -47,20 +47,19 @@
...
      :cram_ref_fasta_index (str cram_ref_fasta ".fai")}))
 
-(defn make-references
+(def ^:private default-references
   "HG38 reference, calling interval, and contamination files."
-  [prefix]
   (let [hg38 "gs://gcp-public-data--broad-references/hg38/v0/"]
     (merge references/contamination-sites
-      (references/hg38-genome-references prefix)
+      (references/hg38-genome-references nil)
```
tbl3rd, line 54: Could we `(def hg38-genome-references …)` and just merge it as data?

tbl3rd (approved):
Not sure I understand the motivation, but looks good to me.

api/src/wfl/module/wgs.clj
ehigham: > Not sure I understand the motivation, but looks good to me.

The idea is that I think it's weird to specify all references via a reference_fasta_prefix - surely that should only define the reference fasta?

ehigham:
> Not sure I understand the motivation, but looks good to me.

The idea is that I think it's weird to specify all references via a reference_fasta_prefix - surely that should only define the reference fasta?

PR #225, opened 2020-11-13
[GH-1063] Remove Serialisation Internals from Workload Responses

ehigham:
### Purpose
During the hOps demo, it became clear that the `:id` property of a workload (and it's workflows) caused confusion. In this change, I propose to purge this property from the workload responses.
The `:items` table in workloads is an internal detail and we should not be sending that back either.

### DIFF
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -43,9 +43,10 @@
           query {:includeSubworkflows false :start start :end end}]
       (succeed {:results (cromwell/query env query)}))))
 
-(defn strip-internals [workload]
+(defn strip-internals
   "Strip internal properties from the `workload` and its `workflows`."
-  (let [prune #(apply dissoc % [:id])]
+  [workload]
+  (let [prune #(apply dissoc % [:id :items])]
     (prune (update workload :workflows (partial mapv prune)))))
 
 (defn append-to-aou-workload
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -81,11 +81,10 @@
                 (workloads/update-workload! tx workload))))]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
       (succeed
-        (strip-internals
-          (mapv (partial go! tx)
-            (if-let [uuid (-> request :parameters :query :uuid)]
-              [(workloads/load-workload-for-uuid tx uuid)]
-              (workloads/load-workloads tx))))))))
+        (mapv (comp strip-internals (partial go! tx))
+          (if-let [uuid (-> request :parameters :query :uuid)]
+            [(workloads/load-workload-for-uuid tx uuid)]
+            (workloads/load-workloads tx)))))))
 
 (defn post-start
   "Start the workload with UUID in REQUEST."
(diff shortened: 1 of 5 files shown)


api/src/wfl/api/handlers.clj
```diff
@@ -43,6 +43,12 @@
           query {:includeSubworkflows false :start start :end end}]
       (succeed {:results (cromwell/query env query)}))))
 
+(defn strip-internals
+  "Strip internal properties from the `workload` and its `workflows`."
+  [workload]
+  (let [prune #(apply dissoc % [:id :items])]
+    (prune (update workload :workflows (partial mapv prune)))))
```
rexwangcc, line 50: It took me some time to understand what this is doing, very elegant and extensible!

(I wish Postgres had something like BigQuery's `SELECT * EXCEPT(ColumnNameX, [ColumnNameY, ...])`)

api/test/wfl/system/v1_endpoint_test.clj
```diff
@@ -27,14 +28,21 @@
 (defn- verify-succeeded-workload [workload]
   (run! verify-succeeded-workflow (:workflows workload)))
 
+(defn ^:private verify-missing-internals [workload]
```
rexwangcc, line 31: **optional** nit: maybe call it `verify-striped-internals` or `verify-excluded-internals` as `missing` makes me feel they should be there to some extent. (I'm bad at naming so this is totally optional)

rexwangcc (approved):
LGTM,  👍  to also exclude `items`!

api/src/wfl/api/handlers.clj
ehigham: The thing is, we want this stuff for the innards of wfl (private state). We just don't want them in the workload responses.

api/test/wfl/system/v1_endpoint_test.clj
ehigham: sounds good!

tbl3rd (approved):
OK, but I'd start with something like this instead.
```clojure
(defn purge-internal-keys
  "Purge top level internal keys from WORKTHING."
  [workthing]
  (let [[_ _ req-un _ opt-un] (clojure.spec.alpha/describe :wfl.api.spec/workload-response)]
    (select-keys workthing (concat req-un opt-un))))
```

ehigham:
> OK, but I'd start with something like this instead.
> 
> ```clojure
> (defn purge-internal-keys
>   "Purge top level internal keys from WORKTHING."
>   [workthing]
>   (let [[_ _ req-un _ opt-un] (s/describe :wfl.api.spec/workload-response)]
>     (select-keys workthing (concat req-un opt-un))))
> ```

That won't work for nested properties though.

ehigham:
> > That won't work for nested properties though.
> 
> Emphasis on _start_.

I like it. Means we don't have to maintain another list of names. I gave it a go: 
```
(defn strip-internals
  "Strip internal properties from the `workload` and its `workflows`."
  [workload]
  (letfn [(restrict-keys [spec coll]
            (let [[_ _ req-un _ opt-un] (clojure.spec.alpha/describe spec)]
              (select-keys coll (concat req-un opt-un))))]
    (restrict-keys :wfl.api.spec/workload-response
      (update workload :workflows
        (partial mapv #(restrict-keys :wfl.api.spec/workflow %))))))
```

The issue is that the keys are all prefixed by `:wfl.api.spec` so `select-keys` removes everything. While we could strip the prefixes, this is more code than I really wanted to write.

PR #234, opened 2020-11-18
[GH-1071] Fix `exec`-ing an AoU workload twice fails to return the `exec`-ed workload

ehigham:
### Purpose
See ticket.

Caused by a type error. in https://github.com/broadinstitute/wfl/pull/196, I changed the implementation from `start-workload!` from
```clojure
(defmethod workloads/start-workload!
  pipeline
  [tx {:keys [id] :as workload}]
  (do
    (start-aou-workload! tx workload)
    (workloads/load-workload-for-id tx id)))
```
to
```clojure
(defoverload workloads/start-workload! pipeline start-aou-workload!)
```
Without updating `start-aou-workload` to return the the workload if it had already been started.

Note that the changes to the test tools are required so that the test I added doesn't fail due to GH-1072. I think there should be a separate test added for that.

### DIFF
--- a/api/src/wfl/module/aou.clj
+++ b/api/src/wfl/module/aou.clj
@@ -192,7 +192,8 @@
   "Use transaction TX to start the WORKLOAD. This is simply updating the
    workload table to mark a workload as 'started' so it becomes append-able."
   [tx {:keys [id] :as workload}]
-  (when-not (:started workload)
+  (if (:started workload)
+    workload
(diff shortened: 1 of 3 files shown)


tbl3rd (approved):
I wish WFL/PTC didn't care whether `:output` ended with a `/`.

ehigham:
> I wish WFL/PTC didn't care whether `:output` ended with a `/`.

Agreed, I also wish we didn't expect `exec` to return the same workload for AoU.

PR #236, opened 2020-11-18
[GH-1072] Standardize AoU output bucket in DB

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1072

### Changes

- Deslashify before querying with the output bucket, not just when inserting it

### Review Instructions

- Integration test checks that requests that differ in their slashifying have the same response from WFL


api/test/wfl/integration/modules/aou_test.clj
```diff
@@ -68,7 +69,16 @@
...
-  "executing a workload-request twice should not create a new workload"
-  (let [request (make-aou-workload-request)]
-    (is (= (workloads/execute-workload! request)
-          (workloads/execute-workload! request)))))
+  (testing
+    "executing a workload-request twice should not create a new workload"
+    (let [request (make-aou-workload-request)]
+      (is (= (workloads/execute-workload! request)
+             (workloads/execute-workload! request)))))
+  (testing
```
ehigham, line 77: I think a `deftest` is better actually as we can disable it if it fails (+ create a ticket to fix later)

ehigham (approved):

api/test/wfl/integration/modules/aou_test.clj
jack-r-warren: good idea, done

tbl3rd (approved):

PR #241, opened 2020-11-19
[GH-1074] Changelog from branches

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1074

### Changes

- Make changelog use your current branch for log calculation
- Make changelog use the raw commit text if it errors out trying to markdown-ify the PR link (this means it will work if the log includes commits that weren't made via PR)

### Review Instructions

Can verify this by Ed's original reproduction steps if you run them from this branch:

```bash
git checkout -B foo
git add foo.txt && git commit -m "add foo"
./ops/cli.py release
```

### DIFF
--- a/ops/cli.py
+++ b/ops/cli.py
@@ -292,7 +292,7 @@ def _markdownify_commit_msg(commit: str) -> str:
 
 def get_git_commits_since_last_tag(config: WflInstanceConfig) -> None:
     "Read commit messages since last tag, store to config and print."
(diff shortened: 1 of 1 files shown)


ehigham (approved):
Thanks for addressing this so quickly.

jack-r-warren:
Of course! Merging now because clojure tests will 
A. fail due to not having your GH-1076 fix
B. not evaluate this PR since it only changes Python

PR #249, opened 2020-11-20
make error logging less noisy

ehigham:
### Purpose
Currently, all exceptions are being logged as errors. This creates a lot of noise as some failures are expected (e.g. wrong pipeline, no such workflow etc)
In this change, I propose to make the logging less noisy for 0.4.2. We can consider an alternative solution for the main branch.


jack-r-warren (approved):
👍

PR #254, opened 2020-11-23
[GH-1020] Import arrays module in handlers

samanehsan:
### Purpose

Fixes the failing system test for the arrays module.

### Changes

Import arrays module in handlers, otherwise WFL will complain that the pipeline is invalid when trying to create a new `GPArrays` workload.

### Review Instructions

- No instructions.


ehigham (approved):

rexwangcc (approved):
Cannot believe this cost us so much time, finally!! 

Could you add a comment to either the docstring or a new line so we don't forget next time? IntelliJ has not been really helpful on this somehow…

PR #256, opened 2020-11-30
Add linting to PR checks

ehigham:
## Purpose
As discussed, adding linting to the pull-request checks.

## Changes:
- implement `lint` alias for api/ and cloud_function/ modules
- use flake8 as out python linter with default settings.
- add to github action
- extend action to release branches

# BIG SCARY THING
All files have been re-formatted - Suggest you view the diff with whitespace changes hidden.


jack-r-warren (approved):
Looks good to me, I'm personally not bothered by the failing PR test since it did actually get past the lint step and we have a bug for the failure 👍

ehigham:
> Looks good to me, I'm personally not bothered by the failing PR test since it did actually get past the lint step and we have a bug for the failure

Thanks - that is a known failure and captured in https://broadinstitute.atlassian.net/browse/GH-1099

PR #258, opened 2020-11-30
[GH-1097] Fix empty append_to_aou response

ehigham:
### Purpose
https://broadinstitute.atlassian.net/browse/GH-1097

As Jack observed, when you send a sample to workflow launcher that has been (or in the process of being) processed, workflow-launcher skips it. If no samples are processed, I thought I was returning an empty list. I even wrote a test for this. The problem is that Clojure treats `nil` to mean empty, meaning that the integration test I wrote wasn’t that useful.

The solution was to explicitly return the list I created, not the list returned by `jdbc/insert` - I thought they were the same, turns out not.


tbl3rd (approved):
Expect response to be ()?

jack-r-warren (approved):

PR #263, opened 2020-12-02
Cache clojure deps between builds

ehigham:
### Purpose
In investigating https://broadinstitute.atlassian.net/browse/GH-1056, I found that caching build dependencies between jobs lead to quite a significant speed-up in PR checks (~2 mins out of 7).
While my solution for GH-1056 isn't quite there yet, this can definitely be submitted.


jack-r-warren (approved):
Nice find!

PR #265, opened 2020-12-02
[GH-1042] AoU supports custom environments

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1042

### Changes

- If we see an `environment` in the inputs in the request, that'll now override WFL's old behavior of picking "dev" or "prod" based on the Cromwell.
- Add a unit test for exactly this.

### Review Instructions

- See PR unit tests


samanehsan (approved):
LGTM! :octocat:

PR #272, opened 2020-12-08
GH-1095: Batch submit workflows to Cromwell

samanehsan:
### Purpose

https://broadinstitute.atlassian.net/browse/GH-1095

### Changes

- Start/exec all items in a XX or WGS workload to Cromwell using the batch endpoint, instead of sending a request for each item. The batch endpoint will apply the same labels to every workflow, so item-specific labels were removed.
- Move `submit-workload!` to the batch module
- Remove WGS module skip workflow functionality

### Review Instructions

- No instructions.

### DIFF
--- a/api/src/wfl/module/wgs.clj
+++ b/api/src/wfl/module/wgs.clj
@@ -167,11 +167,11 @@
       (let [workflows (group-by #(skip-workflow? env workload %) (:workflows workload))
             skipped-workflows (get workflows true)]
         (concat (map
-               #(assoc % :uuid util/uuid-nil
-                       :status "skipped"
-                       :updated (OffsetDateTime/now))
-               skipped-workflows)
-              (mapcat submit-batch! (group-by :options  (get workflows nil))))))))
+                 #(assoc % :uuid util/uuid-nil
+                         :status "skipped"
+                         :updated (OffsetDateTime/now))
+                 skipped-workflows)
+                (mapcat submit-batch! (group-by :options  (get workflows nil))))))))
 
 (defoverload workloads/create-workload! pipeline create-wgs-workload!)
 
--- a/api/src/wfl/module/wgs.clj
+++ b/api/src/wfl/module/wgs.clj
@@ -1,16 +1,13 @@
 (ns wfl.module.wgs
   "Reprocess (External) Whole Genomes."
-  (:require [clojure.java.io :as io]
-            [clojure.data.json :as json]
+  (:require [clojure.data.json :as json]
             [wfl.api.workloads :as workloads :refer [defoverload]]
             [wfl.environments :as env]
             [wfl.jdbc :as jdbc]
             [wfl.module.all :as all]
             [wfl.references :as references]
-            [wfl.service.cromwell :as cromwell]
             [wfl.service.gcs :as gcs]
             [wfl.util :as util]
-            [wfl.wdl :as wdl]
             [wfl.wfl :as wfl]
             [wfl.module.batch :as batch])
   (:import [java.time OffsetDateTime]))
@@ -115,28 +112,6 @@
       (jdbc/insert-multi! tx table (map serialize items (range)))
       (workloads/load-workload-for-id tx id))))
 
-(defn submit-workload!
-  "Use transaction TX to start the WORKLOAD."
-  [{:keys [uuid] :as workload}]
-  (let [path            (wdl/hack-unpack-resources-hack (:top workflow-wdl))
-        env             (get-cromwell-environment workload)
-        default-options (util/make-options env)]
-    (letfn [(update-workflow [workflow cromwell-uuid]
-              (assoc workflow :uuid cromwell-uuid
-                     :status "Submitted"
-                     :updated (OffsetDateTime/now)))
-            (submit-batch! [[options workflows]]
-              (map update-workflow
-                   workflows
-                   (cromwell/submit-workflows
-                    env
-                    (io/file (:dir path) (path ".wdl"))
-                    (io/file (:dir path) (path ".zip"))
-                    (map (partial make-cromwell-inputs env) workflows)
-                    (util/deep-merge default-options options)
-                    (merge cromwell-label {:workload uuid}))))]
-      (mapcat submit-batch! (group-by :options (:workflows workload))))))
-
 (defoverload workloads/create-workload! pipeline create-wgs-workload!)
 
 (defmethod workloads/start-workload!
@@ -145,8 +120,9 @@
   (letfn [(update-record! [{:keys [id] :as workflow}]
             (let [values (select-keys workflow [:uuid :status :updated])]
               (jdbc/update! tx items values ["id = ?" id])))]
-    (let [now (OffsetDateTime/now)]
-      (run! update-record! (submit-workload! workload))
+    (let [now (OffsetDateTime/now)
+          env (get-cromwell-environment workload)]
+      (run! update-record! (batch/submit-workload! workload env workflow-wdl make-cromwell-inputs cromwell-label))
       (jdbc/update! tx :workload {:started now} ["id = ?" id]))
     (workloads/load-workload-for-id tx id)))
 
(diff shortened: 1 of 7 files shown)


api/src/wfl/service/cromwell.clj
```diff
@@ -281,8 +277,13 @@
...
   [environment wdl imports-zip inputs options labels]
-  ; todo: call the batch endpoint
-  (mapv #(submit-workflow environment wdl imports-zip % options labels) inputs))
+  (->> (partify-workflow wdl
+                         imports-zip
+                         inputs
+                         options
+                         labels)
+       (post-workflow (str (api environment) "/batch"))
+       (mapv #(:id %))))
```
ehigham, line 286: Just want to check that that the order of the UUIDs returned by `submit-workflows` matches the order of workflows passed in?

api/test/wfl/integration/modules/wgs_test.clj
```diff
@@ -16,8 +16,8 @@
 
 (clj-test/use-fixtures :once fixtures/temporary-postgresql-database)
 
-(defn ^:private mock-really-submit-one-workflow [& _]
-  (UUID/randomUUID))
+(defn ^:private mock-submit-workflows [& _]
+  [(UUID/randomUUID)])
```
ehigham, line 20: Think this really needs to return as many UUIDs as sets of workflow inputs that are passed in.

ehigham (commented):
Looking good!

api/test/wfl/integration/modules/wgs_test.clj
samanehsan: Good point!

api/src/wfl/service/cromwell.clj
samanehsan: That's what I noticed from testing, but I couldn't find anything about that specifically in the Cromwell docs. I'll double check with them in slack

api/src/wfl/service/cromwell.clj
```diff
@@ -189,9 +189,8 @@
 
 (defn make-workflow-labels
   "Return the workflow labels from ENVIRONMENT, WDL, and INPUTS."
```
tbl3rd, line 191: "Return workflow labels for WDL."

api/src/wfl/service/cromwell.clj
tbl3rd, line 286: Isn't that just `(mapv :id)`?

api/src/wfl/module/wgs.clj
```diff
@@ -22,14 +23,13 @@
   {:release "ExternalWholeGenomeReprocessing_v1.1.1"
    :top     "pipelines/broad/reprocessing/external/wgs/ExternalWholeGenomeReprocessing.wdl"})
 
-(def cromwell-label-map
+(def ^:private cromwell-labels
   "The WDL label applied to Cromwell metadata."
-  {(keyword wfl/the-name)
-   pipeline})
+  {(keyword wfl/the-name) pipeline})
```
tbl3rd, line 28: Delete those docstrings. They have been out of date for many moons.

api/src/wfl/module/wgs.clj
```diff
@@ -96,30 +96,13 @@
 
 (defn ^:private make-cromwell-inputs
   "Return inputs for reprocessing IN-GS into OUT-GS in ENVIRONMENT."
```
tbl3rd, line 98: Another dead docstring.

api/src/wfl/module/wgs.clj
```diff
@@ -96,30 +96,13 @@
...
-  [environment workflow-inputs]
+  [environment {:keys [inputs]}]
   (-> (util/deep-merge cram-ref hack-task-level-values)
       (util/deep-merge {:references default-references})
       (util/deep-merge (env-inputs environment))
-      (util/deep-merge workflow-inputs)
+      (util/deep-merge inputs)
       (util/prefix-keys (keyword pipeline))))
```
tbl3rd, line 104: Call `deep-merge` once?

api/src/wfl/module/wgs.clj
```diff
@@ -22,14 +23,13 @@
   {:release "ExternalWholeGenomeReprocessing_v1.1.1"
    :top     "pipelines/broad/reprocessing/external/wgs/ExternalWholeGenomeReprocessing.wdl"})
 
-(def cromwell-label-map
+(def ^:private cromwell-labels
```
tbl3rd, line 26: Only one label now.

api/src/wfl/module/wgs.clj
```diff
@@ -149,7 +132,7 @@
                                  util/do-or-nil
                                  seq))
           (processing? [in-gs]
-            (->> {:label cromwell-label :status cromwell/active-statuses}
+            (->> {:label (stringify-cromwell-label (first cromwell-labels)) :status cromwell/active-statuses}
```
tbl3rd, line 135: Just inline this? Or just stringify it at top-level and be done?

samanehsan:
> I wonder if you can share some of the logic with the xx module and move code into the batch module?

There's definitely a lot of similar code between the two modules! I was getting caught up on WGS "skipping" some of the workflows but I can probably move that outside of `submit-workload!` to make that more generalized. Is that what you were thinking @ehigham?

api/src/wfl/module/wgs.clj
```diff
@@ -161,36 +144,46 @@
           out-gs (str (all/slashify output) object)]
       (or (exists? out-gs) (processing? in-gs)))))
```
tbl3rd, line 145: Make this `(or (exists? out-gs) (processing? in-gs) false)` and see below.

api/src/wfl/module/wgs.clj
```diff
@@ -161,36 +144,46 @@
...
+                    (map (partial make-cromwell-inputs env) workflows)
+                    (util/deep-merge default-options options)
+                    (merge cromwell-labels {:workload uuid}))))]
+      (let [workflows (group-by #(skip-workflow? env workload %) (:workflows workload))
+            skipped-workflows (get workflows true)]
+        (concat (map
+                 #(assoc % :uuid util/uuid-nil
+                         :status "skipped"
+                         :updated (OffsetDateTime/now))
+                 skipped-workflows)
```
tbl3rd, line 173: skipped-workflows -=> skipped

api/src/wfl/module/wgs.clj
```diff
@@ -161,36 +144,46 @@
...
+                (mapcat submit-batch! (group-by :options  (get workflows nil))))))))
```
tbl3rd, line 174: `(get workflows nil)` -=> `workflows`

api/src/wfl/module/wgs.clj
```diff
@@ -161,36 +144,46 @@
...
+              (map update-workflow
+                   workflows
+                   (cromwell/submit-workflows
+                    env
+                    (io/file (:dir path) (path ".wdl"))
+                    (io/file (:dir path) (path ".zip"))
+                    (map (partial make-cromwell-inputs env) workflows)
+                    (util/deep-merge default-options options)
+                    (merge cromwell-labels {:workload uuid}))))]
+      (let [workflows (group-by #(skip-workflow? env workload %) (:workflows workload))
```
tbl3rd, line 167: `(let [{skipped? true workflows false} (group-by #(skip-workflow? env workload %) (:workflows workload)) …`
or maybe just handle the skipped workflows conditionally in line?

ehigham:
> > I wonder if you can share some of the logic with the xx module and move code into the batch module?
> 
> There's definitely a lot of similar code between the two modules! I was getting caught up on WGS "skipping" some of the workflows but I can probably move that outside of `submit-workload!` to make that more generalized. Is that what you were thinking @ehigham?

Ah yes the `skip-workload?` gift that keeps on giving. I wonder if we remove this and consider another way of tracking inputs/outputs in the future if this is something that's really required

api/src/wfl/service/cromwell.clj
samanehsan: So that is the behavior that the Cromwell team expects

tbl3rd (approved):
Sorry. Thought I'd approved this.

jack-r-warren (approved):
Looks good to me, just one suggestion--since you're removing the workflow skipping, you can add `^:parallel` to `test-start-wgs-workload` and `test-exec-wgs-workload` in wfl.system.v1-endpoint-test to have them run in parallel

PR #275, opened 2020-12-09
[GH-1055] Remove redundant ct yaml file and generate it on the fly

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1055

### Changes

- Mimic what was done at https://github.com/broadinstitute/gotc-deploy/blob/80d4359bca0aa36b17a532c38081e194b75965ec/.github/workflows/HelmChart-PR-Lint.yml#L86, generate the ct.yaml file on the fly in the Github Action, where it's used.

### Review Instructions

- No instructions.


tbl3rd (approved):

pshapiro4broad:
I think if you rename the `helm` directory to `charts` then you won't need a `ct.yaml` at all. `remote: origin` is the default and `namespace` and `release-label` are only used for `ct install`. Also, I don't think you need a separate step to run `yamale` and `yamllint` as `ct lint` already does that. (Not sure about `helm lint` though)

rexwangcc:
> I think if you rename the `helm` directory to `charts` then you won't need a `ct.yaml` at all. `remote: origin` is the default and `namespace` and `release-label` are only used for `ct install`. Also, I don't think you need a separate step to run `yamale` and `yamllint` as `ct lint` already does that. (Not sure about `helm lint` though)

Thanks for the suggestions! I definitely didn't know `charts` helps get rid of the `ct.yaml`… will try to make a follow up PR!

rexwangcc:
@pshapiro4broad I took your suggestions at https://github.com/broadinstitute/wfl/pull/281/files#diff-c0a233e9bfc15726b241c52edfd27eb06c9a884f7975fed0484aa6dd17fc46a0R35, it works perfectly, thank you!

PR #276, opened 2020-12-10
[GH-1086] Make system tests flexible enough to choose Cromwell deployments

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1086

### Changes

There are multiple potential solutions to this, I picked the most straight-forward one as the first iteration to see how it works ℗ease let me know if you have better solutions):
- [x] Now every workload request testing fixture in `wfl.tools.workloads` will attempt to read Cromwell URL string from an env var `CROMWELL` (bad naming…) if it presents, otherwise fallback to the corresponding default Cromwell URLs (They are left as-is instead of getting changed to `prod`)
- [x] Update docs.
- [x] Fix warnings around `clojure -A` across the project.

- This approach does not rely on `kaocha` so it's compatible with both the parallel mode and the `kaocha` mode.
- This approach affects not only system tests but also other tests that are using `wfl.tools.workloads`. But if no env vars are presented, other tests won't feel.

### Review Instructions

- `CROMWELL=https://cromwell-gotc-auth.gotc-prod.broadinstitute.org/ clojure -M:parallel-test wfl.system.v1-endpoint-test` will submit all workflows to the Cromwell deployment that belongs to `gotc-prod`.

### Discussion

Right now there is a major issue no matter which solution we go: permissions need to be handled better for flexible testing. For example, `CROMWELL=https://cromwell-gotc-auth.gotc-prod.broadinstitute.org/ clojure -M:parallel-test wfl.system.v1-endpoint-test` will submit testing workflows to gotc-prod Cromwell, but all of them will fail and report errors such as:
- `AccessDeniedException: 403 341110568487-compute@developer.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket.`
- `AccessDeniedException: 403 430274355642-compute@developer.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket.`
- …
Since gotc-prod's corresponding SAs don't have sufficient permissions to run the workflows or touch stuff in dev testing buckets. We need to either fix the permission issues or take a further step to update the workload request testing fixtures in `wfl.tools.workloads` to behave differently depends on which deployment they are pointed at. Any ideas?


rhiananthony (approved):
LGTM

rexwangcc:
There are some issues with this implementation, I will take a deeper look. Meanwhile, please hold off reviewing, thanks! :octocat:

api/test/wfl/tools/workloads.clj
```diff
@@ -20,6 +20,14 @@
 
 (def git-branch (delay (util/shell! "git" "branch" "--show-current")))
 
+(defn ^:private load-cromwell-from-env-var!
+  "Load Cromwell url from the env var ENVIRONMENT or throw."
+  []
+  (-> "ENVIRONMENT"
+      util/getenv
+      wfl/throw-or-environment-keyword!
+      stuff :cromwell :url))
```
rexwangcc, line 29: I guess a `def` is sufficient here ..

tbl3rd: Won't a def throw on compilation unless ENVIRONMENT is defined?
And what does it mean if this is nil as tested below?

tbl3rd (commented):
Maybe we just want to load an override Cromwell URL from the process environment.

jack-r-warren (approved):
LGTM

jack-r-warren:
WRT the permissions issues I remembered we discussed that it was okay for prod stuff to be able to read our dev buckets for test data. I think it is okay for that to be a separate ticket (and I think Rhian already did a batch of that work)

rexwangcc:
> WRT the permissions issues I remembered we discussed that it was okay for prod stuff to be able to read our dev buckets for test data. I think it is okay for that to be a separate ticket (and I think Rhian already did a batch of that work)

Thanks, that's what I was figuring too, will leave the permission work in another PR.

PR #277, opened 2020-12-10
[GH-1106] `final_gvcf_base_name` and `base_file_name` should be the same as the `sample_name`

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1106

### Changes

- Just copies the logic from WGS
  - Not exactly, because there's some small behavior differences, like WGS including the unmapped bam suffix or XX not having a slash after the destination cloud path, that I thought it would be good to keep. Not wanting to make this a bigger change

### Review Instructions

- Unit tests cover this, see the test in changed files to see exactly what this change does

### DIFF
--- a/api/src/wfl/module/xx.clj
+++ b/api/src/wfl/module/xx.clj
@@ -74,15 +74,16 @@
 
 ;; visible for testing
 (defn make-inputs-to-save [output-url inputs]
-  (let [sample-name (fn [basename] (first (str/split basename #"\.")))
-        [_ path] (gcs/parse-gs-url (some inputs [:input_bam :input_cram]))
-        basename    (or (:base_file_name inputs) (util/basename path))]
+  (let [sample (some inputs [:input_bam :input_cram])
+        [_ base _] (all/bam-or-cram? sample)
+        leaf   (util/basename base)
+        [_ out-dir] (gcs/parse-gs-url base)]
     (-> inputs
-        (assoc :base_file_name basename)
-        (util/assoc-when util/absent? :sample_name (sample-name basename))
-        (util/assoc-when util/absent? :final_gvcf_base_name basename)
+        (util/assoc-when util/absent? :base_file_name leaf)
+        (util/assoc-when util/absent? :sample_name leaf)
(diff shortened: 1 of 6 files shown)


api/test/wfl/unit/modules/xx_test.clj
```diff
@@ -9,17 +9,17 @@
         inputs (xx/make-inputs-to-save output-url {:input_cram sample})]
     (is (= sample (:input_cram inputs)))
     (is (= "sample" (:sample_name inputs)))
-    (is (= "sample.cram" (:base_file_name inputs)))
-    (is (= "sample.cram" (:final_gvcf_base_name inputs)))
+    (is (= "sample" (:base_file_name inputs)))
+    (is (= "sample" (:final_gvcf_base_name inputs)))
```
tbl3rd, line 13: The test agrees with the example inputs now.

https://github.com/broadinstitute/warp/blob/e64b2499b1a04e3df701556231c317200cf30ee1/pipelines/broad/reprocessing/external/exome/ExternalExomeReprocessing.plumbing.input.json#L5-L6

api/src/wfl/module/xx.clj
```diff
@@ -74,15 +74,16 @@
 
 ;; visible for testing
 (defn make-inputs-to-save [output-url inputs]
-  (let [sample-name (fn [basename] (first (str/split basename #"\.")))
-        [_ path] (gcs/parse-gs-url (some inputs [:input_bam :input_cram]))
-        basename    (or (:base_file_name inputs) (util/basename path))]
+  (let [sample (some inputs [:input_bam :input_cram])
+        [_ base _] (all/bam-or-cram? sample)
+        leaf   (util/basename base)
```
tbl3rd, line 79: Looks like the docstring for util/basename is wrong, or it is misnamed.
Maybe it should be named util/leafname or otherwise fixed.

https://github.com/broadinstitute/wfl/blob/0a7f80ea0255f576dc45d0994af1da823df69c37/api/src/wfl/util.clj#L102

tbl3rd (approved):
Maybe fix util/basename too.

PR #280, opened 2020-12-10
[GH-1090] Automatic SG submission

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1090

### Changes

- Submits uploaded BAMs as SG workload requests to WFL. Requires said WFL to have #285
- Tries to have really good error text if things ever go sideways
- Tests (both unit and a sorta-integration one that mocks WFL)
- Deployment script still needs info on our dev and prod buckets, but it does work when pointed at my sandbox

### Review Instructions

- Tests work and this works on my sandbox. You can run something like this to trigger it on gotc-dev via my sandbox:

```bash
gsutil cp 'gs://broad-gotc-dev-wfl-ptc-test-inputs/single_sample/plumbing/truth/develop/20k/NA12878_PLUMBING.unmapped.bam' "gs://jwarren-wfl-inputs/sg/${RANDOM}.NA12878_PLUMBING.unmapped.bam"
```


functions/sg/main.py
```diff
@@ -0,0 +1,116 @@
...
+    input_file = filter_for_input_paths(
+        PATH_PATTERN,
+        f"gs://{event['bucket']}/{event['name']}"
+    )
+    if input_file is None:
+        return
+
+    inputs = make_inputs(input_file)
+    if inputs is None:
+        return
```
ehigham, line 113: I think it would be clearer if the logic of "only do something if this input file is a BAM" here rather than buried in an internal procedure. Perhaps just move the check 
```python
   if input_file is not None and input_file.endswith('.bam'):
       # post the payload
```

functions/sg/deploy.sh
```diff
@@ -0,0 +1,52 @@
...
+    _WORKLOAD_PROJECT="sg-dev"
+    _OUTPUT_BUCKET="gs://???"
+# gotc-dev -- Jack's sandbox
+elif [ "${TRIGGER_BUCKET_NAME}" == "jwarren-wfl-inputs" ]; then
+    GCLOUD_PROJECT="broad-gotc-dev-storage"
+    SA_EMAIL="sg-submission-fn-non-prod@broad-gotc-dev-storage.iam.gserviceaccount.com"
+    _WFL_URL="https://jwarren-wfl.gotc-dev.broadinstitute.org"
+    _CROMWELL_URL="https://cromwell-gotc-auth.gotc-dev.broadinstitute.org/"
+    _WORKLOAD_PROJECT="sg-dev"
+    _OUTPUT_BUCKET="gs://jwarren-wfl-outputs"
```
ehigham, line 33: perhaps you could strip your sandbox-specific info from the PR?

ehigham (commented):
looks

functions/sg/deploy.sh
jack-r-warren: Totally, the deployment script isn't quite ready to go yet because we need to talk about infrastructure for it

functions/sg/main.py
jack-r-warren: Good idea, done

functions/sg/deploy.sh
```diff
@@ -0,0 +1,39 @@
...
+# ex: bash deploy.sh broad-gotc-dev-wfl-sg-test-inputs
+
+TRIGGER_BUCKET_NAME=${1}
+TRIGGER_EVENT=${2:-"google.storage.object.finalize"}
+REGION=${3:-"us-central1"}
+
+# gotc-prod
+if [ "${TRIGGER_BUCKET_NAME}" == "???" ]; then
+    GCLOUD_PROJECT="broad-gotc-prod-storage"
+    SA_EMAIL="sg-submission-fn-non-prod@broad-gotc-prod-storage.iam.gserviceaccount.com"
```
samanehsan, line 12: Should this be `sg-submission-fn-prod` instead of non-prod?

functions/sg/main.py
```diff
@@ -0,0 +1,107 @@
+import os
+import re
+import requests
+
+WFL_URL = os.environ.get('WFL_URL')
+CROMWELL_URL = os.environ.get('CROMWELL_URL')
+WORKLOAD_PROJECT = os.environ.get('WORKLOAD_PROJECT')
+OUTPUT_BUCKET = os.environ.get('OUTPUT_BUCKET')
+PATH_PATTERN = os.environ.get('PATH_PATTERN')
```
samanehsan, line 9: Where does the `PATH_PATTERN` get set? I don't see it passed in as an environment variable

functions/sg/main.py
```diff
@@ -0,0 +1,107 @@
...
+    headers = {
+        'Authorization': 'Bearer {}'.format(access_token)
+    }
+    return headers
+
+
+def filter_for_input_paths(pattern, event_file):
+    if pattern is None or re.search(pattern, event_file) is not None:
+        return event_file
+    else:
```
samanehsan, line 40: You could get rid of the else block here, since the function will return None either way

functions/sg/deploy.sh
```diff
@@ -0,0 +1,39 @@
...
+    _WFL_URL="https://gotc-prod-wfl.gotc-prod.broadinstitute.org"
+    _CROMWELL_URL="https://cromwell-gotc-auth.gotc-prod.broadinstitute.org/"
+    _WORKLOAD_PROJECT="sg-prod"
```
samanehsan, line 15: Are these the actual project names or placeholder values? I'm wondering because I don't see them listed in gcloud 🤔

functions/sg/main.py
jack-r-warren: It isn't currently set. At one point I think the plan was to put this function on `gs://broad-gotc-prod-storage`, and this pattern could be used so we wouldn't trigger on every bam. I suppose our plans have changed enough that this functionality probably isn't necessary, I'll go ahead and remove it.

functions/sg/deploy.sh
jack-r-warren: It doesn't exist but none of the gotc-prod deployment config exists, so fair enough

functions/sg/deploy.sh
jack-r-warren: project gets passed to WFL's `project` field. It felt weird still calling the environment variable `WFL_ENVIRONMENT` [like the aou fn does](https://github.com/broadinstitute/wfl/blob/main/functions/aou/main.py#L57) because we don't ever actually have sg environments anywhere (we use the wgs ones)

jack-r-warren: Do you think I should rename this?

samanehsan: Oh! I didn't see `GCLOUD_PROJECT ` a few lines up 🤦‍♀️ I think the name makes sense

functions/sg/deploy.sh
samanehsan: Oh okay! Could you make a ticket for setting up those service accounts (if there isn't one already)?

functions/sg/main.py
samanehsan: Ok, thanks for clarifying!

samanehsan (approved):
Thanks for making those changes! Great work!! :octocat:

PR #282, opened 2020-12-10
[GH-1111] Handle environments in AOU file paths

samanehsan:
### Purpose

https://broadinstitute.atlassian.net/browse/GH-1111

### Changes

- Change the `env` parameter to accept "staging" in addition to "dev" and "prod".
- Only cleanup files in the input bucket from the `env` directory. For example, running the script with the "staging" environment will clean up files in the `staging` directory of the `dev-aou-arrays-input` bucket. This is required because dev and staging could have samples with the same exact chip name, chip well barcode and analysis version.
- Add "staging" environment to `keep_files`.

### Review Instructions

- No instructions.


tbl3rd (approved):
I hope the overlap of dev and staging won't confuse things more.

samanehsan:
I realized since the script leaves behind the "latest" analysis version, it will need to handle both the old and new file paths...fixing this now!

PR #283, opened 2020-12-11
[GH-757] Grab deployment keys from Vault, not secrets

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-757

### Changes

- Grabs deployment keys from Vault

### Review Instructions

- The same thing is done in both deployment and the PR tests, so hopefully the test here passes
- **Question: should the dockerhub creedentials used for deployment also be replaced with ones from Vault? That's slightly higher risk**
- After merging this ticket there's a few secrets that we can safely delete, see the ticket for details

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -47,13 +47,67 @@ jobs:
         with:
           python-version: '3.8' # Version range or exact version of a Python version to use, using SemVer's version range syntax
 
-      - name: Configure SSH Keys
+        # Get secret from vault
+      - name: Authenticate to GCloud
+        id: gcloud
+        env:
+          ROLE_ID: ${{ secrets.ROLE_ID }}
+          SECRET_ID: ${{ secrets.SECRET_ID }}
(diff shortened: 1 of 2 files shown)


jack-r-warren:
Hm, test failing, says the key is invalid format but it isn't when I try it locally--I'm suspecting that the Vault role doesn't have access to read the `secret/dsde/gotc/dev/ci/wfl_deploy_keys` secret. I've let Rhian know, I think this is probably a simple fix

jack-r-warren:
> Hm, test failing, says the key is invalid format but it isn't when I try it locally--I'm suspecting that the Vault role doesn't have access to read the `secret/dsde/gotc/dev/ci/wfl_deploy_keys` secret. I've let Rhian know, I think this is probably a simple fix

Debugged with Rhian and he realized that some existing keys weren't formatted properly, so I used `fold` to add newlines and they're all good now. The unused `dsde-pipelines` key in Vault was formatted even weirder, it had a newline in it, but I don't want to invalidate it or something by messing it up so I just left it.

rexwangcc (approved):
LGTM, thanks!

PR #284, opened 2020-12-14
[GH-1111] Update AOU cloud fn to handle paths with or without env prefix

samanehsan:
### Purpose

https://broadinstitute.atlassian.net/browse/GH-1111

### Changes

Update the AOU cloud function to handle both the old/existing input file paths and the "new" file paths that will be prefixed with the mercury environment. This will allow us to re-deploy the cloud function without stopping PTC/having to coordinate with BITS.

### Review Instructions

- No instructions.

### DIFF
--- a/functions/aou/main.py
+++ b/functions/aou/main.py
@@ -37,15 +37,20 @@ def get_auth_headers():
 
 
 def get_manifest_path(object_name):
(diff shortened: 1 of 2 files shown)


jack-r-warren (approved):
I'm a fan of this, nice work

PR #290, opened 2021-01-06
[GH-1105] Add update loop plumbing

ehigham:
### Purpose
In this change, I'm adding the plumbing for the update loop + whatever task that needs to be done periodically. I've left the actual update function empty as I want the plumbing reviewed rather than the update code itself.

I'm using futures for the jetty server and the update loop because if either fail, I want the entire service to come down. I also wanted to tie their lifetimes somehow. Unfortunately, This means I'm using 2 extra threads. I could use the main thread to poll the status of the jetty server, but I thought that this approach was simpler for now.

### Review Instructions
What do people think of this approach? It's pretty straightforward but are there other designs people have thought about?

### DIFF
--- a/api/src/wfl/server.clj
+++ b/api/src/wfl/server.clj
@@ -80,7 +80,7 @@
     (log/info "starting workload update loop")
     (update-workloads!)
     (future
-      (while [true]
+      (while true
         (update-workloads!)
         (.sleep TimeUnit/SECONDS 20)))))
 
--- a/api/src/wfl/server.clj
+++ b/api/src/wfl/server.clj
@@ -104,9 +104,12 @@
 
 (defn ^:private await-some [& futures]
   (loop []
-    (when-not (some future-done? futures)
-      (.sleep TimeUnit/SECONDS 1)
-      (recur))))
+    (let [fs (filter future-done? futures)]
+      (if (not-empty fs)
+        @(first fs)
+        (do
+          (.sleep TimeUnit/SECONDS 1)
+          (recur))))))
 
 (defn run
   "Run child server in ENVIRONMENT on PORT."
--- a/api/src/wfl/server.clj
+++ b/api/src/wfl/server.clj
@@ -14,8 +14,7 @@
             [wfl.api.routes :as routes]
             [wfl.util :as util]
             [wfl.wfl :as wfl])
-  (:import [org.eclipse.jetty.util.log Log Logger]
-           (java.util.concurrent TimeUnit)))
+  (:import (java.util.concurrent TimeUnit)))
 
 (def description
   "The purpose of this command."
@@ -72,23 +71,9 @@
       wrap-internal-error
       (wrap-json-response {:pretty true})))
 
-(defn stfu-jetty
-  "Set up a stub logger to shut Jetty up."
-  []
-  (Log/setLog
-   (proxy [Logger] []
-     (debug ([thrown]) ([msg & args]))
-     (getLogger [name] this)
-     (getName [] "JettyStfu")
-     (ignore [ignored])
-     (info ([thrown]) ([msg & args]))
-     (isDebugEnabled [] false)
-     (setDebugEnabled [enabled])
-     (warn ([thrown]) ([msg & args])))))
-
 (defn ^:private start-workload-manager []
   (letfn [(update-workloads! [])] ;; todo
-    (pprint "starting workload update loop")
+    (log/info "starting workload update loop")
     (update-workloads!)
     (future
       (loop []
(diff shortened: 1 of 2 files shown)


api/src/wfl/server.clj
```diff
@@ -85,11 +86,34 @@
      (setDebugEnabled [enabled])
      (warn ([thrown]) ([msg & args])))))
 
+(defn ^:private start-workload-manager []
+  (letfn [(update-workloads! [])] ;; todo
+    (pprint "starting workload update loop")
```
jack-r-warren, line 91: What are the pros and cons of routing this through `pprint` rather than `log/info`?

api/src/wfl/server.clj
```diff
@@ -85,11 +86,34 @@
...
+    (update-workloads!)
+    (future
+      (loop []
+        (update-workloads!)
+        (.sleep TimeUnit/SECONDS 20)
+        (recur)))))
+
+(defn ^:private start-server [args]
+  (let [port {:port (util/is-non-negative! (first args))}]
+    (pprint [wfl/the-name port])
```
jack-r-warren, line 101: Same goes here. Not bad the way it is, just figure routing through `log` might help us redirect output more easily via the log4j config at some point

jack-r-warren (approved):
LGTM!

api/src/wfl/server.clj
ehigham: None - this module doesn't use any of the log stuff presumably because it interferes with jetty (hence tom's `stfu-jetty`)

ehigham: Though, having commented out `stfu-jetty`, i'm not sure what exactly it was doing before

jack-r-warren: I could be totally off the mark here--I think `stfu-jetty` predated our usage of logging and was just another way to quiet it down. There's actually a small chance `stfu-jetty` isn't necessary anymore because [I might've managed to configure Log4j2 to not report anything less than errors from Jetty](https://github.com/broadinstitute/wfl/blob/main/api/resources/log4j2.xml#L21)

ehigham: Switched to `log/info` - jetty seems to have got the picture.

api/src/wfl/server.clj
```diff
@@ -71,25 +71,33 @@
...
+    (future (jetty/run-jetty app port))))
+
+(defn ^:private await-some [& futures]
+  (loop []
+    (let [fs (filter future-done? futures)]
+      (if (not-empty fs)
+        @(first fs)
+        (do
+          (.sleep TimeUnit/SECONDS 1)
+          (recur))))))
```
tbl3rd, line 96: await-some is called once and returns when any future is done?
Do you intend for this to be like select(2)?

api/src/wfl/server.clj
```diff
@@ -71,25 +71,33 @@
...
-     (warn ([thrown]) ([msg & args])))))
+(defn ^:private start-workload-manager []
+  (letfn [(update-workloads! [])] ;; todo
+    (log/info "starting workload update loop")
+    (update-workloads!)
+    (future
+      (loop []
+        (update-workloads!)
+        (.sleep TimeUnit/SECONDS 20)
+        (recur)))))
```
tbl3rd, line 82: When does this loop exit?

api/src/wfl/server.clj
```diff
@@ -71,25 +71,33 @@
...
   "Run child server in ENVIRONMENT on PORT."
   [& args]
-  (pprint (into ["Run:" wfl/the-name "server"] args))
-  (let [port {:port (util/is-non-negative! (first args))}]
-    (pprint [wfl/the-name port])
-    (stfu-jetty)
-    (jetty/run-jetty app port)))
+  (log/info "Run:" wfl/the-name "server" args)
+  (let [manager (start-workload-manager)]
+    (await-some manager (start-server args))))
```
tbl3rd, line 103: Always waiting for start-server?

tbl3rd (approved):
I do not understand the futures handshake.

api/src/wfl/server.clj
ehigham: The loop doesn't end - it loops for the lifetime of the server.
The reason for the first call is I want to update the state of the wfl database BEFORE we start listening to requests.

api/src/wfl/server.clj
ehigham: Yes, why wouldn't we?

api/src/wfl/server.clj
ehigham: I don't understand the question - i could have used select instead of filter I suppose.
The point is that I want to join both futures such that is one returns, the server closes.
Neither of these futures are meant to return - only if an error occurs then will they.

ehigham:
> I do not understand the futures handshake.

What specifically don't you understand?

tbl3rd (approved):
Awesome. Thanks.

PR #294, opened 2021-01-07
[GH-1110] Rename `cromwell` to `executor`

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1110

### Changes

- Changed the column in the database
- Spec accepts either `executor` or `cromwell`, top-level handler standardizes it to `executor`
  - The standardization happens just below reitit, so the behavior can't really be tested from integration tests--system tests address it instead
- Change modules etc to use `executor` key rather than a `cromwell` key
- Update docs
- Update cloud function
  - Don't actually need to redeploy it, WFL has backwards compatibility

### Review Instructions

- I dunno... tests pass. Re implementation is there other stuff I should change? Stuff I changed that I shouldn't have?


api/src/wfl/module/batch.clj
```diff
@@ -19,8 +19,8 @@
   [tx {:keys [release top] :as _workflow-wdl} {:keys [pipeline] :as workload-request}]
   (let [[{:keys [id]}]
         (-> workload-request
-            (select-keys [:creator :cromwell :input :output :project])
-            (update :cromwell all/de-slashify)
+            (select-keys [:creator :executor :input :output :project])
+            (update :executor all/de-slashify)
```
rexwangcc, line 23: Since `workload-request` takes `(or ::executor ::cromwell)` now, I'm worrying if this raises a NPE if user sends in `:cromwell` than `:executor`?

rexwangcc (commented):
Looks good, I left one question.

samanehsan (approved):
I see there's one merge conflict, but once that's resolved I think this is good to merge!

api/src/wfl/module/batch.clj
jack-r-warren: Nope, https://github.com/broadinstitute/wfl/pull/294/files#diff-3d78d4a4fd988b47b806c45fd53571a042130ed23c6685508e2e55de7c5417edR75 ensures that incoming requests that we end up actually processing from WFL always have `:executor`

rexwangcc: Good point! I missed that part. NVM then

rexwangcc (approved):

PR #296, opened 2021-01-07
[GH-1125] Don't finish unstarted workloads

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1125

### Changes

- Make the SQL query for unfinished workflows also include workflows with no status, so we correctly leave such workflows untouched if we create and then call update on it
- Make all the module integration tests catch this error

### Review Instructions

- Integration tests check this via Ed's reproduction steps

### DIFF
--- a/api/src/wfl/service/postgres.clj
+++ b/api/src/wfl/service/postgres.clj
@@ -85,8 +85,9 @@
                     {:finished (OffsetDateTime/now)} ["id = ?" id]))))
 
 (defn update-workload!
-  "Use transaction TX to update WORKLOAD statuses."
+  "Use transaction TX to update WORKLOAD statuses if it has been started."
   [tx workload]
-  (do
-    (update-workflow-statuses! tx workload)
-    (update-workload-status! tx workload)))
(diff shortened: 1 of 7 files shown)


api/src/wfl/service/postgres.clj
```diff
@@ -78,7 +78,7 @@
 (defn update-workload-status!
   "Use `tx` to mark `workload` finished when all `workflows` are finished."
   [tx {:keys [id items] :as _workload}]
-  (let [query (format "SELECT id FROM %%s WHERE status NOT IN %s"
+  (let [query (format "SELECT id FROM %%s WHERE status IS NULL OR status NOT IN %s"
```
ehigham, line 81: While this fixes it, I'm not convinced execution should ever reach here if the workload hasn't been started. Maybe add a check at the top of the `update-workload!` implementation?

jack-r-warren: Good idea, left this change since it is a "more correct" SQL query but added your suggestion

ehigham (approved):

PR #297, opened 2021-01-08
GH-1104: Factor build code out of WFL.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1104

### Changes

- Run System/exit to avoid waiting for the JVM thread scheduler to time out.

### Review Instructions

- The build functions run longer than it seems they should now. Not sure why yet.


api/module.mk
```diff
@@ -33,14 +33,14 @@ JAR_LINK     := $(DERIVED_TARGET_DIR)/wfl.jar
 
 $(PREBUILD): $(SCM_SRC) $(SCM_RESOURCES)
```
ehigham, line 34: I think you can remove the dependence on SCM_SRC here as the only thing that affects prebuild is your new build.clj file + this + deps.edn?

ehigham (approved):
noice!

api/module.mk
tbl3rd: Yes. Thanks.

rexwangcc (approved):
Looks great!

PR #298, opened 2021-01-08
[GH-1105] Update workloads in background periodically

ehigham:
## Purpose
https://broadinstitute.atlassian.net/browse/GH-1105

Implement the `update-workloads!` function called by the "workload-manager" (i.e. the background loop that keeps the workload database up-to-date).
- Only update "live" workloads (i.e. those that are not `:finished`)

Remove updating workloads from the `api/v1/workload` endpoint handler
- This is now done periodically.
- Still guarantees eventual consistency like the lazy implementation before.

In the test tools, get workload updates from workflow-launcher itself instead of talking to terra/cromwell directly.

## Questions
- Not sure how to best handle exceptions in the update loop. I'm expecting some db conflicts but not much else. If we can't connect to the database, we probably don't want to bring the server down. If there's a data issue, maybe we do?

### DIFF
--- a/api/src/wfl/api/workloads.clj
+++ b/api/src/wfl/api/workloads.clj
@@ -104,9 +104,9 @@
   :default
   [_ body]
   (throw
-    (ex-info "Failed to update workload - no such pipeline"
-      {:cause body
-       :type  ::invalid-pipeline})))
+   (ex-info "Failed to update workload - no such pipeline"
+            {:cause body
+             :type  ::invalid-pipeline})))
 
 (defn default-load-workload-impl
   [tx workload]
--- a/api/src/wfl/api/workloads.clj
+++ b/api/src/wfl/api/workloads.clj
@@ -102,9 +102,11 @@
 
 (defmethod update-workload!
   :default
-  [tx workload]
-  (postgres/update-workload! tx workload)
-  (load-workload-for-id tx (:id workload)))
+  [_ body]
+  (throw
+    (ex-info "Failed to update workload - no such pipeline"
+      {:cause body
+       :type  ::invalid-pipeline})))
 
 (defn default-load-workload-impl
   [tx workload]
--- a/api/src/wfl/api/workloads.clj
+++ b/api/src/wfl/api/workloads.clj
@@ -64,13 +64,13 @@
   "Use transaction `tx` to load `workload`(s) with `project`."
   [tx project]
   (let [do-load   (partial load-workload-impl tx)]
-    (map do-load (jdbc/query tx ["SELECT * FROM workload WHERE project = ?" project]))))
+    (mapv do-load (jdbc/query tx ["SELECT * FROM workload WHERE project = ?" project]))))
 
 (defn load-workloads
   "Use transaction TX to load all known `workloads`"
   [tx]
   (let [do-load (partial load-workload-impl tx)]
-    (map do-load (jdbc/query tx ["SELECT * FROM workload"]))))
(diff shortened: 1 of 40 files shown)


api/src/wfl/service/postgres.clj
```diff
@@ -42,6 +42,15 @@
...
     (throw (ex-info (format "Table %s does not exist" table) {:cause "no-such-table"}))))
 
+(defn ^:private cromwell-status
+  "`status` of the workflow with `uuid` in `cromwell`."
+  [cromwell uuid]
+  (-> (str/join "/" [cromwell "api" "workflows" "v1" uuid "status"])
+      (http/get {:headers (once/get-auth-header)})
+      :body
+      util/parse-json
+      :status))
```
tbl3rd, line 52: Consolidate this cromwell-status with the one in workloads.clj?

api/src/wfl/api/workloads.clj
```diff
@@ -64,13 +64,13 @@
...
   (let [do-load   (partial load-workload-impl tx)]
-    (map do-load (jdbc/query tx ["SELECT * FROM workload WHERE project = ?" project]))))
+    (mapv do-load (jdbc/query tx ["SELECT * FROM workload WHERE project = ?" project]))))
 
 (defn load-workloads
   "Use transaction TX to load all known `workloads`"
   [tx]
   (let [do-load (partial load-workload-impl tx)]
-    (map do-load (jdbc/query tx ["SELECT * FROM workload"]))))
+    (mapv do-load (jdbc/query tx ["SELECT * FROM workload"]))))
```
tbl3rd, line 73: Why the `mapv`s?

tbl3rd: Ah, to force query completion within the transaction?

api/test/wfl/tools/workloads.clj
```diff
@@ -206,13 +199,9 @@
...
     (wfl.api.workloads/load-workload-for-id tx id)))
 
-;; `doall` is required here for testing otherwise the moment test uses
-;; the result and tries to force realizing, the db tx is already closed
-;;
 (defn load-workloads-with-project [project]
   (jdbc/with-db-transaction [tx (fixtures/testing-db-config)]
-    (doall
-     (wfl.api.workloads/load-workloads-with-project tx project))))
+    (wfl.api.workloads/load-workloads-with-project tx project)))
```
tbl3rd, line 204: The `mapv`s explained.

api/src/wfl/server.clj
```diff
@@ -98,9 +116,10 @@
       (isCancelled [_] false)
       (isDone [_] (.isStopped server)))))
```
tbl3rd, line 117: The start-workload-manager docstring is in the "wrong" place. Intentional?

tbl3rd (approved):
Better and simpler than without the background updating. Nice.

api/src/wfl/service/postgres.clj
ehigham: Good spot! That one isn't used so I'll just remove it!

api/src/wfl/api/workloads.clj
ehigham: Exactly - I could have added `doall`s everywhere but I think eager evaluation is used more commonly.

api/src/wfl/server.clj
ehigham: Thanks!

PR #300, opened 2021-01-11
[GH-1129] Avoid NPE in /exec by properly renaming keys

jack-r-warren:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1129

### Changes

- Also rename keys in /exec
- Make /create and /exec look more similar\
- Add a test to catch this

### Review Instructions

- Reproduction steps [are here](https://broadinstitute.atlassian.net/browse/GH-1129), against a local instance of this branch it'll not throw a NPE.
- This is my bad, I just didn't get how exec worked--I thought it delegated higher up the chain and it doesn't, so we have to rename the keys in both cases. I tried to make the functions a bit more similar to highlight this fact in the future, since they do actually have to do the same thing.

### DIFF
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -111,10 +111,10 @@
 (defn post-exec
   "Create and start workload described in BODY of REQUEST"
   [request]
-  (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-    (let [workload-request (-> (:body-params request)
-                               (rename-keys {:cromwell :executor}))]
-      (logr/info "post-exec endpoint called: " workload-request)
+  (let [workload-request (-> (:body-params request)
+                             (rename-keys {:cromwell :executor}))]
+    (logr/info "post-exec endpoint called: " workload-request)
+    (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
       (->>
        (gcs/userinfo request)
        :email
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -62,17 +62,17 @@
            succeed))))
 
 (defn post-create
-  "Create the workload described in BODY of REQUEST."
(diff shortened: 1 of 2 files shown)


api/src/wfl/api/handlers.clj
```diff
@@ -112,8 +112,9 @@
   "Create and start workload described in BODY of REQUEST"
   [request]
   (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-    (let [workload-request (:body-params request)]
-      (logr/info "executing workload-request: " workload-request)
+    (let [workload-request (-> (:body-params request)
+                               (rename-keys {:cromwell :executor}))]
```
tbl3rd, line 116: Lift this out of the transaction for consistency with `post-create`.

tbl3rd (approved):

api/src/wfl/api/handlers.clj
```diff
@@ -62,17 +62,17 @@
...
 (defn post-create
-  "Create the workload described in BODY of REQUEST."
-  [{:keys [parameters] :as request}]
-  (let [{:keys [body]} parameters
-        {:keys [email]} (gcs/userinfo request)]
-    (logr/infof "post-create endpoint called: body=%s" body)
+  "Create the workload described in REQUEST."
+  [request]
+  (let [workload-request (-> (:body-params request)
+                             (rename-keys {:cromwell :executor}))
```
rexwangcc, line 68: interesting, now I almost feel like we want to put this renaming in a middleware :P (maybe too crazy?)

rexwangcc (approved):

api/src/wfl/api/handlers.clj
jack-r-warren: :runs screaming:

tbl3rd: No. This renaming is a workaround for API compatibility.

PR #301, opened 2021-01-12
[Snack] Version the docs

jack-r-warren:
### Purpose

- Ed mentioned that it is weird that we don't version docs at all, I think it is a good point

### Changes

- Enabled "keep_files" on our GH Pages deploy action so it doesn't delete things not replicated in the last build, because...
- ...release actions now deploy the docs into a subfolder, so https://broadinstitute.github.io/wfl/v0.5.0 or whatever
- No one would know about that link so now we add a note to the release [here](https://github.com/broadinstitute/wfl/releases)
  - While we're at it, inject the most recent changelog section into those notes
- **TL;DR: current docs stay the same, but now we also get versioned docs linked to from the release for free**

### Review Instructions

- I had a touch of free time and figured I might be able to tackle this. Just an idea, not beholden to it at all

### DIFF
--- a/.github/workflows/release_latest_version.yml
+++ b/.github/workflows/release_latest_version.yml
@@ -27,7 +27,7 @@ jobs:
           # Maps tcp port 5432 on service container to the host
           - 5432:5432
     env:
-      MODULES: api ui functions/aou functions/sg
+      MODULES: api ui functions/aou functions/sg docs
 
     steps:
       - name: prepare sources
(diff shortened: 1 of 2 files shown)


.github/workflows/release_latest_version.yml
```diff
@@ -170,3 +170,29 @@ jobs:
...
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ steps.version.outputs.VERSION }}
+          body: |
+            Docs available at [broadinstitute.github.io/wfl/${{ steps.version.outputs.VERSION }}](https://broadinstitute.github.io/wfl/${{ steps.version.outputs.VERSION }})
+
+            [Changes](https://github.com/broadinstitute/wfl/blob/${{ steps.version.outputs.VERSION }}/CHANGELOG.md):
+            ${{ steps.version.outputs.CHANGELOG }}
```
rexwangcc, line 198: My understanding is this step of Action will create a separate git tag for the docs, in addition to what `make_git_tag` in `cli.py` created?

jack-r-warren: The `create-release` action here only makes a tag if one doesn't already exist (see how `commitish` [here](https://github.com/actions/create-release#inputs) is unused if the tag already exists). The action here is like if we went to [the releases page ourselves](https://github.com/broadinstitute/wfl/releases/new) and made a release from an existing tag.

Tags are already present on the [releases page](https://github.com/broadinstitute/wfl/releases) anyway, this just formalizes it so we can add some description text

rexwangcc: I see. Thanks for the explanation!

rexwangcc (approved):

PR #306, opened 2021-01-13
[GH-1144] Helper script for importing Data Repo data into Terra

samanehsan:
### Purpose

Provide an example of how to get data from a snapshot in the Data Repo into a Terra workspace:
https://broadinstitute.atlassian.net/browse/GH-1144

### Changes

Given a data repo snapshot, this script will:
1. Query the snapshot sample table in BigQuery
2. Save the table contents as a TSV file and return a list of the samples/entities formatted for WFL. This script uses the `datarepo_row_id` as the unique id (AKA the "entity id") of the sample in Terra.
3. Upload the TSV file to the specified Terra workspace data table

Note that the default snapshot uses the datarepo integration test dataset, since we do not have a test arrays dataset yet.

### DIFF
--- a/ops/data_repo_terra/import_snapshot_data.py
+++ b/ops/data_repo_terra/import_snapshot_data.py
@@ -1,0 +1,89 @@
+import argparse
+import csv
+import io
+import requests
+import uuid
+import google.auth
+import google.auth.transport.requests
+from google.cloud import bigquery
+from google.oauth2 import service_account
+
+
+def get_service_account_credentials(service_account_path, scopes):
+    credentials = service_account.Credentials.from_service_account_file(service_account_path, scopes=scopes)
+    if not credentials.valid:
+        credentials.refresh(google.auth.transport.requests.Request())
+    return credentials
+
+def get_snapshot_data(google_project, datarepo_snapshot, service_account_path):
+    """Query the 'sample' table for a datarepo_snapshot. The service account must
+    have at least data custodian permissions on the snapshot in order to
+    run a query job."""
+    scopes = ['https://www.googleapis.com/auth/bigquery']
+    credentials = get_service_account_credentials(service_account_path, scopes)
+    client = bigquery.Client(project=google_project, credentials=credentials)
+    query = f'SELECT * FROM `{google_project}.{datarepo_snapshot}.sample`'
+    query_job = client.query(query)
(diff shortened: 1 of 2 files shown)


ops/data_repo_terra/import_snapshot_data.py
```diff
@@ -0,0 +1,92 @@
...
+    import_url = f'{terra_url}/api/workspaces/{terra_workspace}/flexibleImportEntities'
+    scopes = ['email', 'openid', 'profile']
+    credentials = get_service_account_credentials(service_account_path, scopes)
+    headers = {'Authorization': f'Bearer {credentials.token}'}
+    with open (tsv_file, 'rb') as f:
+        contents = f.read()
+    response = requests.post(import_url,
+                             headers=headers,
+                             files={'entities': io.BytesIO(contents),
+                                    'type': 'text/tab-separated-values'})
```
rexwangcc, line 61:
```suggestion
                                    'type': 'text/tab-separated-values'})
    response.raise_for_status()
```

rexwangcc: nit?

ops/data_repo_terra/import_snapshot_data.py
```diff
@@ -0,0 +1,92 @@
...
+    print(f'The following samples have been uploaded to {terra_workspace}:')
+    for sample in entities:
+        print(sample)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description="Import Data Repo snapshot data into a Terra workspace.",
+                                     usage="%(prog)s [-h] [DATAREPO_SNAPSHOT TERRA_URL TERRA_WORKSPACE SERVICE_ACCOUNT_PATH] [...]")
+    parser.add_argument("--datarepo_snapshot",
+                        default="zerosnapshot",
```
rexwangcc, line 78: should it be `wflsnapshot`, or zero means a different thing here?

rexwangcc (approved):
Looks great! Thanks for sharing this helper script!

ops/data_repo_terra/import_snapshot_data.py
samanehsan: The original dataset had `zero` in it instead of `wfl` so I used it here for consistency!

PR #308, opened 2021-01-14
[GH-1145] Add AppSec Github Trivy Action

dinvlad:
### Purpose
As part of the initial integration of this repo into the Terra ecosystem, this change adds [Trivy](https://github.com/aquasecurity/trivy) security scan for the Docker images.

The scan will run as a GitHub Action on every PR and give you feedback if the Docker image contains Critical OS vulnerabilities. One can then either:

* fix the issue (usually just by updating the packages in the Dockerfile, or the base image itself), or
* ignore it, and document the reason with AppSec. Our compliance process has a timeline of 14 days for patching Critical vulnerabilities, or documenting why they are ignored.

As an alternative, [AppSec "blessed" base images](https://github.com/broadinstitute/dsp-appsec-blessed-images) can be used to skip over this check.

More info:
https://github.com/broadinstitute/dsp-appsec-trivy-action
https://github.com/broadinstitute/dsp-appsec-blessed-images

### Changes

- New GitHub workflow: [.github/workflows/trivy.yml](.github/workflows/trivy.yml)

### Review Instructions

- No instructions.

Thanks!

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -16,6 +16,13 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      # The Dockerfile copies this, so it needs
+      # to exist for the build to succeed
+      # (even though it's not scanned)
+      - run: |
+          mkdir target
+          touch target/wfl-api.jar
+
       - uses: broadinstitute/dsp-appsec-trivy-action@v1
         with:
           context: ${{ matrix.context }}
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      # The Dockerfile copies this, so it needs
+      # API Dockerfile copies this, so it needs
       # to exist for the build to succeed
       # (even though it's not scanned)
       - working-directory: ${{ matrix.context }}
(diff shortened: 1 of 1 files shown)


.github/workflows/trivy.yml
```diff
@@ -0,0 +1,29 @@
...
+        context:
+        - api
+        - ui
+
+    steps:
+      - uses: actions/checkout@v2
+
+      # API Dockerfile copies this, so it needs
+      # to exist for the build to succeed
+      # (even though it's not scanned)
```
rexwangcc, line 21: Thanks for the comment! So I guess that Java(Clojure indeed) code is covered by some sort of Source Clear and does not need to be scanned here?

rexwangcc (approved):
LGTM, thank you @dinvlad !

.github/workflows/trivy.yml
dinvlad: Yep, exactly! It would be helpful to move the Clojure build instruction into the Dockerfile, to make it truly reproducible and so we don't have to use this workaround :-)

rexwangcc: Yea, historically it's outside of Dockerfile since the build has to clone several private github repos and it's not easy to do so within a Docker. We are close to get rid of the private images and once it's done we can move into the Dockerfile! (the last step is to move the config I mentioned to AppSec to WFL's own repo)

dinvlad: Sounds great!!

rexwangcc (approved):

PR #310, opened 2021-01-21
[GH-1130] Miscellaneous WFL configuration and build clean up 

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1130

### Changes

1. Github Actions of WFL don't have to clone private repos anymore. The SSH key hacks are removed now. (We need to remember to clean up the actual keys in their corresponding private repos though)
1. Remove a `touch target/wfl-api.jar` hack from the `trivy` Github Action.
1. WFL's Dockerfile now builds the WFL JAR for us, instead of trying to copy a built JAR from the disk into the image.
1. WFL's `build.clj` does not clone any repos anymore.
1. Remove the `clone` functionalities from `make` modules.
1. Remove the so-called "processing envs" (e.g. `wgs-dev`, `xx-prod`) from the `environments.clj`, refactor them into their own modules and create a map between Cromwell URL and their required default options/inputs
    - **Note: Intentionally** duplicate some of the definitions and functions such as `known-cromwells, inputs+options, cromwell->inputs+options, make-workflow-options` across the modules even though some of them are the same. This is to see if the modules can be self-encapsulated with their own `known cromwells`. I intend to keep this for PR review and abstract them before merging. (or just keep them even if this violates DRY). **I want to get folks' feedback on this**
    - Remove the `env -> cromwell -> env` transformations across the modules and stick with `cromwell URL` whenever and wherever passing around environments.
    - Refactor the whole `cromwell.clj` module, let the functions take `URL` instead of `environment`, as well as remove a bunch of un used functions.
1. Remove `/statuscounts` and `/workflows` endpoints and their corresponding handlers which were experimental endpoints when we started WFL and we never used them.
1. Remove `dx.clj` which is a `zero` legacy and we (not sure about tbl?) never used it.
1. Remove `metadata.clj` which is a `zero` legacy and we never used it.
1. Fix and update the unit and integration tests. Stick with `:debug` everywhere in the tests. (we used `:gotc-dev`, `xx-dev` and `wgs-dev` inconsistently. Now they are all `:debug`)
1. Fix a bug that the `postgres/batch-update-workflow-statuses!` always use "the first environment that matches the cromwell URL" to look for workloads. Reported by @tbl3rd . (This has been working fine simply because most of our processing such as wgs, xx, are sharing the exact same Cromwell).

### Left for Future Work

- Move the Executor (Cromwell URL) validation up from `start-workload` to `create-workload` for all modules.
- Let users pass in arbitrary Executor (Cromwell URL), even if it's not in `know-cromwells` list. (I'm not sure what's the best way to do it yet.)
- (Maybe?) Rename the functions that have`cromwell` to `executor` since we are going to Terra. For example, `known-cromwells` to `known-executors`.
- Technically this PR does not change how users interact with WFL. I will update the developer facing docs in follow up PRs.

### Review Instructions

- No instructions.


.github/workflows/trivy.yml
```diff
@@ -15,15 +15,6 @@ jobs:
...
       - uses: actions/checkout@v2
-
-      # API Dockerfile copies this, so it needs
-      # to exist for the build to succeed
-      # (even though it's not scanned)
-      - working-directory: ${{ matrix.context }}
-        run: |
-          mkdir target
-          touch target/wfl-api.jar
-
```
rexwangcc, line 26: @dinvlad as we discussed before.

dinvlad: LGTM!

api/src/wfl/module/aou.clj
```diff
@@ -47,17 +45,31 @@
...
-  "Array inputs for ENVIRONMENT that do not depend on the input file.
-  Environment here is just a default, input file may specify override."
-  [environment]
-  {:vault_token_path (get-in env/stuff [environment :vault_token_path])
-   :environment      (map-aou-environment environment)})
+(def ^:private inputs+options
+  [{:environment "dev"
+    :vault_token_path "gs://broad-dsp-gotc-arrays-dev-tokens/arrayswdl.token"}
+   {:environment "prod"
+    :vault_token_path "gs://broad-dsp-gotc-arrays-prod-tokens/arrayswdl.token"}])
```
ehigham, line 56: Are these the pipeline environments?
Do we really want to hard code these? Might there be more? Might they change?

api/src/wfl/module/aou.clj
```diff
@@ -47,17 +45,31 @@
    :disk_size                        100
    :preemptible_tries                3})
 
-(defn map-aou-environment
-  "Map AOU-ENV to environment for inputs preparation."
-  [aou-env]
-  ({:aou-dev "dev" :aou-prod "prod"} aou-env))
+(def ^:private known-cromwells
```
ehigham, line 48: I really wouldn't bother with this. If you really wanted to test if you were given a valid cromwell, maybe curl the `/engine/v1/version` endpoint and test it looks something like {"cromwell":"52-606efd2"} (i.e. `(contains? response-body :cromwell)`)

ehigham (approved):

api/src/wfl/module/aou.clj
ehigham: Actually this is probably a security issue... Forget I said anything.

api/src/wfl/module/aou.clj
rexwangcc: Sorry for the confusion, these are specific to `aou` module, as the Arrays pipeline takes an `environment` String as one of the inputs and use it to decide which database (one of dev, staging and prod) to write the cloud metrics during the pipeline execution. We had this block in `environments.clj` for so long and caused a lot of confusion and trouble, thus I moved it into the module itself to narrow the scope of what does `environment` really mean. This can still be overwritten by the incoming notifications as we have to support Mercury "staging" tests.

I was hoping after this PR, WFL will mostly pass (cromwell/executor) URLs across the codebase and stop using `environment` as a term to represent "processing envs" such as `wgs-dev` or `aou-prod` as they are hard-coded in the modules. The only `environment` WFL should care about maybe `debug` which supports local development then.

api/src/wfl/module/aou.clj
rexwangcc: Actually this is a good point for discussion and I really want to get your insights! 

I was talking to @tbl3rd about this and the only reason I used `known-cromwells` here is to prevent "security holes" that I don't know what they are … In fact, `known-cromwells` really prevents us from passing "arbitrary" cromwell/Terra/executor URLs to WFL in runtime, and I want to find a way to support that.

The way it is implemented in the PR is not very different from the `environments.clj`, each module maintains a `Cromwell URL -> default inputs & options` map, choosing what inputs & options to use based on the URLs (like we did for `env` keywords) or throw if unknown Cromwells are given. I'm thinking if we want to loosen the restrictions, when unknown Cromwells are given, some of the workflows, instead of throwing, WFL may just use a set of default inputs & options, such as the one maps to dev cromwell?

tbl3rd: Yeah. We don't want people to pass in arbitrary URLs for WFL to bash on.

api/src/wfl/module/aou.clj
tbl3rd: The "debug" environment was a hack. Would be nice to discard it, but by all means keep it if it solves some problem.

api/src/wfl/module/aou.clj
dinvlad: Just to clarify, are you talking about passing arbitrary Cromwell URLs from user request at runtime? In that case yes, please restrict it to a list of acceptable URLs only, as allowing arbitrary ones would open the service to a kind of "open redirect" attack.

tbl3rd: Yeah, so we're … not … gonna do that.

.github/workflows/trivy.yml
tbl3rd: Nice.

api/build/build.clj
```diff
@@ -33,8 +33,7 @@
         :commit           commit
         :committed        committed
         :built            built
-        :user             (or (System/getenv "USER") "wfl")
-        "pipeline-config" "3f182c0b06ee5f2dfebf15ed8b12d513027878ae"}
+        :user             (or (System/getenv "USER") "wfl")}
```
tbl3rd, line 36: Yay.

api/src/wfl/environments.clj
```diff
@@ -17,53 +13,34 @@
...
-   :server
-   {:project "development"
-    :vault   "path/to/server/secrets"}
-   :data-repo
-   {:url  "https://datarepo-development.example.com"
-    :service-account "datarepo-service-account.iam.gserviceaccount.com"}})
+(def gotc-dev-cromwell
+  {:labels cromwell-label-keys
+   :monitoring_script "gs://broad-gotc-prod-cromwell-monitoring/monitoring.sh"
+   :url "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org"})
```
tbl3rd, line 19: :labels shouldn't depend on environment.

api/src/wfl/environments.clj
```diff
@@ -17,53 +13,34 @@
...
+    :projects ["broad-gotc-dev"]}
    :server
-   {:project "production"
-    :vault   "path/to/server/secrets"}
-   :data-repo
-   {:url  "https://datarepo-production.example.com"
-    :service-account "datarepo-service-account.iam.gserviceaccount.com"}})
+   {:project "broad-gotc-dev"
+    :service-account "secret/dsde/gotc/dev/wfl/wfl-non-prod-service-account.json"
+    :vault "secret/dsde/gotc/dev/zero"}})
```
tbl3rd, line 33: Just dump :debug and we'll fix what breaks.

api/src/wfl/main.clj
```diff
@@ -41,9 +41,7 @@
   "Map command names to their run functions."
   (letfn [(varify [ns] (require ns) (ns-resolve ns 'run))
           (namify [ns] (last (str/split (name ns) #"\.")))]
-    (let [namespaces '[wfl.dx
-                       wfl.metadata
-                       wfl.server]]
+    (let [namespaces '[wfl.server]]
       (assoc (zipmap (map namify namespaces) (map varify namespaces))
              "help"         #'help
              "version"      #'version
```
tbl3rd, line 47: Also left over from command-line days.

api/src/wfl/module/aou.clj
tbl3rd: We could add those to the process environment.
(-In for a penny; in for a pound!-)

api/src/wfl/module/arrays.clj
```diff
@@ -39,11 +38,22 @@
...
-   :environment      ({:arrays-dev "dev" :arrays-prod "prod"} environment)})
+(def ^:private known-cromwells
+  ["https://cromwell-gotc-auth.gotc-dev.broadinstitute.org"
+   "https://cromwell-aou.gotc-prod.broadinstitute.org"])
+
+(def ^:private inputs+options
+  [{:environment "dev"
+    :vault_token_path "gs://broad-dsp-gotc-arrays-dev-tokens/arrayswdl.token"}
+   {:environment "prod"
+    :vault_token_path "gs://broad-dsp-gotc-arrays-prod-tokens/arrayswdl.token"}])
```
tbl3rd, line 49: Ugh.

api/src/wfl/module/aou.clj
```diff
@@ -47,17 +45,31 @@
...
+  ((zipmap known-cromwells inputs+options) (util/de-slashify url)))
+
+(defn ^:private is-known-cromwell-url?
+  [url]
+  (if-let [known-url (->> url
+                          util/de-slashify
+                          ((set known-cromwells)))]
+    known-url
+    (throw (ex-info "Unknown Cromwell URL provided."
+                    {:cromwell url}))))
```
tbl3rd, line 72: Or you could just add both variants to the set.

api/src/wfl/module/wgs.clj
```diff
@@ -51,11 +43,53 @@
     {:wgs_coverage_interval_list
      (str hg38 "wgs_coverage_regions.hg38.interval_list")}))
 
-(defn env-inputs
-  "Genome inputs for ENVIRONMENT that do not depend on the input file."
-  [environment]
+(def ^:private known-cromwells
+  ["https://cromwell-gotc-auth.gotc-dev.broadinstitute.org"
+   "https://cromwell-gotc-auth.gotc-prod.broadinstitute.org"])
```
tbl3rd, line 48: There are a lot of these.

api/src/wfl/util.clj
```diff
@@ -395,3 +361,19 @@ vault.client.http/http-client                               ; Keep :clint eastwo
...
+    (str url "/")))
+
+(defn de-slashify
+  "Ensure URL does not end in a slash /."
+  [url]
+  (if (str/ends-with? url "/")
+    (->> (seq url)
+         drop-last
+         (str/join ""))
+    url))
```
tbl3rd, line 379: Seems like what we want is some kind of pathname canonicalization function.

api/test/wfl/integration/modules/sg_test.clj
```diff
@@ -7,11 +7,14 @@
...
             [wfl.tools.fixtures :as fixtures]
             [wfl.tools.workloads :as workloads]
-            [wfl.util :as util :refer [absent? make-options]]
+            [wfl.util :as util :refer [absent?]]
             [wfl.module.sg :as sg])
   (:import (java.time OffsetDateTime)
            (java.util UUID)))
 
+(def ^:private cromwell-url-for-testing
+  "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org")
```
tbl3rd, line 16: Howzabout just `cromwell-url`?

api/test/wfl/unit/modules/aou_test.clj
```diff
@@ -90,15 +92,16 @@
...
-                                                                           {:control_sample_vcf_index_file "foo"
-                                                                            :control_sample_intervals_file "foo"
-                                                                            :control_sample_vcf_file       "foo"
-                                                                            :control_sample_name           "foo"})))))))))
+      (is (= all-expected-keys (set (keys (aou/make-inputs cromwell-url-for-testing
+                                                           (merge expected-per-sample-inputs
+                                                                  {:control_sample_vcf_index_file "foo"
+                                                                   :control_sample_intervals_file "foo"
+                                                                   :control_sample_vcf_file       "foo"
+                                                                   :control_sample_name           "foo"})))))))))
```
tbl3rd, line 107: I'll probably compulsively break these lines down.

tbl3rd (approved):
A good start …
Sad that metadata.clj is gone.

rexwangcc:
Update: most of the system tests succeeded except for Exomes and WGS workflows which failed due to:

> AccessDeniedException: 403 picard-dev@broad-gotc-dev.iam.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket.

`gsutil cat gs://broad-gotc-dev-cromwell-execution/ExternalExomeReprocessing/d3a81b6f-c2fd-47c5-99e1-f5e3cac20fef/call-CopyFilesFromCloudToCloud/attempt-2/stderr`

I'm not sure if that's what @tbl3rd and @rhiananthony ran into earlier.

api/test/wfl/integration/modules/sg_test.clj
rexwangcc: Sounds good, will update!

api/src/wfl/util.clj
rexwangcc: yup….

api/src/wfl/module/wgs.clj
rexwangcc: yea…. I was intentionally try to keep them in each module to make them self-sufficient, we can move them back to a centralized place if we find more patterns between them?

api/src/wfl/main.clj
rexwangcc: Should it be removed too then?

api/src/wfl/module/aou.clj
rexwangcc: Good idea! Will do.

api/src/wfl/module/aou.clj
rexwangcc: We could, a few string variables should be fine. Don't want to serialize/de-serialize a bunch of maps though from environment variables (process environment).

api/src/wfl/module/wgs.clj
tbl3rd: OK by me.

api/src/wfl/main.clj
tbl3rd: If you feel like it. I believe the PR is OK to merge. This can wait.

tbl3rd:
These are failing on the /main/ branch too.

api/test/wfl/system/v1_endpoint_test.clj
```diff
@@ -120,7 +120,7 @@
       (is (:created workload) "should have a created timestamp")
       (is (:started workload) "should have a started timestamp")
       (is (= (:email @endpoints/userinfo) (:creator workload)) "creator inferred from auth token")
-      (let [include [:pipeline :executor :project]]
+      (let [include [:pipeline :project]]
```
tbl3rd, line 123: Maybe they were both `nil`?

rexwangcc: When I ran the tests, the request didn't have `executor` field while the response got all three fields with not nil values. Apparently:
```
(deftest ^:parallel test-exec-wgs-workload-specifying-cromwell
  ;; All modules make use of the same code/behavior here, no need to spam Cromwell
  (test-exec-workload (-> (workloads/wgs-workload-request (UUID/randomUUID))
                          (rename-keys {:executor :cromwell}))))
```
is not a right test as it renames the request key.

rexwangcc:
Will start rebasing.

api/test/wfl/system/v1_endpoint_test.clj
rexwangcc: Ran it again, here is the error:
```clojure
FAIL in (test-exec-wgs-workload-specifying-cromwell) (v1_endpoint_test.clj:132)
calling api/v1/exec with ExternalWholeGenomeReprocessing workload request
expected: (= (included request) (included workload))
  actual: (not (= {:pipeline "ExternalWholeGenomeReprocessing", :project "(Test) rex/GH-1130-purge-envs-clj"} {:executor "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org", :pipeline "ExternalWholeGenomeReprocessing", :project "(Test) rex/GH-1130-purge-envs-clj"}))
```

rexwangcc:
Finished rebasing and resolving all merge conflicts. Unit and Integration tests are passing, system test is failing due to a picard-dev permission issue mentioned above. Merging this now.

PR #312, opened 2021-01-28
[GH-1153] Google Cloud Pub/Sub Support for Cloud Storage Notifications

ehigham:
### Purpose

One kind of workload might be to process new data as and when it's uploaded to an input bucket. We could use google cloud pub/sub notifications on the input bucket to listen for new files. In this PR, I'm exploring how wfl might:

- create pub/sub topics
- create notification configurations on buckets to publish to the newly created topic
- pull messages off a pub/sub queue after new data is uploaded to a bucket.

To do this, I've added a pubsub namespace to workflow-launcher and a test (+ test tools) for creating and managing Pub/Sub topics and publishing notifications from cloud storage buckets.
I've also included a commit for [GH-1158] as I refactored some of the test fixtures.

[GH-1153]: https://broadinstitute.atlassian.net/browse/GH-1153
[GH-1158]: https://broadinstitute.atlassian.net/browse/GH-1158

This code is work-in-progress towards a new "automatic" workload definition. It's reached a point where I think it could be checked in. Note that the tests are failing because of insufficient permissions for the service account `wfl-non-prod`.

Please review the pubsub namespace in particular. Let me know if the docstrings make sense along with any implementation suggestions.
Note that I have renamed the `gcs` namespace as `google.storage` to be consistent with the newly created `google.pubsub`.


api/src/wfl/util.clj
```diff
@@ -385,3 +385,17 @@ vault.client.http/http-client           ; Keep :clint eastwood quiet.
   [xs]
   (letfn [(between [[first second] x] (str first x second))]
     (between "()" (str/join "," (map (partial between "''") xs)))))
+
+(defn bracket
```
rexwangcc, line 389: dope!

api/src/wfl/service/google/pubsub.clj
```diff
@@ -0,0 +1,186 @@
...
+
+   Parameters
+   ----------
+   project  - Google Cloud Project to create the Pub/Sub topic in.
+   topic-id - unique identifier for the topic
+
+   Example
+   -------
+     (let [topic (create-topic \"broad-gotc-dev\" \"my-topic\")]
+       #_(;; use topic ))"
```
rexwangcc, line 31: like this docstring style, reminds me of the [numpy docs](https://numpydoc.readthedocs.io/en/latest/format.html).

api/src/wfl/service/google/pubsub.clj
```diff
@@ -0,0 +1,186 @@
...
+  [project topic-id]
+  (-> (pubsub-url (str/join "/" ["projects" project "topics" topic-id]))
+      (http/put {:headers (once/get-auth-header)})
```
rexwangcc, line 34: `(once/get-auth-header)` returns a header that's generated from a `scoped` Google Cloud Credential instance, I'm surprised that it works out of box for authenticating with pub/sub!

tbl3rd (approved):

rexwangcc (approved):
LGTM!

api/src/wfl/service/google/pubsub.clj
tbl3rd: To each his own.

api/src/wfl/util.clj
```diff
@@ -385,3 +385,17 @@ vault.client.http/http-client           ; Keep :clint eastwood quiet.
...
+   ----------
+   acquire - thunk returning newly acquired resource
+   release - function to clean up resource, called before this function returns
+   use     - function that uses the resource"
+  [acquire release use]
+  (let [resource (acquire)]
+    (try
+      (use resource)
+      (finally
+        (release resource)))))
```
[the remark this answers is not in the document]
ehigham: It's a design pattern: https://wiki.haskell.org/Bracket_pattern

api/src/wfl/service/google/pubsub.clj
ehigham: Can you suggest what I should do instead?

ehigham: For posterity (and future me), turns out https://www.googleapis.com/auth/cloud-platform is a super scope which, being super, encompasses pub/sub 

https://developers.google.com/identity/protocols/oauth2/scopes#pubsub

api/src/wfl/util.clj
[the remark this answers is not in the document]
tbl3rd: Oh in academia …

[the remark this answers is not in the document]
ehigham: What's nice is that `bracket` converts a statement into a function meaning it can be more easily composed.

PR #317, opened 2021-02-05
[GH-1197] Fix default inputs and options and update documentation about configurations.

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1197

### Changes

- Update the confusing default `inputs+options` in various modules -- **I forgot to clean them up in the pervious PR.**
- Move the Environment Variable loader into `once` and update the codebase to use it instead. **We only read a countable number of vars now so it works as documentation in code**. Please let me know what do you think about this!
- Remove the docs about Environment Variables cauz we don't need them.
- Add a note about permissions required for wgs and xx processing as promised to hOps.

### Review Instructions

- No instructions.


api/src/wfl/once.clj
```diff
@@ -1,19 +1,43 @@
...
+           ["COOKIE_SECRET"
+            "CROMWELL"
+            "GOOGLE_APPLICATION_CREDENTIALS"
+            "USER"
+            "VAULT_TOKEN"
+            "WFL_OAUTH2_CLIENT_ID"
+            "WFL_POSTGRES_URL"
+            "WFL_POSTGRES_USERNAME"
+            "WFL_POSTGRES_PASSWORD"
+            "WFL_DEPLOY_ENVIRONMENT"])))
```
rexwangcc, line 26: Hence we don't need to document them in markdown. Downside is this loses a bit of flexibility (but do we really need that kind of flexibility? i.e. load env vars that defined in runtime dynamically)

api/src/wfl/once.clj
```diff
@@ -1,19 +1,43 @@
...
+;; Only load the relevant environment variables
+;;
 (defonce the-system-environments
-  (reduce #(assoc %1 %2 (System/getenv %2)) {}
-          ["COOKIE_SECRET"
-           "CROMWELL"
-           "GOOGLE_APPLICATION_CREDENTIALS"
-           "WFL_OAUTH2_CLIENT_ID"
-           "WFL_DEPLOY_ENVIRONMENT"]))
+  (delay (reduce #(assoc %1 %2 (System/getenv %2)) {}
```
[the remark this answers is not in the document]
rexwangcc: `memoize` looks exactly like what I was looking for, TIL thanks!

> Maybe rename "environments.clj" to "environment.clj" and export the function get?
So we are going to scope `environment.clj` module to be the "envrionment variables" module from now on? I think that's a good idea! It may also make sense to rename `once` to something like `auth.clj`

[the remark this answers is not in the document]
ehigham: Maybe. I defer to you if you think its a good idea or not!
`auth` does seem like a better name than `once` for what it now contains.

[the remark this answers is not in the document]
tbl3rd: > rename `once` to something like `auth.clj`

Yes there is no `defoncé` left.

api/src/wfl/environment.clj
```diff
@@ -0,0 +1,74 @@
...
+           (log/debug "Perhaps run 'vault login' and try again")))))
+
+(def ^:private defaults
+  "Default values for environment variables, mainly for dev purposes."
+  {"GOOGLE_APPLICATION_CREDENTIALS"  (-> "secret/dsde/gotc/dev/wfl/wfl-non-prod-service-account.json"
+                                         vault-secrets
+                                         (json/write-str :escape-slash false)
+                                         .getBytes)
+   "TERRA_DATA_REPO_URL" "https://jade.datarepo-dev.broadinstitute.org/"
+   "WFL_OAUTH2_CLIENT_ID" (-> "secret/dsde/gotc/dev/zero" vault-secrets :oauth2_client_id)
```
ehigham, line 70: I didn't realise you needed to perform IO here. I don't think we should be doing this stuff when the namespace is being loaded. Perhaps hide this behind a thunk:
```clojure
(def ^:private defaults
  "Thunks for default values for environment variables, mainly for dev purposes."
  {"GOOGLE_APPLICATION_CREDENTIALS" 
   #(-> "secret/dsde/gotc/dev/wfl/wfl-non-prod-service-account.json"
        vault-secrets
        (json/write-str :escape-slash false)
        .getBytes)
   "TERRA_DATA_REPO_URL"
    (constantly "https://jade.datarepo-dev.broadinstitute.org/")
...
)
```
Then change `getenv` to apply that form:
```clojure
(def getenv (memoize (fn [name] (or (System/getenv name) ((defaults name))))))
```

tbl3rd (commented):
This is a good start.
You'll see I did something similar in the SG module when you rebase on /main/.
Let's talk about it next time you get a chance.
Thanks.

rexwangcc:
> This is a good start.
> You'll see I did something similar in the SG module when you rebase on /main/.
> Let's talk about it next time you get a chance.
> Thanks.

I took a look at your changes in `sg.clj`, really like how you handle `unknown cromwells` in `options-for-cromwell`. I'd like to change other modules accordingly in a follow up PR (this one is way too fat than I expected), if you don't mind?

api/src/wfl/environment.clj
rexwangcc: That's a really good point! I updated the code and ended up having something like `#(-> "42")` for consistency.

rexwangcc: Kinda have to list all of the used ENV VARs here in defaults otherwise we will get NPE. (good thing is listing all vars here indeed fulfills the original need of documenting everything together at one place )

rhiananthony (approved):
👍🏼

api/src/wfl/api/routes.clj
```diff
@@ -43,11 +42,11 @@
...
+           :handler   (fn [_] (handlers/succeed {:oauth2-client-id (env/getenv "WFL_OAUTH2_CLIENT_ID")}))
            :responses {200 {:body {:oauth2-client-id string?}}}
            :swagger   {:tags ["Informational"]}}}]
-   ["/api/v1/environments"
-    {:get  {:summary "Get all of the environments the server knows"
+   ["/api/v1/environment"
+    {:get  {:summary "Get all of the environment the server knows"
             :parameters nil
             :responses {200 {:body map?}}
             :handler (handlers/success env/stuff)}}]
```
tbl3rd, line 52: Will this expose `GOOGLE_APPLICATION_CREDENTIALS` and `WFL_COOKIE_SECRET`?
Maybe: `:summary "Show relevant variables deployed in the process environment."`

api/src/wfl/module/xx.clj
```diff
@@ -56,13 +56,11 @@
...
-                                                    :regulatory_designation
-                                                    :sample_name
-                                                    :version]
-                                :monitoring_script "gs://broad-gotc-prod-cromwell-monitoring/monitoring.sh"
-                                :url               "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org"}
+  [{:labels                    [:data_type
+                                :project
+                                :regulatory_designation
+                                :sample_name
+                                :version]
```
tbl3rd, line 63: Not sure anything still uses `:labels`, unfortunately.

tbl3rd (approved):
Just scanned most of this, and left a few comments.

api/src/wfl/api/routes.clj
rexwangcc: Actually this endpoint is not that useful anymore since it shows only the `debug` map which is simply just:

```
{:debug {:data-repo {:service-account "jade-k8-sa@broad-jade-dev.iam.gserviceaccount.com",
                     :url "https://jade.datarepo-dev.broadinstitute.org"},
         :google {:jes_roots ["gs://broad-gotc-dev-cromwell-execution"], :projects ["broad-gotc-dev"]},
         :server {:project "broad-gotc-dev",
                  :service-account "secret/dsde/gotc/dev/wfl/wfl-non-prod-service-account.json",
                  :vault "secret/dsde/gotc/dev/zero"},
         :doc "A local environment for development and debugging.",
         :keyword :debug,
         :name "debug"}}
```
I guess we could just take this endpoint out?

tbl3rd: OK by me.

rexwangcc: done. Later on if we need to look at the default options and inputs of a module, we could add an endpoint to expose them.

ehigham (approved):
looks good.

PR #319, opened 2021-02-08
[GH-1182] fix Pub/Sub test data race

ehigham:
The test naively assumed that no files would be updated between the point the notification configuration was added to the bucket and when the subscription was pulled.
To fix this, we can simply swap the order of events: make sure the queue is empty and then add the notification config.


rexwangcc (approved):
LGTM!

PR #321, opened 2021-02-10
[GH-1181] Ingest Workflow Outputs into TDR

ehigham:
### Purpose

https://broadinstitute.atlassian.net/browse/GH-1181

This change is a demonstrator for generic workflow output ingest into TDR. The main idea is:
- Use womtool to describe the outputs of the WDL. Using this description, build a type environment (a mapping from name -> type).
- When ingesting outputs, lookup the type of the output and dispatch on the type to a specific handler to perform any work needed to ingest that object 
  - e.g. ingest a file to get its TDR resource identifier
- Transform the result into some form that can be ingested as a row in a dataset table (`rename-gather`)
  - map output names onto dataset table names (rename)
  - collect together the outputs (gather)
- Upload result and have TDR ingest it.

The main file to look at is api/test/wfl/integration/datarepo_test.clj.
In the test, I've made use of thunks to delay work that can be performed later (e.g. polling). I've tried to add comments to clarify the intent. Let me know if you have questions.


api/test/wfl/tools/fixtures.clj
```diff
@@ -110,6 +92,49 @@
     (create-local-database name)
     (setup-local-database name)))
 
+(defmacro with-fixtures
```
ehigham, line 95: Not sure if this should be a macro or a fixture.
If I made it a function and the consuming lambda unary, it would itself be a fixture which might be kinda nice.

ehigham: Guess you can do this as a macro? 
```clojure
(fixtures/with-fixtures
     [(fixtures/with-fixtures 
        [(fixtures/with-fixtures 
           [(fixtures/with-fixtures [])])])]
    identity)
=> [[[[]]]]
```

tbl3rd (approved):
Thank you for upload-content.
Not sure how with-fixtures differs from use-fixtures,
but confident using it will bring understanding.

ehigham:
> Thank you for upload-content.
> Not sure how with-fixtures differs from use-fixtures,
> but confident using it will bring understanding.

The difference comes from what I've called a `fixture` and what clojure test calls a `fixture`. In clojure test, a `fixture` has the form [1]: 
```clojure
(defn my-test-fixture [f]
  (create-resource)
  (f)
  (destroy-resource))
```
And it's applied to the testing environment via `use-fixtures`:
```clojure
(use-fixtures my-test-fixture)
```
The thunk `f` passed to the `fixture` cannot have state passed into it. Thus `fixtures` in the clojure test sense have to modify a global variable.
I've introduced a different kind of fixture that allows values to be passed into the function. They can be applied within the test itself have the form: 
```
(def with-my-fixture [data use]
  (let [resource (acquire data)]
    (try
      (use resource)
      (finally
        (release resource)))))
```
`with-fixtures` is a combinator for using multiple of these fixtures without excess indentation.

[1] https://clojuredocs.org/clojure.test/use-fixtures

api/test/wfl/integration/datarepo_test.clj
```diff
@@ -3,59 +3,195 @@
...
+        return      (fn [x]   (constantly x))
+        bind        (fn [f g] (comp g f))
+        ingest-file (partial datarepo/ingest-file dataset-id profile-id)]
+    ((fn go [type value]
+       (case (:typeName type)
+         "File"
+         (let [[bkt obj] (gcs/parse-gs-url value)
+               target    (str/join "/" ["" workload-id obj])
+               mime      (mime-type/ext-mime-type value genomics-mime-types)]
+           (-> (get-in env/stuff [:debug :data-repo :service-account])
```
rexwangcc: Just fyi, I took the `stuff` and `:debug` out from the `environment.clj` so this needs to be something like `(env/getenv "WFL_DATA_REPO_SA")` now

rexwangcc (approved):
`with-fixtures` is really interesting, I can see `macro` makes sense here.

api/test/wfl/integration/datarepo_test.clj
ehigham: Thanks! Merged with your changes now!

PR #323, opened 2021-02-12
Updating AppVersion to the latest tag (0.5.1)

rhiananthony:
### Purpose
The latest tag of WFL is 0.5.1. However the AppVersion was set to 0.5.0. They should be the same

### Changes
Updating the AppVersion in the Chart.yaml file

### Review Instructions
Verify that this is the correct version to be on


ehigham (commented):
We generally keep the next released version in 'version'. Is there a way that we can manage this so that we don't have to manually bump the value in the yaml?

ehigham (approved):

PR #325, opened 2021-02-18
[GH-1181] Add Dispatch for "Object" Type

ehigham:
### Purpose
Extend the type-dispatching scheme to handle "Object" types.
- In doing this, I noticed we can eliminate some code duplication by passing in the outputs "type" as if it were an "Object".

I moved the mime-types into a new namespace. This was done for a couple of reasons
- Not expose the use of `ring` 
- Not merge ring's default mime-type set with our mime-types on every call to `ext-mime-type`
- Unit test: Include a test to verify that every file in an input/output has a mime type.

You'll notice in `mime-type-test` that the function `get-files` looks really similar to `ingest` in the `data-repo-test`. I would like to extract the logic of the inner "go" function into a re-usable function. I'm not sure how to do this so suggestions would be really welcome.


api/test/wfl/unit/mime_type_test.clj
```diff
@@ -0,0 +1,47 @@
...
+               [".cram" "application/octet-stream"]
+               [".vcf" "text/plain"]
+               [".gz" "application/gzip"]
+               [".fasta" "application/octet-stream"]
+               [".html" "text/html"]]]
+    (doseq [[filename expected] cases]
+      (testing filename
+        (is (= expected (mime-type/ext-mime-type filename)))))))
+
+(defn ^:private get-files
```
ehigham, line 19: This function has the same "shape" as `ingest` in `wfl.integration.data-repo-test`. It would be nice to extract this logic. Not clear how to do this yet.

rexwangcc (approved):
LGTM! I don't think I have sufficient knowledge to give any constructive suggestions to the question about monad.

ehigham:
> LGTM! I don't think I have sufficient knowledge to give any constructive suggestions to the question about monad.

Thanks. I'll merge this and if Tom has any thoughts I'll look forward to hearing them.
Don't worry about the monad comment, it's just one thought about how to do it!

PR #331, opened 2021-02-24
[GH-1172] Bulk Ingest Workflow Outputs

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1172
The `sarscov2_illumina_full` pipeline has many outputs, many of which are files. Ingesting them one at a time is really slow and so Jade recommend using the "bulk" insert endpoints.
In this change:
- Add bulk insert endpoint wrapper
- Gather all (unique) files in a workflow and bulk ingest them 
  - the same file can appear in more than one location in pipeline outputs
  - it's an error to try to ingest the same file twice.
  - We're limited to 1000 files per bulk ingest, Muscles says it is "probably fine" to pin ingest sizes to this, so the implementation uses batches of 1000. Probably should test if there's a sweet spot but it's good enough for now.
- Add mime-type support for new file extensions
  - There are some that are not determinable from the file extension, so in the test I've added a way to exclude those
- Refactor the type dispatch logic into a function `traverse` for re-use.

Note:
- In this change, I'm suggesting that we give TDR bucket reader access instead of reader access for each object in the outputs. This may be controversial.


api/src/wfl/mime_type.clj
```diff
@@ -1,17 +1,29 @@
...
+         "vcf"     "text/plain"}]
+    (merge mime-type/default-mime-types extensions)))
+
+;; visible for testing
+(defn ext-mime-type-no-default
+  "Look up the mime-type of the filename by file extension."
+  [filename]
+  (loop [filename (util/basename filename)]
+    (if-let [ext (util/extension filename)]
+      (or (mime-types ext) (recur (util/remove-extension filename))))))
```
rexwangcc, line 24: just curious about the usage of `loop recur` here, is it for some use cases like `(ext-mime-type-no-default "a.cram.unknown")`?

api/src/wfl/service/datarepo.clj
```diff
@@ -33,15 +35,38 @@
       util/response-body-json
       :id))
 
+;; While TDR does assign `loadTag`s, they're not always unique - submitting
+;; requests in parallel can cause bad things to happen. Use this to create a
+;; unique `loadTag` instead.
```
rexwangcc, line 40: This is really good to know!!

rexwangcc (commented):
Did one pass, looks good to me!

api/src/wfl/mime_type.clj
ehigham: I've see files like "biosample_attributes-merged.genbank.src" in the outputs. Now i have no idea what a `.src` file is, but I know what a `genbank` file is, so if I strip the .src extension off, I'll converge on genbank and return its mime-type by successively trying to find match extensions. Likewise, I have no idea what a .tmp file is, but if it has a sensible file extension before .tmp, it's likely that it's one of them (e.g. cram.tmp is likely a cram).
This might actually be the wrong thing to do.

ehigham: So yes.

rexwangcc (approved):

PR #333, opened 2021-02-26
Add workaround for GH-1212.

ehigham:
Not exactly sure what's causing this, or if it's a real issue at
all. It failed twice (once locally and then in a PR) but now I
can't reproduce it at all.
I suspect that it's being caused by querying the submission before
the workflows have been queued but I have no real evidence to
support this. If it is what I think, getting the submission again
might mitigate the problem.


rexwangcc (approved):
I saw it once in tests on `main` and re-running bypassed it. Not sure how that happens but this looks like a reasonable mitigation.

PR #334, opened 2021-02-26
[GH-1172] Start System Test for `sarscov2_illumina_all` Automation

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1172

This PR includes the start of an end-to-end system test for the `sarscov2_illumina_all`. The test aims to test creating, starting and stopping a workload to automate pipeline submission from and to the source and sink TDR datasets. We don't have much of this functionality yet so this test just sketches the expected behaviour.

In the test, I've expended on `rename-gather` as a means of converting pipeline names to dataset names (and vice versa). I'm currently using it to conform build a dataset ingest request from inputs. 
The mapping (`dataset-from-inputs.edn`) defines how to construct a row in the dataset table from the inputs of the pipeline. `rename-gather` is an interpreter for this mapping. There are issues which I'll solve another time, namely how to coerce between json and a map.

There's some code duplication in dataset ingest code as it's not clear how to group this yet or if it will need to be split up, say if we want to record what the dataset ingest job IDs were if WFL crashes.


api/test/wfl/tools/datasets.clj
```diff
@@ -0,0 +1,45 @@
+(ns wfl.tools.datasets
+  (:require [clojure.string :as str]
+            [clojure.data.json :as json]
+            [wfl.environment :as env]
+            [wfl.service.datarepo :as datarepo]
+            [wfl.service.google.storage :as storage])
+  (:import (java.util UUID)))
+
+(defn unique-dataset-request
```
ehigham, line 9: I think it's better that we don't adjust permissions in code like this. Better we require users to grant permissions than we grant them ourselves (likewise with all IAM-related endpoints)
- Add the TDR dev SA to be a reader of this bucket in gotc-deploy
- Figure our how TDR production works and do the same so we don't have to test permissions in production.

api/test/wfl/tools/datasets.clj
```diff
@@ -0,0 +1,45 @@
...
+  [dataset-basename]
+  (let [profile (env/getenv "WFL_TERRA_DATA_REPO_DEFAULT_PROFILE")]
+    (-> (str "test/resources/datasets/" dataset-basename)
+        slurp
+        json/read-str
+      ;; give it a unique name to avoid collisions with other tests
+        (update "name" #(str % (-> (UUID/randomUUID) (str/replace "-" ""))))
+        (update "defaultProfileId" (constantly profile)))))
+
+(defn ^:private ensure-tdr-can-read [bkt-obj-pairs]
```
ehigham, line 20: I think it's better that we don't adjust permissions in code like this. Better we require users to grant permissions than we grant them ourselves (likewise with all IAM-related endpoints)

Add the TDR dev SA to be a reader of this bucket in gotc-deploy
Figure our how TDR production works and do the same so we don't have to test permissions in production.

api/test/wfl/tools/datasets.clj
```diff
@@ -0,0 +1,51 @@
...
+(defn unique-dataset-request
+  "Create a dataset request for a uniquely-named dataset defined by the json
+   file `dataset-basename` with the specified `tdr-profile`."
+  [tdr-profile dataset-basename]
+  (-> (str "test/resources/datasets/" dataset-basename)
+      slurp
+      json/read-str
+      ;; give it a unique name to avoid collisions with other tests
+      (update "name" #(str % (-> (UUID/randomUUID) (str/replace "-" ""))))
+      (update "defaultProfileId" (constantly tdr-profile))))
```
tbl3rd, line 17: Nit: with the specified `tdr-profile` -=> with `tdr-profile`?
What would an UNspecified one be?
Move the interior `;;` comment outside the `defn`?

api/test/wfl/tools/datasets.clj
```diff
@@ -0,0 +1,51 @@
...
+            (let [[_ obj] (storage/parse-gs-url url)]
+              (str/join "/" ["" prefix obj])))
+          (ingest-batch [batch]
+            (->> (for [url batch] [url (target-name url)])
+                 (datarepo/bulk-ingest dataset-id tdr-profile)))]
+    (->> (split-at 1000 files)
+         (mapv ingest-batch)
+         (mapcat #(-> % datarepo/poll-job :loadFileResults))
+         (map #(mapv % [:sourcePath :fileId]))
+         (into {}))))
```
tbl3rd, line 38: Relate `tdr-profile` in the docstring?
Maybe: Ingest `files` for `tdr-profile` into a TDR …
`#(mapv % [:sourcePath :fileId])` looks like `juxt`.

api/test/wfl/tools/workflows.clj
```diff
@@ -40,6 +40,6 @@
      type object)))
 
 (defn get-files [type value]
-  "Return all values in `value` with WDL type `File`."
+  "Return all unique values in `value` with WDL type `File`."
   (letfn [(f [type object] (if (= "File" type) [object] []))]
-    (flatten (vals (traverse f type value)))))
+    (set (flatten (vals (traverse f type value))))))
```
tbl3rd, line 45: Return set of unique values in …

api/test/wfl/tools/datasets.clj
```diff
@@ -0,0 +1,51 @@
...
+  [values mapping]
+  (letfn [(literal? [x] (str/starts-with? x "$"))
+          (go! [v]
+            (cond (literal? v) (subs v 1 (count v))
+                  (string?  v) (get values (keyword v))
+                  (map?     v) (json/write-str (rename-gather values v)
+                                               :escape-slash false)
+                  (coll?    v) (remove nil? (map go! v))
+                  :else        (throw (ex-info "unknown type" {:type v}))))]
+    (into {} (for [[k v] mapping] [k (go! v)]))))
```
tbl3rd, line 51: Not sure what this does, but `(remove nil? (map go! v))` looks like `(keep go! v)`.

api/test/wfl/unit/datasets_test.clj
```diff
@@ -0,0 +1,18 @@
...
+           (datasets/rename-gather inputs {:workspace_name "$SARSCoV2-Illumina-Full"})))
+    (is (= {:instrument_model "Illumina NovaSeq 6000"}
+           (datasets/rename-gather inputs {:instrument_model "instrument_model"})))
+    (is (= {:extra ["broad_gcid-srv"]}
+           (datasets/rename-gather inputs {:extra ["package_genbank_ftp_submission.account_name"]})))
+    (is (= {:extra "{\"account_name\":\"broad_gcid-srv\",\"workspace_name\":\"SARSCoV2-Illumina-Full\"}"}
+           (datasets/rename-gather
+            inputs
+            {:extra {:account_name   "package_genbank_ftp_submission.account_name"
+                     :workspace_name "$SARSCoV2-Illumina-Full"}})))))
```
tbl3rd, line 18: Ah.  Hm … This processes an inputmap or outputmap.

tbl3rd (approved):

api/src/wfl/environment.clj
```diff
@@ -49,14 +47,22 @@
...
+   "WFL_TDR_DEFAULT_PROFILE"
+   #(-> "390e7a85-d47f-4531-b612-165fc977d3bd")
+   "WFL_TDR_SA"
+   #(-> "jade-k8-sa@broad-jade-dev.iam.gserviceaccount.com")
    "WFL_WFL_URL"
    #(-> "https://dev-wfl.gotc-dev.broadinstitute.org")})
 
-(def ^:private __getenv (memoize #(or (System/getenv %) ((defaults %)))))
+(def ^:private __getenv
+  (memoize #(or (System/getenv %) (when-let [init (defaults %)] (init)))))
```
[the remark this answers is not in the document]
ehigham: Instead of throwing a `NullPointerException`, yes.

api/test/wfl/tools/datasets.clj
ehigham: Moving the comment as the doc string explains this now.

api/test/wfl/tools/workflows.clj
ehigham: Thanks.

api/test/wfl/tools/datasets.clj
ehigham: Nice! ```clojure 
(keep go! v) ``` is definitely much nicer!

api/test/wfl/unit/datasets_test.clj
ehigham: both. Maybe it could just be called mapping, they're really the same thing. Improvements on nomenclature are always welcome :D

ehigham: I've just called it a mapping. The file is named `inputs-to-dataset` now.

api/test/wfl/tools/datasets.clj
ehigham: Thanks for `juxt`
I'm not going to stress too much with these docstrings. These functions are going to change quite a bit as the implementation comes together. I might even group the dataset-id with the profile-id as more-often than not, they're both required.

api/test/wfl/tools/datasets.clj
tbl3rd: Sorry about the environment.clj conflict, but resolution should be trivial.

PR #336, opened 2021-03-02
[GH-1203]: fix reading from vault at load-time

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1203
This was caused by server code reading `WFL_COOKIE_SECRET`.
Solution was to make the `cookie-store` and `app` functions
rather than a `def`ed value.

### DIFF
--- a/api/src/wfl/environment.clj
+++ b/api/src/wfl/environment.clj
@@ -63,8 +63,8 @@
   (memoize #(or (System/getenv %) (when-let [init (defaults %)] (init)))))
 
 (def testing
-  "Override the environment used by `getenv` for testing. DO NOT USE THIS.
-  Use `wfl.tools.fixtures/with-temporary-environment` instead."
+  "Override the environment used by `getenv` for testing. Use
+  `wfl.tools.fixtures/with-temporary-environment` instead of this."
   (atom {}))
 
 (defn getenv
--- a/api/src/wfl/environment.clj
+++ b/api/src/wfl/environment.clj
@@ -1,11 +1,12 @@
 (ns wfl.environment
   "Map environment to various values here."
-  (:require [clojure.data.json  :as json]
+  (:require [clojure.data.json :as json]
             [clojure.tools.logging :as log]
             [clojure.string :as str]
-            [vault.client.http]         ; vault.core needs this
-            [vault.core :as vault]
-            [wfl.util  :as util]))
+            [vault.client.http] ; vault.core needs this
+            [vault.core :as vault]))
+
+(declare getenv)
 
 ;; TODO: `is-known-cromwell-url?` in modules means new projects require new releases
 ;;  since this is baked in code. Can we improve this?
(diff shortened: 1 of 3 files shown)


api/src/wfl/environment.clj
```diff
@@ -1,28 +1,27 @@
...
+  (:require [clojure.data.json :as json]
             [clojure.tools.logging :as log]
             [clojure.string :as str]
-            [vault.client.http]         ; vault.core needs this
-            [vault.core :as vault]
-            [wfl.util  :as util]))
+            [vault.client.http] ; vault.core needs this
+            [vault.core :as vault]))
+
+(declare getenv)
```
tbl3rd, line 9: Please move vault-secrets instead.
And please remove the `is-known-cromwell-url?` TODO also.
And `DO NOT USE THIS`?

tbl3rd (approved):
I don't understand how this helps, but …

ehigham:
> I don't understand how this helps, but …

It helps because
- we're not performing IO when the namespace loads.
- failing to retrieve vault secrets is exceptional - we can't recover.
- all PRs should have been failing from this.

api/src/wfl/environment.clj
ehigham: vault secrets depends on getenv.

ehigham: And `DO NOT USE THIS` is so that people don't use `testing` to hack stuff.

tbl3rd:
> > I don't understand how this helps, but …
> 
> It helps because
> 
> * we're not performing IO when the namespace loads.
> * failing to retrieve vault secrets is exceptional - we can't recover.
> * all PRs should have been failing from this.

I understand the issue.  I even opened a bug about it and forwarded the link to you.
What I don't understand is how this PR can work, but I guess it does so I approve.

ehigham:
> > > I don't understand how this helps, but …
> > 
> > 
> > It helps because
> > 
> > * we're not performing IO when the namespace loads.
> > * failing to retrieve vault secrets is exceptional - we can't recover.
> > * all PRs should have been failing from this.
> 
> I understand the issue. I even opened a bug about it and forwarded the link to you.
> What I don't understand is how this PR can work, but I guess it does so I approve.

Oh. The GC root has changed to be via the future that's created by `start-webserver`

tbl3rd:
Alright.  I think I get it now.
Would still leave `(System/getenv "VAULT_TOKEN")`, but whatever.

api/src/wfl/environment.clj
tbl3rd: That's because System/getenv -=> wfl.environment/getenv which I wouldn't do, but whatever.

tbl3rd: > And `DO NOT USE THIS` is so that people don't use `testing` to hack stuff.

Then howzabout?

  "Override the environment used by `getenv` for testing. Use
  `wfl.tools.fixtures/with-temporary-environment` instead of this."

tbl3rd (approved):

ehigham:
> Alright. I think I get it now.
> Would still leave `(System/getenv "VAULT_TOKEN")`, but whatever.

`environment/getenv` is now debug logged. This was really helpful for identifying where the root was so I'm tempted to keep it as no doubt we'll run into this issue again.

api/src/wfl/environment.clj
ehigham: That's exactly the same this witj `DO NOT USE THIS` removed?

PR #337, opened 2021-03-03
[GH-1225] POST /api/v1/stop 

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1225
For COVID-19 Surveillance, we'll need a way of stopping the current processing (or workload) so that the pipeline can be upgraded or to swap in different datasets. This change introduces a "stop" route to stop accepting new workflows. Note this is different from abort where abort would abort all current processing. This may be added later as it serves a different use case.

For "batch" workloads, this doesn't really do much. For "streaming" workloads, this allows current processing to finish and stops the workload from accepting more work.

From this, I'll create a Jira to refactor the integration tests as they have been fairly copy and paste for a while.


api/src/wfl/util.clj
```diff
@@ -108,6 +108,9 @@
     (apply merge-with deep-merge a maps)
     (apply merge-with deep-merge maps)))
 
+(defn when->   [x p f] (if     (p x) (f x) x))
+(defn unless-> [x p f] (if-not (p x) (f x) x))
```
tbl3rd, line 112: These are not really flow macros like `->` or `cond->`.
OK, but pretty sure I do not like them.  (TIHI?)

api/test/wfl/integration/modules/aou_test.clj
```diff
@@ -46,21 +46,64 @@
...
+    (doto (workloads/create-workload! $)
+      (-> (contains? :created)  is)
+      (-> (absent?   :started)  is)
+      (-> (absent?   :stopped)  is)
+      (-> (absent?   :finished) is))
+    (doto (workloads/stop-workload! $)
+      (-> (contains? :created)  is)
+      (-> (contains? :started)  is)
+      (-> (contains? :stopped)  is)
+      (-> (contains? :finished) is))))
```
tbl3rd, line 104: Clever, Yoda.  Why `$` instead of `request`?
Why not a table of expected states?
(Maybe this is the expected refactoring you mentioned?)

api/test/wfl/integration/modules/arrays_test.clj
```diff
@@ -40,12 +40,49 @@
...
+    (doto (workloads/create-workload! $)
+      (-> (contains? :created)  is)
+      (-> (absent?   :started)  is)
+      (-> (absent?   :stopped)  is)
+      (-> (absent?   :finished) is))
+    (doto (workloads/stop-workload! $)
+      (-> (contains? :created)  is)
+      (-> (contains? :started)  is)
+      (-> (contains? :stopped)  is)
+      (-> (contains? :finished) is))))
```
tbl3rd, line 85: Similarly.

api/test/wfl/integration/modules/copyfile_test.clj
```diff
@@ -49,7 +44,7 @@
               (verify-workflow-options options)
               (is (= defaults (select-keys options (keys defaults))))
               (UUID/randomUUID)))]
-    (with-redefs-fn {#'submit-workflow verify-submitted-options}
+    (with-redefs-fn {#'cromwell/submit-workflow verify-submitted-options}
```
tbl3rd, line 47: Thanks, and so on.

api/test/wfl/integration/modules/copyfile_test.clj
```diff
@@ -94,3 +89,55 @@
...
+  [tx {:keys [workflows items] :as _workload}]
+  (letfn [(update! [{:keys [id]}]
+            (jdbc/update! tx items
+                          {:status "Succeeded" :updated (OffsetDateTime/now)}
+                          ["id = ?" id]))]
+    (run! update! workflows)))
+
+(deftest test-workload-state-transition
+  (with-redefs-fn
+    {#'cromwell/submit-workflow                 (constantly (UUID/randomUUID))
```
tbl3rd, line 103: Why `constantly`?

api/test/wfl/integration/modules/sg_test.clj
```diff
@@ -39,7 +42,7 @@
       "gs://broad-gotc-dev-storage/temp_references/gdc/dbsnp_144.hg38.vcf.gz.tbi"
       :input_cram
       "gs://broad-gotc-dev-wfl-sg-test-inputs/pipeline/G96830/NA12878/v23/NA12878.cram"}}]
-   :creator "@workloads/email"})        ; quoted to avoid log spam
+   :creator @workloads/email})
```
tbl3rd, line 45: Weird.  Thanks.

api/test/wfl/integration/modules/wgs_test.clj
```diff
@@ -58,7 +59,7 @@
                 :workflows)))))
 
 (deftest test-start-wgs-workload!
-  (with-redefs-fn {#'submit-workflows mock-submit-workflows}
+  (with-redefs-fn {#'cromwell/submit-workflows mock-submit-workflows}
```
tbl3rd, line 62: Thanks, and so on.

api/test/wfl/integration/modules/xx_test.clj
```diff
@@ -73,6 +69,7 @@
            workloads/create-workload!
            :workflows
            (run! (comp go! :inputs))))))
+
```
tbl3rd, line 72: Thanks, and so on.

api/src/wfl/module/batch.clj
```diff
@@ -63,10 +63,22 @@
...
-    (do
-      (postgres/batch-update-workflow-statuses! tx workload)
-      (postgres/update-workload-status! tx workload)
-      (workloads/load-workload-for-id tx id))))
+  (letfn [(patch-workload! [cols]
+            (jdbc/update! tx :workload cols ["id = ?" id]))
+          (stop! [{:keys [id] :as workload}]
+            (let [now (OffsetDateTime/now)]
+              (when-not (:started workload)
+                (patch-workload! {:started now :finished now}))
```
rexwangcc, line 81: interesting, so the rationale is that a stopped workload must be a "finished" one?

tbl3rd (approved):
Why does a workload need both `:stopped` and `:finished`?

rexwangcc (commented):
Did a quick pass, in general looks good.

ehigham:
> Why does a workload need both `:stopped` and `:finished`?

I think this stems from the difference between "stopping" a workload and "aborting" a workload.
When you "stop" a workload, you stop accepting new data but you allow any active workflows (i.e. ones still processing in cromwell) to complete. In that sense, the user has requested that the workload be "stopped" but it hasn't "finished" yet. The addition of `:stopped` is an implementation detail in this sense as a way of recording the user's request while allowing the update code to function (i.e. not `:finished`).

api/src/wfl/api/handlers.clj
```diff
@@ -56,36 +57,41 @@
...
-                (logr/infof "getting workloads by project %s" project)
-                (workloads/load-workloads-with-project tx project))
-              (do
-                (logr/infof "getting all workloads")
-                (workloads/load-workloads tx))))))))
+          (if-let [prj (:project query)]
+            (do (logr/infof "getting workloads by project %s" prj)
+                (workloads/load-workloads-with-project tx prj))
+            (do (logr/infof "getting all workloads")
+                (workloads/load-workloads tx)))))))))
```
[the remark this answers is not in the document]
ehigham: I guess not binding earlier comes from my hatred of `nil`. 
I'll bind earlier than use `cond` perhaps.

api/src/wfl/api/handlers.clj
```diff
@@ -56,36 +57,41 @@
...
+(defn stop-workload
+  "Stop a workload, allowing all active processing to complete."
+  [request]
+  (let [{uuid :uuid} (:body-params request)]
+    (logr/infof "stop-workload called workload:%s" uuid)
+    (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+      (->> (workloads/load-workload-for-uuid tx uuid)
+           (workloads/stop-workload! tx)
+           strip-internals
+           succeed))))
```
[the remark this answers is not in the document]
ehigham: So far the implementations of `stop-workload!` are idempotent which isn't true of `start-workload!` beyond this handler. Maybe it should. This change is already pretty large so maybe for another change?
Nice comment suggestion.

api/src/wfl/api/spec.clj
```diff
@@ -51,6 +51,7 @@
 (s/def ::workload-response (s/keys :opt-un [::finished
                                             ::input
                                             ::started
+                                            ::stopped
```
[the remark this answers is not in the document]
ehigham: Thanks.

api/test/wfl/integration/modules/copyfile_test.clj
ehigham: Take any number of inputs and return a UIUD. Ah but it'll be the same UUID. OK I'll fix.

api/src/wfl/module/aou.clj
```diff
@@ -267,22 +270,18 @@
...
-      (postgres/update-workflow-statuses! tx workload)
-      (workloads/load-workload-for-id tx (:id workload))
-      (catch Throwable cause
-        (throw (ex-info "Error updating aou workload"
-                        {:workload workload} cause))))))
+  (letfn [(update! [{:keys [id] :as workload}]
+            (postgres/update-workflow-statuses! tx workload)
+            (when (:stopped workload)
+              (postgres/update-workload-status! tx workload))
+            (workloads/load-workload-for-id tx id))]
```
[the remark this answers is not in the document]
ehigham: I don't think so. Wrapping the exception added noise. @tbl3rd was right I think.

api/src/wfl/module/batch.clj
ehigham: Nope. `:stopped` is a state somewhere between `:started` and `:finished`. If a workload wasn't `:started`, when you stop it, it's a no op to transition to `:finsihed` so I just did it here.
Also I'm not sure the update code that talks to terra/cromwell will handle :started` workloads that haven't actually had any workflows submitted.

tbl3rd:
> > Why does a workload need both `:stopped` and `:finished`?
> 
> I think this stems from "stopping" a workload and allowing any active workflows (i.e. ones still processing in cromwell) to complete. In that sense, the user has requested that the workload be "stopped" but it hasn't "finished" yet. The addition of `:stopped` is an implementation detail in this sense as a way of recording the user's request while allowing the update code to function (i.e. not `:finished`).

Oh, good point.  So :stopped means "Yo WFL, stop looking at these workflows".

api/src/wfl/module/aou.clj
```diff
@@ -267,22 +270,18 @@
...
+    (util/when-> workload #(and (:started %) (not (:finished %))) update!)))
```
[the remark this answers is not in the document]
ehigham: Let's expand: 
```clojure
(defn when-> [x p f] (if (p x) (f x) x))

(when-> workload #(and (:started %) (not (:finished %))) update!)))
((fn [x p f] (if (p x) (f x) x)) workload #(and (:started %) (not (:finished %))) update!)))
(if (#(and (:started %) (not (:finished %)))  workload) (update! workload) workload) 
(if (and (:started workload) (not (:finished workload)) (update! workload) workload) 
```

api/src/wfl/module/batch.clj
```diff
@@ -63,10 +63,22 @@
 
 (defn update-workload!
   "Use transaction TX to batch-update WORKLOAD statuses."
+  [tx workload]
+  (letfn [(update! [{:keys [id] :as workload}]
+            (postgres/batch-update-workflow-statuses! tx workload)
+            (postgres/update-workload-status! tx workload)
+            (workloads/load-workload-for-id tx id))]
+    (util/when-> workload #(and (:started %) (not (:finished %))) update!)))
```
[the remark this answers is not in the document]
ehigham: yep, exactly.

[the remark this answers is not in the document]
ehigham: I just can't be bothered with the default case of when the predicate isn't satisfied, hence `util/when->` and `util/assoc-when` (which is a specialisation i think)

api/src/wfl/module/batch.clj
```diff
@@ -63,10 +63,22 @@
...
-      (workloads/load-workload-for-id tx id))))
+  (letfn [(patch-workload! [cols]
+            (jdbc/update! tx :workload cols ["id = ?" id]))
+          (stop! [{:keys [id] :as workload}]
+            (let [now (OffsetDateTime/now)]
+              (when-not (:started workload)
+                (patch-workload! {:started now :finished now}))
+              (patch-workload! {:stopped now})
+              (workloads/load-workload-for-id tx id)))]
+    (util/unless-> workload #(or (:stopped %) (:finished %)) stop!)))
```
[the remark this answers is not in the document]
ehigham: because it isn't correct to set a workload to `:finished` if the workflows are currently processing - the update code won't run if the workload is `:finished`.

api/src/wfl/module/sg.clj
```diff
@@ -189,25 +189,27 @@
...
-  (let [{:keys [finished] :as result} (workloads/load-workload-for-id tx id)]
-    (when finished (register-workload-in-clio tx workload))
-    result))
+  [tx workload]
+  (letfn [(update! [{:keys [id] :as workload}]
+            (postgres/batch-update-workflow-statuses! tx workload)
+            (postgres/update-workload-status! tx workload)
+            (doto (workloads/load-workload-for-id tx id)
+              (register-workload-in-clio tx)))]
+    (util/when-> workload #(and (:started %) (not (:finished %))) update!)))
```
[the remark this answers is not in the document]
ehigham: Honestly, I find `when` more understandable in this case. "when started and not finished" makes more sense to me than `finished or not started`. Guess it's personal preference?
Also I suppose it's a style thing. I like combinators.

api/src/wfl/service/postgres.clj
```diff
@@ -90,11 +90,17 @@
...
   "Use `tx` to mark `workload` finished when all `workflows` are finished."
-  [tx {:keys [id items] :as _workload}]
-  (let [select "SELECT id FROM %%s WHERE status IS NULL OR status NOT IN %s"
-        query (format select (util/to-quoted-comma-separated-list finished?))]
-    (when (empty? (jdbc/query tx (format query items)))
-      (jdbc/update! tx :workload
-                    {:finished (OffsetDateTime/now)} ["id = ?" id]))))
+  [tx {:keys [id] :as workload}]
+  (when (empty? (active-workflows tx workload))
+    (jdbc/update! tx :workload {:finished (OffsetDateTime/now)} ["id = ?" id])))
```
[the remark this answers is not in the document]
ehigham: Why what?
I think naming what the sql query is trying to achieve is more readable than the sql query itself?

api/src/wfl/util.clj
ehigham: I completely agree, though not sure what TIHI means...
maybe `apply-when` or `return-unless`? IDK - suggestions welcome.

api/test/wfl/integration/modules/aou_test.clj
ehigham: Ah yeah. I personally don't like the yoda but not sure how to get around it. I used `$` as 
1. It's not always a request, it's also a workload. I'd like to eliminate the symbol altogether tbh
2. The clojure doc uses `$` so why not

ehigham: The refactoring is because it's in every module integ test. They're all fairly similar tbh which reflects the state of source code.

api/src/wfl/module/aou.clj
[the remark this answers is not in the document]
ehigham: > > > Why does a workload need both `:stopped` and `:finished`?
> > 
> > 
> > I think this stems from "stopping" a workload and allowing any active workflows (i.e. ones still processing in cromwell) to complete. In that sense, the user has requested that the workload be "stopped" but it hasn't "finished" yet. The addition of `:stopped` is an implementation detail in this sense as a way of recording the user's request while allowing the update code to function (i.e. not `:finished`).
> 
> Oh, good point. So :stopped means "Yo WFL, stop looking at these workflows".

Yeah exactly. I think it makes more sense for continuous delivery where you're like "stop looking for more data `workflow-launcher`"

ehigham:
> > > Why does a workload need both `:stopped` and `:finished`?
> > 
> > 
> > I think this stems from "stopping" a workload and allowing any active workflows (i.e. ones still processing in cromwell) to complete. In that sense, the user has requested that the workload be "stopped" but it hasn't "finished" yet. The addition of `:stopped` is an implementation detail in this sense as a way of recording the user's request while allowing the update code to function (i.e. not `:finished`).
> 
> Oh, good point. So :stopped means "Yo WFL, stop looking at these workflows".

Yeah exactly. I think it makes more sense for continuous delivery where you're like "stop looking for more data workflow-launcher"

PR #338, opened 2021-03-04
[GH-1214] Create TDR Snapshots

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1214

### Changes

- Add code/functions to create (by `row-id`s), list and delete TDR snapshots from TDR datasets and BigQuery.
- Add test code as well as fixtures to validate the snapshot by row-ids works.
- Add a create-snapshots in batches function that uses the same mechanism to create 500-sized batches of snapshots.

### Review Instructions

- No instructions.


ehigham (commented):
tests?

rexwangcc:
> tests?

unit tests is on their way in. Integrations test is currently living in the comment blocks until we can really create snapshots in TDR…

ehigham:
> > tests?
> 
> unit tests is on their way in. Integrations test is currently living in the comment blocks until we can really create snapshots in TDR…

I thought you were doing snapshot by row? In which case can we exclude the rows that are arrays of fileids for now?

api/test/wfl/integration/datarepo_test.clj
```diff
@@ -59,3 +61,28 @@
...
+                yesterday (util/datetime->str (util/n-day-from-now -1))
+                row-ids   (->> (datarepo/compose-snapshot-query dataset table yesterday today)
+                               (bigquery/query-sync dataProject)
+                               (bigquery/flatten-rows)
+                               ; the fixture does not support cleaning up multiple snapshots yet
+                               (take 1))]
+            (fixtures/with-temporary-snapshot
+              (snapshots/unique-snapshot-request tdr-profile dataset table row-ids)
+              #(let [snapshot (datarepo/snapshot %)]
+                 (is (= % (:id snapshot)))))))))))
```
rexwangcc, line 86: This test is failing at this moment since the parent temp dataset looks like:
```
{:id "9331fab1-02ca-41d7-aed2-13fbd9c08c40",
 :name "sarscov2_illumina_full_inputs4edf44bf4c5145c682715881dc71ba7f",
 :description "COVID-19 sarscov2_illumina_full pipeline inputs",
 :defaultProfileId "XXX",
 :dataProject "broad-jade-dev-data",
 :defaultSnapshotId nil,
 :schema
 {:tables
  [{:name "sarscov2_illumina_full_inputs",
    :columns
    [{:name "flowcell_tgz", :datatype "fileref", :array_of false}
     {:name "reference_fasta", :datatype "fileref", :array_of false}
     {:name "amplicon_bed_prefix", :datatype "string", :array_of false}
     {:name "biosample_attributes",
      :datatype "fileref",
      :array_of true}
     {:name "instrument_model", :datatype "string", :array_of false}
     {:name "sra_title", :datatype "string", :array_of false}
     {:name "min_genome_bases", :datatype "int64", :array_of false}
     {:name "max_vadr_alerts", :datatype "int64", :array_of false}
     {:name "workspace_name", :datatype "string", :array_of false}
     {:name "terra_project", :datatype "string", :array_of false}
     {:name "extra", :datatype "string", :array_of false}],
    :primaryKey
    ["flowcell_tgz"
     "reference_fasta"
     "amplicon_bed_prefix"
     "instrument_model"
     "sra_title"],
    :partitionMode "date",
    :datePartitionOptions {:column "datarepo_ingest_date"},
    :intPartitionOptions nil,
    :rowCount nil}],
  :relationships [],
  :assets []},
 :createdDate "2021-03-11T05:09:25.655987Z"}
```
as you can see there are no rows in the dataset, thus no snapshot can be made. 

How to ingest rows into a created dataset (or, does it even make sense to create dataset during testing, rather than using a pre-defined testing dataset that has rows) is a question I'd like to get insights from @ehigham

api/test/wfl/tools/snapshots.clj
```diff
@@ -0,0 +1,41 @@
...
+        table                table
+        today                (util/datetime->str (util/now))
+        yesterday            (util/datetime->str (util/n-day-from-now -1))
+        row-ids (->> (datarepo/compose-snapshot-query dataset table yesterday today)
+                     (bigquery/query-sync dataProject)
+                     (bigquery/flatten-rows))
+        unique-request-batch (partial unique-snapshot-request tdr-profile dataset table)]
+    (->> (partition-all 500 row-ids)
+         (map unique-request-batch)
+         (map datarepo/create-snapshot))))
```
rexwangcc, line 41: This is not being used at the moment, I prefer to leave it here as we will need it very soon in the next step of the work. It also demonstrates the limitation of the snapshotting mechanism we are using and how to bypass it.

api/test/wfl/integration/datarepo_test.clj
ehigham: For this, I wouldnt worry about creating a temp dataset. I'd create a temp snapshot of this https://jade.datarepo-dev.broadinstitute.org/datasets/28dbedad-ca6b-4a4a-bd9a-b351b5be3617/query

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +118,126 @@
...
+                     WHERE
+                        datarepo_ingest_date > '%s' AND datarepo_ingest_date <= '%s'"))))
+
+(defn ^:private legalize-tdr-columns
+  "Legalize TDR columns by stripping out columns that are Array[File] types, as
+   TDR does not support them yet."
+  [cols]
+  (-> cols
+      (->> (filter #(not (and (= (:datatype %) "fileref")
+                              (= (:array_of %) true)))))))
```
ehigham, line 213: The threading here gets in the way - I'd `letfn` your predicate:
```clojure
(letfn [(is-fileref-array? [{:keys [datatype array_of]}] 
          (and (= datatype "fileref") array_of))] 
  (remove is-fileref-array? cols))
```

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +118,126 @@
...
+(defn ^:private all-columns
+  "Parse out and legalize all column names for a `table` in TDR `dataset`."
+  [dataset table]
+  (-> dataset
+      (get-in [:schema :tables])
+      (->> (filter #(= (:name %) table)))
+      first
+      :columns
+      legalize-tdr-columns
+      (#(map :name %))))
```
ehigham, line 224: Why `->>` above and `#` here?

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +118,126 @@
...
+   Example
+   -------
+     (list-snapshots)
+     (list-snapshots \"48a51f71-6bab-483d-a270-3f9ebfb241cd\" \"85efdfea-52fb-4698-bee6-eef76104a7f4\")"
+  [& dataset-ids]
+  (letfn [(maybe [m k v] (if (seq v) (assoc m k {:datasetIds v}) m))]
+    (-> (http/get (repository "snapshots")
+                  (-> {:headers (auth/get-service-account-header)
+                       :query-params {:limit 999}}
+                      (maybe :query-params dataset-ids)))
```
ehigham, line 162: Again, I don't think the threading helps here, it's just adding parenthesis (in a lisp dialect? Heaven forfend!). Perhaps?
```clojure
(maybe
  {:headers      (auth/get-service-account-header)
   :query-params {:limit 999}}
  :query-params dataset-ids)
```

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +118,126 @@
...
+
+   Example
+   -------
+     (list-snapshots)
+     (list-snapshots \"48a51f71-6bab-483d-a270-3f9ebfb241cd\" \"85efdfea-52fb-4698-bee6-eef76104a7f4\")"
+  [& dataset-ids]
+  (letfn [(maybe [m k v] (if (seq v) (assoc m k {:datasetIds v}) m))]
+    (-> (http/get (repository "snapshots")
+                  (-> {:headers (auth/get-service-account-header)
+                       :query-params {:limit 999}}
```
ehigham, line 161: I wonder if there's a way to paginate this somehow so we don't have to set arbitrary lengths. I imagine that lots of snapshots will be created over the lifetime of any given project.

api/src/wfl/util.clj
```diff
@@ -377,3 +379,34 @@
...
+                       DateTimeFormatter/ofPattern
+                       (.withZone (ZoneId/from ZoneOffset/UTC)))]
+     (.format formatter datetime)))
+  ([^Instant datetime]
+   (datetime->str datetime "yyyy-MM-dd")))
+
+(defn now
+  "Return timestamp of now in ^Instant format."
+  []
+  (Instant/now))
```
ehigham, line 405: We use `OffsetDateTime/now` for our database stuff, presumably because https://stackoverflow.com/questions/58434148/what-is-better-to-persist-in-db-offsetdatetime-or-instant#:~:text=Instant%20is%20the%20simplest%2C%20simply,model%20data%20in%20simpler%20applications.
Maybe we should do the same here?

Given your usage in tests, getting the date from this instant may not be correct, especially if you're doing late night coding in the USA. See my comment below but I think we can remove much of this formatting stuff.

api/src/wfl/util.clj
```diff
@@ -377,3 +379,34 @@
...
+
+(defn n-day-from-now
```
ehigham, line 407: Maybe `days-from-now`?

api/test/wfl/integration/datarepo_test.clj
```diff
@@ -59,3 +61,28 @@
...
+  (let [tdr-profile (env/getenv "WFL_TDR_DEFAULT_PROFILE")
+        definition "sarscov2-illumina-full-inputs.json"]
+    (testing "creating snapshots from dataset"
+      (fixtures/with-temporary-dataset
+        (datasets/unique-dataset-request tdr-profile definition)
+        (fn [dataset-id]
+          (let [{:keys [dataProject] :as dataset} (datarepo/dataset dataset-id)
+                table     "sarscov2_illumina_full_inputs"
+                today     (util/datetime->str (util/now))
+                yesterday (util/datetime->str (util/n-day-from-now -1))
```
ehigham, line 77: Any reason why you're not using https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html? That will remove the need for the string formatting I think.

api/src/wfl/service/google/bigquery.clj
```diff
@@ -46,29 +46,35 @@
...
+  [project query]
+  (-> (str/join "/" ["projects" project "queries"])
+      bigquery-url
+      (http/post {:headers (auth/get-auth-header)
+                  :body    (json/write-str
+                            {:query          query
+                             :use_legacy_sql false})})
+      util/response-body-json))
+
+;; visible for testing
```
ehigham, line 67: lol

api/test/wfl/integration/datarepo_test.clj
```diff
@@ -59,3 +61,28 @@
...
+                row-ids   (->> (datarepo/compose-snapshot-query dataset table yesterday today)
+                               (bigquery/query-sync dataProject)
+                               (bigquery/flatten-rows)
+                               ; the fixture does not support cleaning up multiple snapshots yet
+                               (take 1))]
```
ehigham, line 82: Can't you snapshot all these rows at once? Why do you need multiple snapshots?

ehigham (commented):
Looks good! A few puzzles suggestions for tidy-ups. Thanks for adding the tests.
I'm not sure about the time/date formatting though.

api/test/wfl/integration/datarepo_test.clj
ehigham: If you want to use UTC, you can do this:
```clojure
(defn today-utc [] (java.time.LocalDate/now (java.time.Clock/systemUTC))
```

api/src/wfl/service/datarepo.clj
rexwangcc: The `->>` was for the `(filter #(= (:name %) table))` to work, thus later I have to `#`?

api/src/wfl/service/datarepo.clj
rexwangcc: Totally makes sense! I'm just a bit obsessed with the beautiful threadings recently 🤓

api/src/wfl/service/datarepo.clj
ehigham: Ah sorry! I was dumb.

api/src/wfl/service/datarepo.clj
rexwangcc: Let me look into it!

It's annoying that you have to at least hit the endpoint once to fetch the page total and page token in order to start the pagination, which makes a simple map solution more complex. I wasn't patient enough to dig into that as this PR ended up using this func.

api/src/wfl/service/datarepo.clj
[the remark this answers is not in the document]
rexwangcc: Hmm, good question, I struggled when adding this. I ended up having this simply because `all-columns` returns only column/field names (not `:datatype` or `:array_of`)that could be directly consumed by other functions. Putting the filter elsewhere will require this function to return more information than a vector of name strings, and the filter can be simply removed. (I was also expecting the bug to be fixed pretty soon. ) What's your thought? I'm OK with either way.

api/src/wfl/util.clj
rexwangcc: I see, I forgot we used `OffsetDateTime/now` for out DB stuff, and made an assumption we want everything in UTC. Will address this.

api/src/wfl/util.clj
rexwangcc: 👍 , naming is hard…

api/test/wfl/integration/datarepo_test.clj
rexwangcc: I was thinking we will probably to have to support `TIMESTAMP`s plus `DATE`s if there are requirements on how many samples to be processed within a day. (and we don't know how many times / # of samples will be ingested everyday yet...) so I thought making it support timestamps from now might be a good idea.

Plus I didn't realize we use `OffsetDateTime/now` for out DB stuff right now looking at

 

I chose the UTC one… I can swap with either `LocalDate` or `OffsetDateTime`, what do you think?

api/test/wfl/integration/datarepo_test.clj
rexwangcc: Hmm, I think this cline and comment might be outdated, at that moment I wasn't sure how many rows I would use (and there's a hard limit requires it to split at 500 batches), that piece was moved to the other module. I will update this.

api/src/wfl/service/datarepo.clj
rexwangcc: Based on offline discussion, will defer this work to the next PR that looks for snapshots. potential solutions could be using `lazy-seq`.

api/test/wfl/integration/datarepo_test.clj
rexwangcc: Based on offline discussion, will use `OffsetDateTime` for timestamps and `(str LocalDate)` for dates for date formatting. I will also remove the string formatting utilities as we don't use them.

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +117,116 @@
...
+  [{:keys [name dataProject] :as _dataset} table start end]
+  (let [dataset-name (str "datarepo_" name)
+        query "SELECT
+                 datarepo_row_id
+               FROM
+                 `%s.%s.%s`
+               WHERE
+                 datarepo_ingest_date > '%s' AND datarepo_ingest_date <= '%s'"]
+    (->> [dataProject dataset-name table start end]
+         (apply format query))))
```
ehigham, line 205: you don't need to break up the line here?

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +117,116 @@
...
+
+(defn all-columns
+  "Parse out information of all columns for a `table` in TDR `dataset`."
+  [dataset table]
+  (-> dataset
+      (get-in [:schema :tables])
+      (->> (filter #(= (:name %) table)))
+      first
+      :columns))
```
ehigham, line 214: Or just 
```clojure 
(->> (get-in dataset [:schema :tables])
     (filter #(= (:name %) table))
     first
     :columns)
```

api/src/wfl/service/google/bigquery.clj
```diff
@@ -46,29 +46,35 @@
...
+(defn parse-row
+  "Parse row information from BigQuery `rows` entry in query response."
+  [rows]
+  (map :v (:f rows)))
```
ehigham, line 71: Does this need to be public now? I think it's better to `flatten-rows` at the end of `query-sync` and hide the `:f` `:v` nonsense.

api/src/wfl/util.clj
```diff
@@ -377,3 +379,34 @@
...
+(defn datetime->str
+  "Parse `datetime` object into string, with optional `format`.
+
+   Note this function uses UTC as the time zone offset by default."
+  ([^Instant datetime format]
+   (let [formatter (-> format
+                       DateTimeFormatter/ofPattern)]
+     (.format formatter datetime)))
+  ([^Instant datetime]
+   (datetime->str datetime "yyyy-MM-dd")))
```
ehigham, line 399: You don't need this as we discussed? Just use `toString()` of `LocalDate`?

api/src/wfl/util.clj
```diff
@@ -377,3 +379,34 @@
...
+
+;; see https://github.com/broadinstitute/wfl/pull/338#discussion_r592618718
+(defn now
+  "Return timestamp of now in ^OffsetDateTime format."
+  []
+  (OffsetDateTime/now))
+
+(defn days-from-now
+  "Return the timestamp of n-days from now, either
```
ehigham, line 408: this doesn't return a timestamp

ehigham: At least, it's confusing because it doesn't return a `Timestamp`.

api/test/wfl/integration/google/bigquery_test.clj
```diff
@@ -33,12 +17,6 @@
     (is (every? #(= (:kind %) "bigquery#table") tables))))
 
 (deftest test-query-table-sync
-  (let [query-result (bigquery/query-table-sync google-project dr-dataset dr-view)]
+  (let [query (format "SELECT * FROM `%s.%s.%s`" google-project dr-dataset dr-view)
+        query-result (bigquery/query-sync google-project query)]
     (is (= "bigquery#queryResponse" (:kind query-result)))))
```
ehigham, line 22: Why are you testing this?

ehigham: Surely you'd want to make sure the data is what you expected, not that you got a big query response?

api/src/wfl/util.clj
ehigham:
```clojure
(format "%s" (java.time.LocalDate/now))
=> "2021-03-17"
```

api/src/wfl/util.clj
ehigham: Likewise with `now`

api/src/wfl/service/google/bigquery.clj
ehigham: I thought your "flatten-rows" just removes the `:f` and the `:v`. We'd probably always want to do that?

api/src/wfl/util.clj
ehigham: But you're not using it for that - you're using it to get the date, not an instant?

rexwangcc: AH, I think I over-complicated it. Lemme switch to LocalDate with UTC…

tbl3rd (commented):
I didn't run the tests.

api/src/wfl/util.clj
rexwangcc: Updated per offline discussion, use `LocalDate` as we don't need DateTime yet.

api/test/wfl/tools/snapshots.clj
```diff
@@ -0,0 +1,54 @@
...
+  (:require [clojure.string :as str]
+            [clojure.test :refer :all]
+            [wfl.service.google.bigquery :as bigquery]
+            [wfl.service.datarepo :as datarepo]
+            [wfl.util          :as util])
+  (:import (java.util UUID)))
+
+(defn ^:private legalize-tdr-columns
+  "Legalize TDR columns by stripping out columns that are Array[File] types, as
+   TDR does not support them yet."
```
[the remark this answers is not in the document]
rexwangcc: Unfortunately I cannot find a bug link. So putting a link to the slack thread which is the best thing I can get.

api/test/wfl/tools/snapshots.clj
```diff
@@ -0,0 +1,54 @@
...
+        table                table
+        today                (util/datetime->str (util/now))
+        yesterday            (util/datetime->str (util/days-from-now -1))
+        row-ids (->> (datarepo/compose-snapshot-query dataset table yesterday today)
+                     (bigquery/query-sync dataProject)
+                     (bigquery/flatten-rows))
+        unique-request-batch (partial unique-snapshot-request tdr-profile dataset table)]
+    (->> (partition-all 500 row-ids)
+         (map unique-request-batch)
+         (map datarepo/create-snapshot))))
```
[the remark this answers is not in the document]
rexwangcc: Thanks for the simplification suggestion! Looks much better, also the dateutils are simplified too.

api/test/wfl/unit/google/bigquery_test.clj
```diff
@@ -0,0 +1,38 @@
...
+               {:v "drs://jade.datarepo-dev.broadinstitute.org/v1_f1c765c6-5446-4aef-bdbe-c741ff09c27c_f2a7d885-4fd3-4faf-bd16-06219a8eef99"}
+               {:v "wfl-test-a830fe00-7ef2-430a-9d5e-fa0c18dc99e1/"}
+               {:v "drs://jade.datarepo-dev.broadinstitute.org/v1_f1c765c6-5446-4aef-bdbe-c741ff09c27c_2b67ed53-ccac-49c6-8ad6-8952a1dfaf98"}]}
+          {:f [{:v "8d529c08-bc21-4ea0-9254-d99b9c12dfd2"}
+               {:v "drs://jade.datarepo-dev.broadinstitute.org/v1_f1c765c6-5446-4aef-bdbe-c741ff09c27c_f2a7d885-4fd3-4faf-bd16-06219a8eef99"}
+               {:v "wfl-test-a830fe00-7ef2-430a-9d5e-fa0c18dc99e1/"}
+               {:v "drs://jade.datarepo-dev.broadinstitute.org/v1_f1c765c6-5446-4aef-bdbe-c741ff09c27c_2b67ed53-ccac-49c6-8ad6-8952a1dfaf98"}]}],
+   :totalBytesProcessed "221025",
+   :jobComplete true,
+   :cacheHit false})
```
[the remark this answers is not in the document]
rexwangcc: This is copied from one BQ response, so I guess `,` are from there. Will re-order to make it look better but not sure how rows are related. They're supposed to be more user -friendly. I updated this according to a change to the query-sync.

api/test/wfl/tools/snapshots.clj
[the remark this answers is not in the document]
ehigham: @rexwangcc https://broadworkbench.atlassian.net/browse/DR-1696?atlOrigin=eyJpIjoiOWUzOGI2NmI4MjNjNDk4ZWEyZWY4ZDhjOGM5NWU1NjUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +117,114 @@
...
+   -------
+     (list-snapshots)
+     (list-snapshots \"48a51f71-6bab-483d-a270-3f9ebfb241cd\" \"85efdfea-52fb-4698-bee6-eef76104a7f4\")"
+  [& dataset-ids]
+  (letfn [(maybe [m k v] (if (seq v) (assoc m k {:datasetIds v}) m))]
+    (-> (http/get (repository "snapshots")
+                  (maybe {:headers (auth/get-service-account-header)
+                          :query-params {:limit 999}}
+                         :query-params dataset-ids))
+        util/response-body-json)))
```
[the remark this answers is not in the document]
ehigham: I think it merges the `:query-params`

api/test/wfl/tools/snapshots.clj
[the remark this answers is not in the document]
rexwangcc: Oh thanks!

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +117,114 @@
...
+   https://jade.datarepo-dev.broadinstitute.org/swagger-ui.html#/"
+  [snapshot-request]
+  (-> (repository "snapshots")
+      (http/post {:headers      (auth/get-service-account-header)
+                  :content-type :application/json
+                  :form-params  snapshot-request})
+      util/response-body-json
+      :id
+      poll-job
+      :id))
```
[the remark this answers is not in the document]
rexwangcc: Sounds good! Will also add a link to https://cloud.google.com/bigquery/docs/reference/standard-sql/migrating-from-legacy-sql

api/src/wfl/service/datarepo.clj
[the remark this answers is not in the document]
rexwangcc: updated the name to `maybe-merge` to reduce confusion here.

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +117,114 @@
...
+   end        - The end date object in the timeframe to query inclusively."
+  [{:keys [name dataProject] :as _dataset} table start end]
+  (let [dataset-name (str "datarepo_" name)
+        query "SELECT
+                 datarepo_row_id
+               FROM
+                 `%s.%s.%s`
+               WHERE
+                 datarepo_ingest_date > '%s' AND datarepo_ingest_date <= '%s'"]
+    (->> [dataProject dataset-name table start end] (apply format query))))
```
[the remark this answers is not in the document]
rexwangcc: I sometimes favor `compose` or `forge` over `make` and `create`, idk why though 😅 Will update to `make` if it conveys better.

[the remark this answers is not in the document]
rexwangcc: The `(]` usage was not a standard but based on my decision that when you query, you probably want "Everything added since last date I checked but also include what's added today". Not sure if makes sense to you though.

api/src/wfl/service/datarepo.clj
```diff
@@ -117,3 +117,114 @@
...
+
+(defn all-columns
+  "Parse out information of all columns for a `table` in TDR `dataset`."
+  [dataset table]
+  (->> (get-in dataset [:schema :tables])
+       (filter #(= (:name %) table))
+       first
+       :columns))
```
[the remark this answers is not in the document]
rexwangcc: Updated, it `table`'s not there, this should return `nil`, e.g.

```clojure
(all-columns {:schema {:tables [{:a 1} {:b 2}]}} :c)
=> nil
```

api/src/wfl/service/google/bigquery.clj
```diff
@@ -46,29 +46,34 @@
...
+(defn ^:private parse-row
+  "Parse row information from BigQuery `rows` entry in query response."
+  [rows]
+  (map :v (:f rows)))
+
+(defn flatten-rows
+  "Given BigQuery `query-response`, return all flatten rows."
+  [query-response]
+  (->> (get-in query-response [:rows])
+       (mapcat parse-row)))
```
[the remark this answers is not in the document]
rexwangcc: yeah, @ehigham and I talked about this. Now the flatten is moved to the `query-sync` function.

Tbh, the response from BigQuery API is really nonsense and we need some level of abstraction, here in particular, the flatten we want is a shallow flatten not deep:

```clojure
;; we want
{:rows [(1 2 3) (a b c) (α β γ)]}
;; not we want
{:rows [(1 2 3 a b c α β γ)]}
```
so I replaced `mapcat` with `map` and left the further flatten work to the users by using `flatten`.

api/src/wfl/util.clj
```diff
@@ -377,3 +379,33 @@
...
+(defn now
+  "Return now in ^OffsetDateTime."
+  []
+  (OffsetDateTime/now))
+
+(defn days-from-now
+  "Return n-days from now in ^OffsetDateTime, either
+   backward or forward depends on the sign of n."
+  [^Integer n]
+  (.plus (now) n ChronoUnit/DAYS))
```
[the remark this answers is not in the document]
rexwangcc: changed this to `today` and `days-from-today`.

api/src/wfl/service/google/bigquery.clj
rexwangcc: done this

api/src/wfl/service/datarepo.clj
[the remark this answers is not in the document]
rexwangcc: done

api/test/wfl/unit/google/bigquery_test.clj
```diff
@@ -0,0 +1,38 @@
...
+              (count (get-in dr-view-content [:schema :fields])))
+           (count (bigquery/flatten-rows dr-view-content))))))
+
+(deftest test-dump-table->tsv
+  (testing "Dumping from BigQuery table response to TSV works"
+    (let [terra-table-name "test-name"
+          contents (-> (bigquery/dump-table->tsv dr-view-content "test-name")
+                       slurp
+                       (csv/read-csv :separator \tab))]
+      (is (= (format "entity:%s_id" terra-table-name) (first (first contents))) "The result TSV header is not properly formatted!"))))
```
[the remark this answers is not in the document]
rexwangcc: Updated `(comp first first)`.

Not sure if we really need to travel the other way around? I lean towards we don't add it until we need it?

ehigham (approved):

api/src/wfl/service/datarepo.clj
tbl3rd: Find examples of pagination in various namespaces.

PR #344, opened 2021-03-11
Changing cli.py to reflect deployment from helmfile rather than helm

rhiananthony:
### Purpose
This is for ticket GH-1254. It updates the cli.py file to use helmfile iinstead of helm for deployment.

### Changes
cli.py


ehigham (approved):

ehigham (approved):
Lovely.

PR #353, opened 2021-04-02
GH-1226 Support entity set creation when creating submissions.

okotsopoulos:
### Purpose
https://broadinstitute.atlassian.net/browse/GH-1226

To support submission creation for >1 entity, we must submit the entity references to FireCloud as an entity set (see [`TsvTypes.MEMBERSHIP`](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala#L13-L20
) in `firecloud-orchestration`).  The resulting entity set is itself a single entity which can be used in submission creation.

FireCloud's [membership type import code](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala#L140-L162) shows us that if we've specified the tsv type correctly, then we can specify one entity reference per line and the upload will group them by their id into lists.

For entity set imports, FireCloud expects a tsv like this (note that the samples don't need to match, in my example we were working with a single entity):
```
membership:sample_set_id\tsample
930360b9-cdc5-446e-b13a-a47b45d9ca4a\tNA12878
930360b9-cdc5-446e-b13a-a47b45d9ca4a\tNA12878
```

Result following entity set creation:
Screen Shot 2021-04-08 at 11 05 48 AM

When creating a submission for an entity set, note the following expectations for an entity set comprised of `sample` type entities.  I assume that that your method is meant to run on a single entity (not an entity set) and that the desired behavior when specifying an entity set during submission creation is that one workflow be kicked off for each entity in the set.
- Within the Terra UI, the method configuration's root entity type should be unchanged, i.e. remain set to `sample` and not changed to `sample_set`.
- The FireCloud submission creation request should also specify `expression=this.samples`, which will be evaluated to resolve entity references to the entities themselves.  Otherwise your submission will fail immediately.

### Changes

- Introduced `firecloud/create-submission-for-entity-set` to consolidate entities into a single entity before creating submission
- Now require `expression=this.<entity-type>s` specification in submission creation request when creating submissions with an entity set
- Refactored `bigquery/dump-file->tsv`, pulling out code now common with `firecloud` for Terra-compatible tsv generation
- Now require specification of tsv type when generating a Terra-compatible tsv (previously we only supported `:entity` tsvs)
- Added new integration and unit tests

### TODO
- Likely in a separate PR, refactor existing `firecloud/create-submission` to support multiple entity specification.
- Would like to be able to run integration tests for submission creation off of an entity set with >1 distinct entities, to confirm that multiple workflows are triggered (I observed this manually).

### Review Instructions

Entry point: `firecloud/create-submission-for-entity-set`.

- Successfully run new tests off of feature branch.
- Verify that BigQuery file dump to TSV behavior remains as expected.
- Verify that existing submission creation behavior (for a single entity) remains as expected.
- `firecloud/import-entity-set` should create a new entry in `<entity-type>_set` table.  Entity references should resolve as a list and indicate their source table.
- `firecloud/create-submission-for-entity-set` should create a single submission with one workflow triggered for each *distinct* member of the entity set.

### DIFF
--- a/api/src/wfl/service/firecloud.clj
+++ b/api/src/wfl/service/firecloud.clj
@@ -49,21 +49,21 @@
 ;; Third pass: consolidate to single method supporting 1+ entities.
 
 (defn create-submissions
-  "First pass:
-  Creates one submission in `workspace` for each element of `_entities` using the specified
-  `methodconfig`.
+  "
+  First pass:
+  Creates one submission in `workspace` for each of the specified `entities` using `methodconfig`.
+  Throws an `AssertionError` if no entities specified.
+  Returns a sequence of submission-ids for the newly-created submissions.
 
-  QUESTIONS:
-  Should we enforce entity count > 0?
-  Is `entity` a reserved word?
-  Should I hang tight for Ed's PR to be merged?
-  Does FireCloud support submission creation with multiple entities?
-
-  TODO:
-  Add specifications for inputs, usage examples"
-  [workspace methodconfig & _entities]
-  {:pre [(seq _entities)]} ; Checks for at least one entity specified.  Desired?
-  (for [_entity _entities] (create-submission workspace methodconfig _entity)))
+  Parameters
+  ----------
+  workspace    - Fully-qualified Terra Workspace in which the submissions will be created.
+  methodconfig - Fully-qualified method configuration.
+  entities     - [optional] Pairs of the form [Type Name]
+  "
+  [workspace methodconfig & entities]
+  {:pre [(seq entities)]}
+  (for [entity entities] (create-submission workspace methodconfig entity)))
 
 ;; OLIVIA END
 
--- a/api/src/wfl/service/firecloud.clj
+++ b/api/src/wfl/service/firecloud.clj
@@ -38,6 +38,29 @@
         util/response-body-json
         :submissionId)))
 
+;; OLIVIA START
+;; Second pass: creates one submission in a workspace encompassing all specified entities (1+).
+;; Third pass: consolidate to single method supporting 1+ entities.
+
+(defn create-submissions
+  "First pass:
+  Creates one submission in `workspace` for each element of `_entities` using the specified
+  `methodconfig`.
+
+  QUESTIONS:
+  Should we enforce entity count > 0?
+  Is `entity` a reserved word?
+  Should I hang tight for Ed's PR to be merged?
+  Does FireCloud support submission creation with multiple entities?
+
+  TODO:
+  Add specifications for inputs, usage examples"
+  [workspace methodconfig & _entities]
+  {:pre [(seq _entities)]} ; Checks for at least one entity specified.  Desired?
+  (for [_entity _entities] (create-submission workspace methodconfig _entity)))
+
+;; OLIVIA END
+
 (defn get-submission
   "Return the submission in the Terra `workspace` with `submission-id`."
   [workspace submission-id]
--- a/api/src/wfl/service/firecloud.clj
+++ b/api/src/wfl/service/firecloud.clj
@@ -5,7 +5,8 @@
             [clojure.string :as str]
             [wfl.auth :as auth]
             [wfl.environment :as env]
-            [wfl.util :as util]))
+            [wfl.util :as util])
+  (:import [java.util UUID]))
 
 (defn ^:private firecloud-url [& parts]
   (let [url (util/de-slashify (env/getenv "WFL_FIRECLOUD_URL"))]
@@ -43,29 +44,6 @@
         http/request
         util/response-body-json
         :submissionId)))
-
-;; OLIVIA START
-;; Second pass: creates one submission in a workspace encompassing all specified entities (1+).
-;; Third pass: consolidate to single method supporting 1+ entities.
-
-(defn create-submissions
-  "
-  First pass:
-  Creates one submission in `workspace` for each of the specified `entities` using `methodconfig`.
-  Throws an `AssertionError` if no entities specified.
-  Returns a sequence of submission-ids for the newly-created submissions.
-
-  Parameters
-  ----------
-  workspace    - Fully-qualified Terra Workspace in which the submissions will be created.
-  methodconfig - Fully-qualified method configuration.
-  entities     - [optional] Pairs of the form [Type Name]
-  "
-  [workspace methodconfig & entities]
-  {:pre [(seq entities)]}
-  (for [entity entities] (create-submission workspace methodconfig entity)))
-
-;; OLIVIA END
 
 (defn create-workspace
   "Create an empty Terra workspace with the fully-qualified `workspace` name,
@@ -148,6 +126,81 @@
                   :multipart (util/multipart-body
                               {:Content/type "text/tab-separated-values"
                                :entities     (slurp file)})})))
+
+(defn import-entity-set
+  "
+  Formats existing sample ENTITIES into a Terra-compatible tsv
+  and uploads to WORKSPACE as an entity set entry identified by SET-NAME.
+
+  All entity references should correspond to existing entities.
+  All entity references should be of the same type.
+
+  Parameters
+  ----------
+    workspace - Fully-qualified Terra Workspace in which the entity set
+                will be created
+    set-name  - Unique identifier for the entity set
+    entities  - List of entity [Type Name] pairs
+
+  Example
+  -------
+    (import-entity-set \"workspace-namespace/workspace-name\"
+                       \"c25ee04b-...\"
+                       [[\"sample\" \"NA12878\"] [\"sample\" \"NA12879\"]])
+  "
+  [workspace set-name [[entity-type _] & _ :as entities]]
+  ;; TODO: Should we enforce common entity-type for all entities?
+  (let [columns [entity-type entity-type]
+        rows (for [[_ entity-name] entities] [set-name entity-name])]
+    (->> (util/columns-rows->terra-tsv :membership columns rows)
+         .getBytes
+         (import-entities workspace))))
+
+(defn consolidate-entities-to-set
+  "
+  Optionally consolidates ENTITIES into an entity set.
+  Returns the [Type Name] pair for the resulting entity,
+  or the passed-in entity if it was the only one specified.
+
+  Parameters
+  ----------
+    workspace - Fully-qualified Terra Workspace in which the entity set
+                would be created
+    entities  - List of entity [Type Name] pairs
+
+  Example
+  -------
+    (consolidate-entities-to-set \"workspace-namespace/workspace-name\"
+                                 [[\"sample\" \"NA12878\"] [\"sample\" \"NA12879\"]])
+  "
+  [workspace [entity & maybe-entities :as entities]]
+  ;; TODO: Should we enforce common entity-type for all entities?
+  (if (seq maybe-entities)
+    (let [id (str (UUID/randomUUID)) ; TODO: do we have access to a workload-affiliated UUID?
+          response (import-entity-set workspace id entities)]
+      [(:body response) id])
+    entity))
+
+(defn create-submission-for-entity-set
+  "
+  Optionally consolidates ENTITIES into an entity set and uses it to create a
+  submission in WORKSPACE with METHODCONFIG.
+
+  At least one entity should be specified.
+
+  Parameters
+  ----------
+    workspace    - Fully-qualified Terra Workspace in which the submission
+                   will be created.
+    methodconfig - Fully-qualified method configuration.
+    entities     - [Optional] Entity [Type Name] pairs
+  "
+  [workspace methodconfig & entities]
+  ;; TODO: refactor create-submission to support entity set specification.
+  ;; TODO: instead of variadic approach, should we take in a list of entities?
+  {:pre [(seq entities)]}
+  (->> (consolidate-entities-to-set workspace entities)
+       (create-submission workspace methodconfig)))
 
 (defn list-entities
   "List all entities with `entity-type` in `workspace`."
--- a/api/src/wfl/service/firecloud.clj
+++ b/api/src/wfl/service/firecloud.clj
@@ -28,22 +28,26 @@
 
 (defn create-submission
   "Submit samples in a workspace for analysis with a method configuration in Terra."
-  [workspace methodconfig [entity-type entity-name :as _entity]]
-  (let [[mcns mcn] (str/split methodconfig #"/")]
-    (-> {:method       :post
-         :url          (workspace-api-url workspace "submissions")
-         :headers      (auth/get-auth-header)
-         :content-type :application/json
-         :body         (json/write-str
-                        {:methodConfigurationNamespace mcns
-                         :methodConfigurationName mcn
-                         :entityType entity-type
-                         :entityName entity-name
-                         :useCallCache true}
-                        :escape-slash false)}
-        http/request
-        util/response-body-json
-        :submissionId)))
+  ([workspace methodconfig [entity-type entity-name :as _entity] body-override]
+   {:pre [map? body-override]}
+   (let [[mcns mcn] (str/split methodconfig #"/")]
+     (-> {:method       :post
+          :url          (workspace-api-url workspace "submissions")
+          :headers      (auth/get-auth-header)
+          :content-type :application/json
+          :body         (json/write-str
+                         (util/deep-merge {:methodConfigurationNamespace mcns
+                                           :methodConfigurationName mcn
+                                           :entityType entity-type
+                                           :entityName entity-name
+                                           :useCallCache true}
+                                          body-override)
+                         :escape-slash false)}
+         http/request
+         util/response-body-json
+         :submissionId)))
+  ([workspace methodconfig entity]
+   (create-submission workspace methodconfig entity {})))
 
 (defn create-workspace
   "Create an empty Terra workspace with the fully-qualified `workspace` name,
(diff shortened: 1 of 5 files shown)


api/src/wfl/util.clj
```diff
@@ -432,3 +434,81 @@
    (poll task seconds 3))
   ([task]
    (poll task 1)))
+
+(defn columns-rows->tsv
```
okotsopoulos, line 438: Of the new code I've added here, this is the only method I feel confident should live in `util`.

api/src/wfl/util.clj
```diff
@@ -432,3 +434,81 @@
...
+     (csv/write-csv writer columns :separator \tab)
+     (csv/write-csv writer rows :separator \tab)
+     file))
+  ([columns rows]
+   (str (columns-rows->tsv columns rows (StringWriter.)))))
+
+"A set of all mutually supported TSV file types (by FireCloud and WFL)"
+(s/def ::tsv-type #{:entity :membership})
+
+(defn terra-id
```
okotsopoulos, line 458: This feels specific enough to be moved somewhere other than `util` but is code that is used in both `bigquery` and `firecloud`.

api/src/wfl/util.clj
```diff
@@ -432,3 +434,81 @@
...
+    (terra_id :membership \"flowcell\")
+  "
+  [tsv-type col]
+  {:pre [(s/valid? ::tsv-type tsv-type)]}
+  (let [stripped-col (-> (unsuffix col "_id")
+                         (unsuffix "_set"))
+        maybe-set (if (= :membership tsv-type) "_set" "")]
+    (format "%s:%s%s_id" (name tsv-type) stripped-col maybe-set)))
+
+(defn columns-rows->terra-tsv
```
okotsopoulos, line 491: Same here (seems too specific for `util`).

api/src/wfl/service/firecloud.clj
```diff
@@ -126,6 +127,81 @@
...
+
+  Example
+  -------
+    (consolidate-entities-to-set \"workspace-namespace/workspace-name\"
+                                 [[\"sample\" \"NA12878\"] [\"sample\" \"NA12879\"]])
+  "
+  [workspace [entity & maybe-entities :as entities]]
+  ;; TODO: Should we enforce common entity-type for all entities?
+  (if (seq maybe-entities)
+    (let [id (str (UUID/randomUUID)) ; TODO: do we have access to a workload-affiliated UUID?
```
okotsopoulos, line 179: Flagging this: currently generating a random UUID to serve as a unique identifier for a new entity set entry.  While this works it doesn't reveal much about when / how the entity set entry was uploaded.  Is there some other identifier we can use, perhaps a UUID affiliated with the workload?

api/src/wfl/service/firecloud.clj
```diff
@@ -126,6 +127,81 @@
...
+  Parameters
+  ----------
+    workspace    - Fully-qualified Terra Workspace in which the submission
+                   will be created.
+    methodconfig - Fully-qualified method configuration.
+    entities     - [Optional] Entity [Type Name] pairs
+  "
+  [workspace methodconfig & entities]
+  ;; TODO: refactor create-submission to support entity set specification.
+  ;; TODO: instead of variadic approach, should we take in a list of entities?
```
okotsopoulos, line 200: Flagging this: I wasn't sure how this method will be called.  The variadic declaration allows for backwards compatibility when integrating with existing `create-submission`.  But I also thought it was possible this would instead be easier to call with a list of entities.

api/src/wfl/util.clj
```diff
@@ -432,3 +434,81 @@
...
+    entity:[entity-type]_id
+    membership:[entity-type]_set_id
+
+  For tsv-type :entity, 'entity:sample_id' will upload the .tsv data into a `sample` table
+  in the workspace (or create one if it does not exist).
+  If the table already contains a sample with that id, it will get overwritten.
+
+  For tsv-type :membership, 'membership:sample_set_id' will append sample names to a `sample_set` table
+  in the workspace (or create one if it does not exist).
+  If the table already contains a sample set with that id, it will be appended to and not overwritten.
```
okotsopoulos, line 472: Flagging this: for entity set uploads (`tsv-type :membership`) the FireCloud behavior for existing ids doesn't match that for entity uploads (`tsv-type :entity`).

api/src/wfl/service/firecloud.clj
```diff
@@ -126,6 +127,81 @@
                               {:Content/type "text/tab-separated-values"
                                :entities     (slurp file)})})))
 
+(defn import-entity-set
+  "
+  Formats existing sample ENTITIES into a Terra-compatible tsv
+  and uploads to WORKSPACE as an entity set entry identified by SET-NAME.
+
+  All entity references should correspond to existing entities.
+  All entity references should be of the same type.
```
okotsopoulos, line 136: This was an assumption I made but wanted to verify.  If so, I can also enforce this via a precondition.

api/src/wfl/service/firecloud.clj
```diff
@@ -126,6 +127,81 @@
...
+
+  Parameters
+  ----------
+    workspace - Fully-qualified Terra Workspace in which the entity set
+                will be created
+    set-name  - Unique identifier for the entity set
```
ehigham, line 142: I don't think this has to be unique right? I guess the entity will be updated if it already exists?

api/src/wfl/service/firecloud.clj
```diff
@@ -126,6 +127,81 @@
...
+    entities  - List of entity [Type Name] pairs
+
+  Example
+  -------
+    (consolidate-entities-to-set \"workspace-namespace/workspace-name\"
+                                 [[\"sample\" \"NA12878\"] [\"sample\" \"NA12879\"]])
+  "
+  [workspace [entity & maybe-entities :as entities]]
+  ;; TODO: Should we enforce common entity-type for all entities?
+  (if (seq maybe-entities)
```
ehigham, line 178: I think you can just always create the entity-set, rather than conditionally. There's no harm in having a singleton set I think.

api/src/wfl/service/firecloud.clj
ehigham: Hmm - I think we only need one endpoint wrapper for submissions since there is only one. That said, this code is definitely useful and will probably live in the `covid` namespace @rexwangcc is working on. I'd put it in the test as a demonstrator for HOW you might bundle a collection of entities into one for submission.

ehigham: As a separate comment, I'd also add the name of the entity you want to create as an input because we'd probably want to append the set to an existing table in the workspace.

ehigham: And yes, I'd take a list of entities!

api/src/wfl/service/google/bigquery.clj
```diff
@@ -1,13 +1,11 @@
...
-            [clojure.data.csv :as csv]
-            [clojure.java.io :as io]
+  (:require [clj-http.client   :as http]
             [clojure.data.json :as json]
-            [clojure.string :as str]
-            [wfl.auth :as auth]
-            [wfl.util         :as util])
+            [clojure.string    :as str]
+            [wfl.auth          :as auth]
+            [wfl.util          :as util])
```
ehigham, line 8: sad :'(

ehigham (commented):
Looks good. Well done. A few comments!

api/src/wfl/service/firecloud.clj
okotsopoulos: Ah, the behavior when uploading with an existing identifier is that the members will be *appended* to the existing entity, and not overwritten as for `:entity` tsv uploads.  This felt undesirable to me and like something we'd wish to avoid.

https://github.com/broadinstitute/wfl/pull/353/files/5e8c027c361415cf372907d951a0eb9334e89a40#r609784639

When I ran the same entity set upload twice for the same id, the set size doubled from 2 -> 4 rather than being overwritten by the new upload.

api/src/wfl/service/firecloud.clj
ehigham: I don't think you need to enforce this - firecloud will probably error if you try to do this!

api/src/wfl/service/firecloud.clj
okotsopoulos: All noted.  Yes, my ultimate intention in this PR is to consolidate to a single endpoint, and now that I know how it will be called that helps!

api/src/wfl/util.clj
```diff
@@ -432,3 +434,81 @@
...
+
+  Examples
+  --------
+    (terra_id :entity \"flowcell\")
+    (terra_id :membership \"flowcell\")
+  "
+  [tsv-type col]
+  {:pre [(s/valid? ::tsv-type tsv-type)]}
+  (let [stripped-col (-> (unsuffix col "_id")
+                         (unsuffix "_set"))
```
ehigham, line 487: Why do you need to strip these suffixes? a set of ids seems completely reasonable to me.

api/src/wfl/util.clj
```diff
@@ -432,3 +434,81 @@
...
+        maybe-set (if (= :membership tsv-type) "_set" "")]
+    (format "%s:%s%s_id" (name tsv-type) stripped-col maybe-set)))
```
ehigham, line 489: You could do this in one line with :D
```clojure
(str column (when (= :membership tsv-type) "_set") "_id")
```

api/src/wfl/util.clj
ehigham: It's OK - `util` is quite vast these days. We could break it down at some point.

api/src/wfl/util.clj
okotsopoulos: Good question!  We ultimately do want the `_id` suffix everywhere, for example, but the value is passed in inconsistently.  Otherwise I'd need to check if it already had the `_id` suffix before adding.  Seemed cleaner and more intentional to start with a blank slate first.

api/src/wfl/util.clj
okotsopoulos: Rad.  Thanks!

api/src/wfl/service/firecloud.clj
tbl3rd: Just a suggestion …

```clojure
(defn import-entity-set
  "Format sample ENTITIES into a TSV, and upload it to WORKSPACE as an
  entity set named SET-NAME.  WORKSPACE is a \"namespace/name\" string.
  ENTITIES must share the same type.

  (import-entity-set \"workspace-namespace/workspace-name\"
                     \"c25ee04b-...\"
                     [[\"sample\" \"NA12878\"] [\"sample\" \"NA12879\"]])"
  [workspace set-name [entity & _more :as entities]]
  (let [[type _name] entity]
    (->> entities
         (map (partial vector set-name))
         (util/columns-rows->terra-tsv :membership [type type])
         .getBytes
         (import-entities workspace))))
```

tbl3rd: Should this generate `set-name`?

tbl3rd: Only 2 columns?  Both named `type`?
Ah. Maybe there is only 1 column.

api/src/wfl/service/firecloud.clj
tbl3rd: Suggest:
```clojure
(defn import-entity-set
  "Format sample ENTITIES into a TSV, and upload it to WORKSPACE as an
  entity set named SET-NAME.  WORKSPACE is a \"namespace/name\" string.
  ENTITIES must share the same type.

  (import-entity-set \"workspace-namespace/workspace-name\"
                     \"c25ee04b-...\"
                     [[\"sample\" \"NA12878\"] [\"sample\" \"NA12879\"]])"
  [workspace set-name [entity & _more :as entities]]
  (let [[type _name] entity]
    (->> entities
         (map (partial vector set-name))
         (util/columns-rows->terra-tsv :membership [type type])
         .getBytes
         (import-entities workspace))))
```

tbl3rd (commented):
I see the rest of the file has the big docstrings with labeled sections and so on. :sigh:

api/src/wfl/service/google/bigquery.clj
rexwangcc: My IDE does this alignment too (and I quite like it tbh :P)

api/src/wfl/service/firecloud.clj
okotsopoulos: - Method signature updated to accept a list of entities
- Integration test added to model how to call this method
- RE: " I'd also add the name of the entity you want to create as an input because we'd probably want to append the set to an existing table in the workspace" -- I believe I've addressed this by allowing the entity set name to be optionally specified, this proved useful for writing tests too.  LMK if I've misunderstood.

api/src/wfl/service/firecloud.clj
okotsopoulos: > Should this generate `set-name`?

No, we want access to this name further upstream so that we can specify it in submission creation, among other things (testing applications).

> Only 2 columns? Both named `type`?
> Ah. Maybe there is only 1 column.

I admittedly don't love this, but these base column names are then formatted by the `util/columns-rows->terra-tsv`.  For example `type=sample`:
- First column gets transformed into the Terra-compatible primary key column name, i.e. `membership:sample_set_id`
- Second column undergoes no further transformation on our end, but Terra needs it to match the passed-in entity type.  The resulting entity set table will see this column pluralized, i.e. `sample -> samples`.

api/src/wfl/service/firecloud.clj
okotsopoulos: Done.  And as a bonus it allowed me to remove this intermediate method!

api/src/wfl/util.clj
okotsopoulos: Done.

api/test/wfl/integration/firecloud_test.clj
```diff
@@ -39,6 +59,20 @@
...
+   #(let [workflow (-> (firecloud/get-submission workspace %) :workflows first)]
+      (is (= (second entity) (get-in workflow [:workflowEntity :entityName])))
+      (is (#{"Queued" "Submitted"}
+           ;; GH-1212: `get-submission` can return the workflow before it has
+           ;; been queued. In such cases :status is nil so try again
+           (or (:status workflow) (-> (firecloud/get-submission workspace %)
+                                      :workflows
+                                      first
+                                      :status)))))))
+
```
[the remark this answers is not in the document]
okotsopoulos: D'oh!  Thanks.  Cleaned up the duplicated method.

I learned that what I thought was the "method configuration" was actually the method.  After my last push, I altered the root entity type for a method configuration in the Terra UI and this broke several tests.  I changed the root entity type back and they work again.

To support submission creation for entity sets, we don't need to change the root entity type from `<entity-type> -> <entity-type>_set` but we do need to specify `expression=this.<entity-type>s` in the submission creation request.

api/test/wfl/integration/firecloud_test.clj
```diff
@@ -25,6 +25,26 @@
...
+   #(firecloud/abort-submission workspace %)
+   #(let [workflow (-> (firecloud/get-submission workspace %) :workflows first)]
+      (is (= (second entity) (get-in workflow [:workflowEntity :entityName])))
+      (is (#{"Queued" "Submitted"}
+            ;; GH-1212: `get-submission` can return the workflow before it has
+            ;; been queued. In such cases :status is nil so try again
+           (or (:status workflow) (-> (firecloud/get-submission workspace %)
+                                      :workflows
+                                      first
+                                      :status)))))))
```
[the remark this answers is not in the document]
okotsopoulos: https://github.com/broadinstitute/wfl/pull/353#discussion_r611991092

api/test/wfl/integration/firecloud_test.clj
```diff
@@ -39,6 +59,20 @@
...
+      (is (= (second entity) (get-in workflow [:workflowEntity :entityName])))
+      (is (#{"Queued" "Submitted"}
+           ;; GH-1212: `get-submission` can return the workflow before it has
+           ;; been queued. In such cases :status is nil so try again
+           (or (:status workflow) (-> (firecloud/get-submission workspace %)
+                                      :workflows
+                                      first
+                                      :status)))))))
+
 (defmacro ^:private using-assemble-refbased-workflow-bindings
```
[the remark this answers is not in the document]
okotsopoulos: I will reach out tomorrow to ask about this.

api/src/wfl/service/google/bigquery.clj
okotsopoulos: Ah, you know I think I might have introduced this alignment myself because I saw it done in someone else's PR!

For future reference, do we have a preference here?

okotsopoulos:
> I see the rest of the file has the big docstrings with labeled sections and so on. :sigh:

I am getting the sense that everyone has a different preference for docstrings 😅 you've figured out mine, but if there is a universal team standard I can adjust.

api/src/wfl/util.clj
okotsopoulos: I've since observed that while an entity set shows duplicate entry references, if you create a submission off of the entity set it will only create one workflow for each distinct entity reference.

api/src/wfl/service/firecloud.clj
tbl3rd: Thanks for the imperative mood.  Feel free to ignore the rest of my docstring comments as we move to a new standard.

ehigham (approved):

rexwangcc (approved):
LGTM, thank you!! :octocat:

PR #355, opened 2021-04-06
Snapshot creation uses datetime rather than date

rhiananthony:
### Purpose
The purpose of this ticket is to allow snapshot creation to use a datetime field rather than a date field for creation. 

This is for ticket https://broadinstitute.atlassian.net/browse/GH-1269

### Changes
Using datetime field instead of date field for selecting rows to use in creation of a dataset snapshot

### Review Instructions
Nothing special

### DIFF
--- a/api/src/wfl/service/datarepo.clj
+++ b/api/src/wfl/service/datarepo.clj
@@ -228,8 +228,7 @@
   (let [bq-name (bigquery-name dataset)
         query   "SELECT %s
                  FROM   `%s.%s.%s`
-                 WHERE  datarepo_ingest_date > '%s'
-                 AND    datarepo_ingest_date <= '%s'"]
+                 WHERE updated BETWEEN '%s' AND '%s'"]
     (->> (format query col-spec dataProject bq-name table start end)
          (bigquery/query-sync dataProject))))
 
--- a/api/src/wfl/service/datarepo.clj
+++ b/api/src/wfl/service/datarepo.clj
@@ -224,12 +224,12 @@
         (query-table-impl dataset table))))
 
 (defn ^:private query-table-between-impl
-  [{:keys [dataProject] :as dataset} table [start end] col-spec]
+  [{:keys [dataProject] :as dataset} table column-name [start end] col-spec]
   (let [bq-name (bigquery-name dataset)
         query   "SELECT %s
                  FROM   `%s.%s.%s`
-                 WHERE updated BETWEEN '%s' AND '%s'"]
-    (->> (format query col-spec dataProject bq-name table start end)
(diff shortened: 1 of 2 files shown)


api/src/wfl/service/datarepo.clj
```diff
@@ -228,8 +228,7 @@
   (let [bq-name (bigquery-name dataset)
         query   "SELECT %s
                  FROM   `%s.%s.%s`
-                 WHERE  datarepo_ingest_date > '%s'
-                 AND    datarepo_ingest_date <= '%s'"]
+                 WHERE updated BETWEEN '%s' AND '%s'"]
```
ehigham, line 231: I'd like to not hard-code this field as no doubt it'll be different in the real dataset. Can you pass in the name of the field instead?

tbl3rd: `wfl.util/today` and so on still return a `LocalDate`?

ehigham: Yes i think so. I thought they returned a `LocalDate` at UTC+0

tbl3rd: How does that fit with the change in datarepo_test.clj?

ehigham: Guess it doesn't?

tbl3rd: Seems WFL should either use datetime or timestamp,
or query-table-between should take date,
but it works if it works.
(And test-create-snapshot passes.)

ehigham: I think the idea is that you give it an interval of the same time as the field.

rhiananthony: I resolved the original ask, but am confused on the conversation direction. Does this mean that we should be able to take in both inputs (date and datetime)?

tbl3rd: Presumably it should take whatever datatype BQ SQL's `BETWEEN` syntax expects.
I saw the interval to query-table-between constructed with `util/today` and so on.
`util/today` returns a `LocalDate` but it looks like query-table-between takes `DATETIME`s,
(or something) at least according to the `datarepo_test.clj` change below.
(But maybe I'm confused again about how to handle time in queries.)

tbl3rd (approved):

api/src/wfl/service/datarepo.clj
rexwangcc: I think `wfl.tools.snapshots/create-snapshots` is an outdated test now as we don't want to use DATEs anymore. (that test is not used anyways). Maybe that's where @tbl3rd 's confusion came from?

PR #356, opened 2021-04-06
GitHub `develop` infrastructure + release changes

ehigham:
Update test actions to only run on PRs.
Update release action to only run on new commits into `main`.
Update + format docs about branching off `develop` and releasing into `main`.


docs/md/dev-process.md
```diff
@@ -1,26 +1,22 @@
...
-and encourage ourselves
-to follow in most cases.
+This is a development process we are tying to standardize within the team and
+encourage ourselves to follow in most cases.
 
 ## The Swagger page
 
-WFL ships with a swaggger UI that documents all of the
-available endpoints. It's available at path `/swagger`, e.g.
+WFL ships with a swaggger UI that documents all of the available endpoints.
```
okotsopoulos, line 8: Good time to remove the third g in swagger :)

docs/md/dev-process.md
```diff
@@ -77,7 +72,7 @@ For the release process, please refer to the [release guide](../dev-release/)
     shown below so JIRA could pick up the branch and link it
```
okotsopoulos, line 72: Off topic but I thought I did this correctly on my PR and did not see the link take place.  Do we know if this still works?

docs/md/dev-process.md
```diff
@@ -1,26 +1,22 @@
...
+This is a development process we are tying to standardize within the team and
+encourage ourselves to follow in most cases.
 
 ## The Swagger page
 
-WFL ships with a swaggger UI that documents all of the
-available endpoints. It's available at path `/swagger`, e.g.
+WFL ships with a swaggger UI that documents all of the available endpoints.
+It's available at path `/swagger`, e.g.
 https://dev-wfl.gotc-dev.broadinstitute.org/swagger
```
okotsopoulos, line 10: I tripped on this when setting up WFL: because Swagger is bundled with the UI and not the API we can't hit the endpoint directly.  Ultimately I know we wish to fix this but in the meantime, could update the docs to indicate that we first must go to WFL UI directly and log in before accessing Swagger.

docs/md/dev-process.md
```diff
@@ -31,15 +27,16 @@ The steps for getting this project set up with very recent versions of IntelliJ
 differ from Cursive's docs:
 
 ???+ tip
-    It is recommended to run `prebuild` before launching IntelliJ
-    as it sets up all libraries and derived resources and sources:
+    Run `make prebuild` before launching IntelliJ as it sets up all libraries
+    and derived resources and sources:
     ```bash
     make TARGET=prebuild - jN
```
okotsopoulos, line 33: I think no space between `-` and `jN`?

docs/md/dev-process.md
ehigham: Ha! Good spot!

docs/md/dev-process.md
ehigham: It should be in the "development" section of the ticket on the right-hand-side?

docs/md/dev-process.md
ehigham: Thanks!

.github/workflows/release.yml
```diff
@@ -1,12 +1,9 @@
...
 on:
-  workflow_dispatch:
-    inputs:
-      ref:
-        description: branch, tag or SHA to checkout.
-        required: false
-        default: 'main'
+  push:
+    branches:
+      - main
```
rexwangcc, line 6: I don't think this Action works anymore, as it calls `./ops/cli.py tag-and-push-images` which was removed to moved to `gotc-deploy`, See https://github.com/broadinstitute/gotc-deploy/pull/274. We may need to think about how to release from `gotc-deploy`, or add the cli back.

docs/md/dev-process.md
rexwangcc: interesting, I never followed this naming convention on my branches, but as long as I use `[GH-xxx]` in my PR title, Jira picks it up.

rexwangcc (approved):
Thanks for updating the docs and actions! LGTM, I left a comment about the action.

.github/workflows/release.yml
ehigham: Ugh why did that get absorbed into cli.py????

ehigham: `gotc-deploy` should not control software releases.

ehigham: Thank you - I've restored it.

docs/md/dev-process.md
ehigham: Added a clarifying comment

PR #361, opened 2021-04-09
GH-1282: Document WFL's support for Somatic Genomes.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1282

### Changes

- Like it says.

### Review Instructions

- This is documentation.  Read it?

### DIFF
--- a/docs/md/modules-sg.md
+++ b/docs/md/modules-sg.md
@@ -1,0 +1,332 @@
+# GDCWholeGenomeSomaticSingleSample
+
+## Inputs
+
+In addition to the standard workload request inputs:
+
+- `executor` : URL of the Cromwell service
+- `output`   : GCS URL prefix for output files
+- `pipeline` : literally `"GDCWholeGenomeSomaticSingleSample"`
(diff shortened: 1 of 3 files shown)


docs/md/modules-sg.md
```diff
@@ -0,0 +1,332 @@
...
+
+#### `base_file_name`
+
+The leaf name
+of a sample input or output path
+without the `.` suffix.
+The `base_file_name`
+is usually the same
+as the _sample name_
+and differs in every workflow.
```
ehigham, line 35: Why are the columns so short?

ehigham (commented):
Looks great. Learned something from reading this! 
Where does the term "leaf name" come from? I always called it the "base name" after the POSIX tool `basename`

ehigham (approved):

docs/md/modules-sg.md
tbl3rd: Because that's the way everyone should write markup documentation.

PR #365, opened 2021-04-12
Release v0.6.1 Patch

ehigham:
This patch inlcudes:
- Fixed docker image generation
- Fixed workload update loop termination


api/Dockerfile
```diff
@@ -3,7 +3,7 @@ WORKDIR /tmp
 
 # Note this step requires docker builds in the root repo context
 # e.g. `docker build -t wfl-api:test -f ./api/Dockerfile .`
-ADD .. .
+ADD . .
```
rexwangcc, line 6: Screen Shot 2021-04-12 at 2 44 04 PM

I'm still puzzled by the issue the Action is reporting, locally, from with the root `make api TARGET=images` works both before and after the change. Let's see if this fixes the Action :octocat:

rexwangcc (approved):
Thanks for fixing these!

api/Dockerfile
ehigham: Interesting. This failed on my machine. Did the docker build actually run when you tried the before and after?

rexwangcc: I believe so.. I ran both `make distclean` and `docker system prune -a` before `make api TARGET=images` 🤔 I can double check.

tbl3rd: How did that ever work then?

tbl3rd (approved):

api/Dockerfile
ehigham: Honestly I have no idea. Different versions of docker perhaps? I never thought you could add the parent directory.

PR #369, opened 2021-04-14
Create an interface all COVID work can build on top of.

rexwangcc:
### Purpose

### Changes

- Try to make a common ground for GH-1213, GH-1215 and GH-1285.

### Review Instructions

- Please look at both [SPIKE doc](https://broadinstitute.atlassian.net/wiki/spaces/GHConfluence/pages/2128117992/COVID-19+Surveillance+in+Terra) and here, especially the DB schema changes (note this is likely not the final design we would have, but a starting point)


api/src/wfl/module/covid.clj
```diff
@@ -0,0 +1,41 @@
...
+            (postgres/update-workload-status! tx workload)
+            (workloads/load-workload-for-id tx id))]
+    (if (and started (not finished)) (update! workload) workload)))
+
+(defoverload workloads/create-workload!   pipeline create-covid-workload!)
+(defoverload workloads/start-workload!    pipeline start-covid-workload!)
+(defoverload workloads/update-workload!   pipeline update-covid-workload!)
+(defoverload workloads/stop-workload!     pipeline batch/stop-workload!)
+(defoverload workloads/load-workload-impl pipeline
+  batch/load-batch-workload-impl)
```
rexwangcc, line 41:
```suggestion
  batch/load-batch-workload-impl)

```

database/changesets/20210414_Sarscov2IlluminaFullTable.xml
```diff
@@ -0,0 +1,32 @@
...
+            snapshot_id                 text,        -- Terra DataRepo snapshot id
+            datarepo_row_id             text,        -- Unique id to identify a row in snapshot
+            submission_id               text,        -- FireCloud submissionId
+            workflow_id                 text,        -- FireCloud workflow id
+            status                      text,        -- FireCloud workflow status
+            updated                     timestamptz  -- status update time
+            )
+        </sql>
+    </changeSet>
+</databaseChangeLog>
```
rexwangcc, line 32:
```suggestion
</databaseChangeLog>

```

api/src/wfl/module/covid.clj
```diff
@@ -0,0 +1,41 @@
...
+  [tx {:keys [source sink executor] :as request}])
+
+;; TODO: implement progressive (private) functions inside this update loop
+(defn update-covid-workload!
+  "Use transaction `tx` to batch-update `workload` statuses."
+  [tx {:keys [started finished] :as workload}]
+  (letfn [(update! [{:keys [id] :as workload}]
+            (postgres/batch-update-workflow-statuses! tx workload)
+            (postgres/update-workload-status! tx workload)
+            (workloads/load-workload-for-id tx id))]
```
okotsopoulos, line 33: I think I understand what the other functions are doing in this loop but not this one.  Would be a help to hear to explain it!

api/src/wfl/module/covid.clj
```diff
@@ -0,0 +1,41 @@
...
+(def pipeline "Sarscov2IlluminaFull")
+
+;; TODO: implement COVID workload creation
+;;  - make sure permissions/inputs are right upfront
+;;  - dispatch on source/sink/executor
+;;  - store information into database
+(defn create-covid-workload!
+  [tx {:keys [source sink executor] :as request}])
+
+;; TODO: implement progressive (private) functions inside this update loop
```
okotsopoulos, line 26: Thanks for flagging this.  Looks like this is the space in which we'll all be working simultaneously.

database/changesets/20210414_Sarscov2IlluminaFullTable.xml
```diff
@@ -0,0 +1,32 @@
...
+        <sql>
+            CREATE TYPE Sarscov2IlluminaFull AS (
+            id                          bigint,      -- just an unique id
+            dataset_id                  text,        -- Terra DataRepo dataset id
+            snapshot_job_id             text,        -- Terra DataRepo snapshot creation job id
+            snapshot_id                 text,        -- Terra DataRepo snapshot id
+            datarepo_row_id             text,        -- Unique id to identify a row in snapshot
+            submission_id               text,        -- FireCloud submissionId
+            workflow_id                 text,        -- FireCloud workflow id
+            status                      text,        -- FireCloud workflow status
```
okotsopoulos, line 27: Is this value a candidate for an enum definition (not sure how widely we use these)?

api/src/wfl/module/covid.clj
rexwangcc: Yup! That's what I'm imagining, we could write function outside of this and thread them in this block with `cond` or `when`s.

database/changesets/20210414_Sarscov2IlluminaFullTable.xml
rexwangcc: That's a good point! If FireCloud's workflows are similar to Cromwell workflows, then this will be a finite set of strings (`Failed`, `Succeeded`, `Running`, `On Hold`, etc.). Historically we have just used plain text for this field, so I inherited that behavior. Maybe we could turn this into an Enum now?

okotsopoulos: If not too much trouble to change, that would be my vote.  When I see a finite set of strings I like enums in part for the built-in explanation they provide on expected / supported values.

If you'd rather merge now in the name of expediency and revisit later (if not too difficult to change an existing schema) then fine by me as well.

tbl3rd: On the other hand, WFL doesn't control those values.
Enumerating them means a Firecloud change requires a WFL schema change.
On the third hand, WFL probably uses those status values in code too.
On the fourth hand, WFL code changes are somewhat easier than database schema changes.

api/src/wfl/module/covid.clj
```diff
@@ -0,0 +1,41 @@
...
+            [clojure.tools.logging.readable :as log]
+            [wfl.api.workloads :as workloads :refer [defoverload]]
+            [wfl.jdbc :as jdbc]
+            [wfl.module.batch :as batch]
+            [wfl.service.datarepo :as datarepo]
+            [wfl.service.firecloud :as firecloud]
+            [wfl.service.google.storage :as gcs]
+            [wfl.service.postgres :as postgres]
+            [wfl.util :as util]
+            [wfl.wfl :as wfl]))
```
tbl3rd, line 15: `wfl.module.covid` *may* `require` all of those eventually.

readme.md
```diff
@@ -1,6 +1,6 @@
 # WorkFlow Launcher
 
-[![Build Board](https://img.shields.io/badge/-The%20Build%20Boad%20(VPN)-blue)](https://internal.broadinstitute.org/~chengche/green-hornet-status.html)
+[![Build Board](https://img.shields.io/badge/-The%20Build%20Board%20(VPN)-blue)](https://internal.broadinstitute.org/~chengche/green-hornet-status.html)
```
tbl3rd, line 3: Thanks.

tbl3rd (approved):

database/changesets/20210414_Sarscov2IlluminaFullTable.xml
okotsopoulos: Given the count of hands here, I'll approve the PR and we can elect to discuss further later.

okotsopoulos (approved):

database/changesets/20210414_Sarscov2IlluminaFullTable.xml
rexwangcc: Sounds good, thank you both!!

PR #370, opened 2021-04-16
[GH-1215] Add wfl.service/rawls.clj for interacting directly with Rawls API

okotsopoulos:
### Purpose

TDR = Terra Data Repository
TW = Terra Workspace

For my ticket shared above as well as other COVID tickets to do and in progress, we will want to communicate directly with the Rawls API to interact with snapshots from TDR.

The general relationship between values in snapshot-land:
- An existing snapshot in TDR will be identified by a **snapshot-id**.
- When we link it in a TW, we will supply a **name**: subsequent attempts to import a snapshot with the same name will fail with a 409.
- When the snapshot is linked, we will get back a **reference-id**, the linked snapshot's unique ID within the TW.  This is the value Rawls expects when we fetch snapshot information or delete the snapshot from its TW.  Note that I couldn't see the reference-id exposed within the Terra UI, you have to dig a bit to find it if you don't have it from snapshot creation.

This PR does not close out my ticket, but as I'll be out through 4/21 please feel free to add to it and / or merge if it would be helpful for your own development in the space.

### Changes

- New `wfl.service/rawls.clj` with initial support for snapshot fetch, creation, and deletion.
- New integration tests.

### Review Instructions

- On this FB, can use new integration tests as a guide for how to interact with new endpoints.
- Note that if the following snapshot is altered or deleted, integration test(s) may fail:
Screen Shot 2021-04-16 at 5 25 00 PM

### DIFF
--- a/api/src/wfl/environment.clj
+++ b/api/src/wfl/environment.clj
@@ -48,6 +48,8 @@
    #(-> nil)
    "WFL_FIRECLOUD_URL"
    #(-> "https://api.firecloud.org")
+   "WFL_RAWLS_URL"
+   #(-> "https://rawls.dsde-dev.broadinstitute.org")
 
    ;; -- variables used in test code below this line --
    "WFL_CROMWELL_URL"
--- a/api/src/wfl/service/rawls.clj
+++ b/api/src/wfl/service/rawls.clj
@@ -1,0 +1,52 @@
+(ns wfl.service.rawls
+  "Analyze data in Terra using the Rawls API.
+  Note that Firecloud is built on top of Rawls:
+  Rawls may expose functionality that hasn't yet been promoted to Firecloud."
+  (:require [clj-http.client :as http]
+            [clojure.data.json :as json]
+            [clojure.string :as str]
+            [wfl.auth :as auth]
+            [wfl.environment :as env]
+            [wfl.util :as util]))
+
+(defn ^:private rawls-url [& parts]
+  (let [url (util/de-slashify (env/getenv "WFL_RAWLS_URL"))]
+    (str/join "/" (cons url parts))))
+
+(def ^:private workspace-api-url
+  (partial rawls-url "api/workspaces"))
+
+(defn ^:private get-workspace-json [& parts]
+  (-> (apply workspace-api-url parts)
+      (http/get {:headers (auth/get-auth-header)})
+      util/response-body-json))
+
+(defn create-snapshot
+  "Link a Terra Data Repo snapshot with id SNAPSHOT-ID to a fully-qualified
+  Terra WORKSPACE as NAME, optionally described by DESCRIPTION."
+  ([workspace snapshot-id name description]
+   (-> {:method       :post
+        :url          (workspace-api-url workspace "snapshots")
+        :headers      (auth/get-auth-header)
+        :content-type :application/json
+        :body         (json/write-str {:snapshotId snapshot-id
+                                       :name name
+                                       :description description}
+                                      :escape-slash false)}
+       http/request
+       util/response-body-json
+       :referenceId))
+  ([workspace snapshot-id name]
+   (create-snapshot workspace snapshot-id name "")))
+
+(defn get-snapshot
+  "Return the snapshot in fully-qualified Terra WORKSPACE with REFERENCE-ID."
+  [workspace reference-id]
+  (get-workspace-json workspace "snapshots" reference-id))
+
+(defn delete-snapshot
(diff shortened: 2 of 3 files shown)


api/src/wfl/environment.clj
```diff
@@ -48,6 +48,8 @@
    #(-> nil)
    "WFL_FIRECLOUD_URL"
    #(-> "https://api.firecloud.org")
+   "WFL_RAWLS_URL"
+   #(-> "https://rawls.dsde-dev.broadinstitute.org")
```
okotsopoulos, line 52: Per Ed's comment in Slack, we can only interact with TDR from Rawls when both are in the same environment, i.e. both dev or both prod.  The existing TDR in this file points to dev.

api/src/wfl/service/rawls.clj
```diff
@@ -0,0 +1,52 @@
...
+  Note that Firecloud is built on top of Rawls:
+  Rawls may expose functionality that hasn't yet been promoted to Firecloud."
+  (:require [clj-http.client :as http]
+            [clojure.data.json :as json]
+            [clojure.string :as str]
+            [wfl.auth :as auth]
+            [wfl.environment :as env]
+            [wfl.util :as util]))
+
+(defn ^:private rawls-url [& parts]
```
okotsopoulos, line 12: These private functions are basically copy-pasted from `firecloud` which makes me feel :shame:

api/test/wfl/integration/rawls_test.clj
```diff
@@ -0,0 +1,36 @@
...
+  (testing "Creating snapshot with same name of existing throws 409 error"
+    (is (thrown-with-msg? ExceptionInfo #"clj-http: status 409"
+                          (rawls/create-snapshot workspace
+                                                 snapshot-id
+                                                 (:name known-snapshot-link)))))
+  (let [name (str (UUID/randomUUID))]
+    (util/bracket
+     #(rawls/create-snapshot workspace snapshot-id name "wfl.rawls-test/test-create-snapshot")
+     #(rawls/delete-snapshot workspace %)
+     #(let [snapshot (rawls/get-snapshot workspace %)]
```
okotsopoulos, line 33: I worry this test could flap, if a similar issue to [GH-1212](https://broadinstitute.atlassian.net/browse/GH-1212) occurs.

If this happens, maybe we leverage the newish [polling utility](https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/util.clj#L414-L429).  I ran out of time before I could fully understand how it works.

api/src/wfl/environment.clj
rexwangcc: Thanks for flagging it, makes sense to me. In order to use prod Terra and RAWLs in WFL, we probably need to update the ENVs in deployment helmfile.

rexwangcc (approved):
LGTM! Thanks for putting this together so quickly! All makes sense to me, I'm not sure if we want to call it `create-snapshot` or `link`/`import`, naming is hard to me.

api/src/wfl/service/rawls.clj
```diff
@@ -0,0 +1,52 @@
...
+  (let [url (util/de-slashify (env/getenv "WFL_RAWLS_URL"))]
+    (str/join "/" (cons url parts))))
```
tbl3rd, line 14: I see this was copied, but do we need to `de-slashify` in either place?
Is there a source for `WFL_RAWLS_URL` that WFL does not control?

tbl3rd (approved):
Thanks.

api/src/wfl/environment.clj
okotsopoulos: Good to know, that makes sense.

api/src/wfl/service/rawls.clj
okotsopoulos: Speculation only: my understanding of how things work in `environment.clj` tells me that a user could override the value of `WFL_RAWLS_URL` in the command line.

okotsopoulos:
> LGTM! Thanks for putting this together so quickly! All makes sense to me, I'm not sure if we want to call it `create-snapshot` or `link`/`import`, naming is hard to me.

Thanks for making this suggestion, I feel you on the naming question.  I think I went with `create-snapshot` because it's consistent with the Rawls endpoint top-level description, but it is an overloaded concept that fits better for true snapshot creation in the TDR.

https://rawls.dsde-dev.broadinstitute.org/#/snapshots/createDataRepoSnapshot

I am inclined to rename to `create-snapshot-reference`: is consistent with the Rawls endpoint detailed descriptions and makes it more clear why reference IDs are being returned / passed around.  I'll open a separate PR for this.

PR #371, opened 2021-04-20
[GH-1222] Check for new work in TDR

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1222

### Changes

- Create functions that poll TDR and create snapshots. They also write snapshot creation job id into the database.
- Fix the system test by bumping the timeout.
- Update some docs.

### Review Instructions

- Note I didn't add test cases since the create snapshot functionalities are covered by integration tests of the `datarepo` and `bigquery` modules.


api/src/wfl/module/covid.clj
```diff
@@ -23,7 +24,51 @@
 (defn create-covid-workload!
   [tx {:keys [source sink executor] :as request}])
 
-;; TODO: implement progressive (private) functions inside this update loop
+(defn ^:private snapshot-created?
+  "Check if snapshot has been created in TDR, return snapshot id if so."
+  [{:keys [snapshot_job_id] :as _record}]
```
okotsopoulos, line 29: Is the idea that this check would be called with a record from the `Sarscov2IlluminaFull` instance as the argument?

api/src/wfl/module/covid.clj
```diff
@@ -34,7 +79,7 @@
     (if (and started (not finished)) (update! workload) workload)))
 
 (defoverload workloads/create-workload!   pipeline create-covid-workload!)
-(defoverload workloads/start-workload!    pipeline start-covid-workload!)
+(defoverload workloads/start-workload!    pipeline batch/stop-workload!)
```
okotsopoulos, line 82: This looks like it is incorrect.

rexwangcc: yeah, this needs to be updated when we implement the `start` for covid anyways.

tbl3rd: FWIW: I have a mock on my branch.

api/src/wfl/module/covid.clj
```diff
@@ -25,8 +18,65 @@
...
+(defn ^:private go-create-snapshot!
+  "Create snapshot in TDR from `dataset`, `table`
+   and `row-ids` and write job info as well as rows
+   into `source-details-name` table."
+  [tx dataset table source-details-name [start end] row-ids]
+  (let [columns     (-> (datarepo/all-columns dataset table)
+                        (->> (map :name) set)
+                        (conj "datarepo_row_id"))
+        job-id (-> (datarepo/make-snapshot-request dataset columns table row-ids)
+                   (update :name util/randomize)
```
ehigham, line 39: I don't like the idea of using randomisation in product code. Furthermore, this name will mean nothing to users of the workspace.
Can we work along the lines of [Dataset Name][YYYYMMDDTHHMMSS] - the slightly unconventional timestamp is because it has to be an identifier.

api/src/wfl/module/covid.clj
```diff
@@ -25,8 +18,65 @@
...
+                   ;; FIXME: this sometimes runs into: Cannot update iam permissions, need failure handling here!
```
ehigham, line 40: This is a problem with the DEV data repo. We should test this using the production TDR. You can use the credentials in this ticket for testing:
https://broadinstitute.atlassian.net/browse/GH-1294

api/src/wfl/module/covid.clj
```diff
@@ -25,8 +18,65 @@
...
+                   :start_time start
+                   :end_time end})))
+
+(defn ^:private check-for-new-data!
+  "Check for new data in TDR, create new snapshots, insert
+   resulting job creation ids into database and update the
+   timestamp for next time using transaction TX."
+  [tx
+   {:keys [source_item] :as workload}
+   {:keys [dataset table last_checked snapshot details] :as _source}]
```
ehigham, line 55: Maybe we want to rename this `snapshot` - I think it's confusing.

api/src/wfl/module/covid.clj
```diff
@@ -25,8 +18,65 @@
...
+  (let [now (util/utc-now)
+        dataset (datarepo/dataset dataset)
+        table table
```
ehigham, line 58: remove this unnecessary rebinding?

api/src/wfl/module/covid.clj
```diff
@@ -25,8 +18,65 @@
...
+           :id))
+
+(defn ^:private go-create-snapshot!
+  "Create snapshot in TDR from `dataset`, `table`
+   and `row-ids` and write job info as well as rows
+   into `source-details-name` table."
+  [tx dataset table source-details-name [start end] row-ids]
+  (let [columns     (-> (datarepo/all-columns dataset table)
+                        (->> (map :name) set)
+                        (conj "datarepo_row_id"))
```
ehigham, line 37: We could cache this I think in the source table to avoid making extra calls - what do you think?

api/src/wfl/module/covid.clj
```diff
@@ -25,8 +18,65 @@
...
+
+(defn ^:private check-for-new-data!
+  "Check for new data in TDR, create new snapshots, insert
+   resulting job creation ids into database and update the
+   timestamp for next time using transaction TX."
+  [tx
+   {:keys [source_item] :as workload}
+   {:keys [dataset table last_checked snapshot details] :as _source}]
+  (let [now (util/utc-now)
+        dataset (datarepo/dataset dataset)
```
ehigham, line 57: Perhaps lift this out of the function? This information won't change so we probably only need to store it once - see comment about caching above?

api/src/wfl/module/covid.clj
```diff
@@ -25,8 +18,65 @@
...
+        row-ids (-> (datarepo/query-table-between
+                     dataset
+                     table
+                     col-name
+                     [last_checked now]
+                     [:datarepo_row_id])
+                    :rows
+                    flatten)]
+    (when row-ids
+      (let [shards (partition-all 500 row-ids)]
```
ehigham, line 69: i like `shards`!

api/src/wfl/module/covid.clj
```diff
@@ -25,8 +18,65 @@
...
+        (map #(go-create-snapshot! tx dataset table details [last_checked now] %) shards)))
+    (do
+      (jdbc/update! tx
+                    source_item
+                    {:last_checked now}
+                    ["id = ?" (:id workload)])
+      (jdbc/update! tx
```
ehigham, line 76: Let another handler update the workload table. (i.e. the handler that calls `update-source!`)

api/src/wfl/protocols.clj
```diff
@@ -0,0 +1,23 @@
+(ns wfl.protocols
```
ehigham, line 1: Ha! we did the same thing - perhaps move to our `jdbc` module to avoid an extra import?
https://github.com/broadinstitute/wfl/pull/386/files

api/src/wfl/service/datarepo.clj
```diff
@@ -97,6 +97,15 @@
   ([job-id]
    (poll-job job-id 5)))
 
+(defn job-done?
```
ehigham, line 100: I really really don't like this style of writing predicates. Likewise, I hate it when `throw-if-foo` functions return the foo when it's valid. 
I'd be fine if it was called something else like `get-result-when-done`. I'd be fine with two functions `job-done?` and `get-job-result` to be used in a `when` which might make sense as you'll want to find the first job that's finished in the queue.

api/src/wfl/service/datarepo.clj
```diff
@@ -97,6 +97,15 @@
...
+  "Check the job with `job-id` to see if it's done yet."
+  [job-id]
+  (let [result   #(get-repository-json "jobs" job-id "result")
+        running? #(-> (get-repository-json "jobs" job-id)
+                      :job_status
+                      #{"running"})]
+    (if (running?) false (result))))
```
ehigham, line 107: If you tweaked `running?` to test if it were done, the following might be more natural
`(and (done?) (result))`

api/test/wfl/integration/protocol_test.clj
```diff
@@ -0,0 +1,25 @@
+(ns wfl.integration.protocol-test
+  (:require [clojure.test :refer :all :as clj-test]
+            [clojure.test]
+            [wfl.jdbc :as jdbc]
+            [wfl.tools.fixtures :as fixtures]
+            ; require protocols namespace is important
```
ehigham, line 6: see my comment about putting it in `wfl.jdbc`

api/src/wfl/util.clj
```diff
@@ -503,3 +493,9 @@
      (columns-rows->tsv [(format-entity-type columns)] rows file)))
   ([tsv-type columns rows]
    (str (columns-rows->terra-tsv tsv-type columns rows (StringWriter.)))))
+
+(defn utc-now
+  []
+  (-> (ZoneId/of "UTC")
+      (OffsetDateTime/now)
+      (.format (DateTimeFormatter/ofPattern "yyyy-MM-dd HH:mm:ss"))))
```
ehigham, line 501: I think this will cause problems.
First we don't know what time zone the dataset is using for its `run-date` column.
Second, big query needs a Timestamp in the form `yyyy-MM-ddTHH:mm:ss` - this way your queries only have date precision.

api/src/wfl/module/covid.clj
ehigham: Note that timestamp is when the snapshot was created

api/src/wfl/module/covid.clj
```diff
@@ -25,8 +18,65 @@
...
+                    :workload
+                    {:updated (OffsetDateTime/now)}
+                    ["id = ?" (:id workload)]))))
```
tbl3rd, line 79: `"Create and record snapshot jobs in WORKFLOW with new data from SOURCE."`?
Why rename `snapshot` to `col-name` (and `table` to … `table`)?
Don't need `do`, I think.
What does `utc-now` add?
Maybe `java.time.Instant/now` instead, if necessary.
Can `row-ids` be `nil` or is it always a seq?
Let's talk about this in the afternoon session.

api/src/wfl/protocols.clj
```diff
@@ -0,0 +1,23 @@
...
+      (if-let [elem-type (when (= (first type-name) \_) (apply str (rest type-name)))]
+        (.setObject stmt i (.createArrayOf conn elem-type (to-array v)))
+        (.setObject stmt i v)))))
+
+;; extend protocol so psql arrays will be returned
+;; as clojure native vectors
+(extend-protocol clojure.java.jdbc/IResultSetReadColumn
+  Array
+  (result-set-read-column [val _ _]
+    (into [] (.getArray val))))
```
tbl3rd, line 23: It will be fun to see that work.  Thanks.

tbl3rd (commented):
Let's talk.

api/src/wfl/util.clj
rexwangcc: Per offline discussion, we need to:
- Ask the working group why this `run_date` field is not a TIMESTAMP but a STRING
- Make sure it's UTC in any cases
- Add a `T` here.

api/src/wfl/module/covid.clj
okotsopoulos: > the slightly unconventional timestamp is because it has to be an identifier

Can you elaborate on this, are there forbidden characters to be aware of? I ask because I should probably make a similar change to the snapshot reference naming.  Otherwise I'd just cast `now` to string and get something like `2021-05-07T14:47:54.440594-04:00`.

rexwangcc: I updated the `(util/utc-now)` so it optionally takes a `format`  argument now, e.g.

```clojure
(util/utc-now "YYYYMMDD'T'HHMMSS")
=> "202105130T190537"
```

api/src/wfl/protocols.clj
rexwangcc: done!

api/src/wfl/protocols.clj
rexwangcc: Moved the tests to `jdbc_test.clj` now

api/src/wfl/util.clj
rexwangcc: Moved (and broke) this into the covid module, based on the latest info, `run_date` field is now a TIMESTAMP! Let's talk about this again.

api/test/wfl/tools/workloads.clj
```diff
@@ -224,7 +224,7 @@
             (let [skipped? #(-> % :uuid util/uuid-nil?)]
               (or (skipped? workflow) ((set cromwell/final-statuses) status))))]
     (let [interval 10
-          timeout  3600]                ; 1 hour
+          timeout  4800]                ; 80 minutes
```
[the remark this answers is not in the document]
rexwangcc: Yeah… should we change then?

api/test/wfl/tools/workloads.clj
```diff
@@ -224,7 +224,7 @@
             (let [skipped? #(-> % :uuid util/uuid-nil?)]
               (or (skipped? workflow) ((set cromwell/final-statuses) status))))]
     (let [interval 10
```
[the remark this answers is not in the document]
tbl3rd: Maybe factor out the two concerns into `with-timeout` and `with-retry` forms,
or 2 different thunkifiers if that's more functional.

api/src/wfl/module/covid.clj
```diff
@@ -207,17 +209,55 @@
...
+                                     snapshots-creation-jobs)))
+          (update-last-checked [tx now]
+            (jdbc/update! tx (keyword tdr-source-table) {:last_checked now} ["id = ?" id]))]
+    (let [now-obj     (OffsetDateTime/now (ZoneId/of "UTC"))
+          now         (.format now-obj (DateTimeFormatter/ofPattern "yyyy-MM-dd HH:mm:ss"))
+          row-ids     (find-new-rows now)
+          tdr-job-ids (make-snapshots now-obj row-ids)]
+      (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+        (write-snapshots-creation-jobs tx details now row-ids tdr-job-ids)
+        (update-last-checked tx now)))))
```
tbl3rd, line 260: Move the letfn into the let and this won't have to pass nows and row-ids around so much.
Or just break them out into separate defns like go-create-snapshot!.
Maybe stringify an #inst and truncate it instead?

api/src/wfl/module/covid.clj
```diff
@@ -207,17 +209,55 @@
...
+   `now` string and `row-ids` then write job info as well
+   as rows into `source-details-name` table."
+  [now dataset table row-ids]
+  (let [columns     (-> (datarepo/all-columns dataset table)
+                        (->> (map :name) set)
+                        (conj "datarepo_row_id"))
+        job-id (-> (datarepo/make-snapshot-request dataset columns table row-ids)
+                   (assoc :name now)
+                   (datarepo/create-snapshot-job))]
+    job-id))
```
tbl3rd, line 223: Whitespace nit.

api/src/wfl/service/datarepo.clj
tbl3rd: Is job-done still used somewhere?

api/src/wfl/util.clj
```diff
@@ -11,7 +11,7 @@
   (:import [java.io File IOException StringWriter Writer]
            [java.nio.file Files]
            [java.nio.file.attribute FileAttribute]
-           [java.time Instant OffsetDateTime]
+           [java.time OffsetDateTime]
            [java.time.temporal ChronoUnit]
            [java.util ArrayList Collections Random UUID]
            [java.util.concurrent TimeUnit TimeoutException]
```
tbl3rd, line 17: Thanks.

api/test/wfl/tools/workloads.clj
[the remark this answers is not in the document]
tbl3rd: :sigh: I don't know where my comments went …
Factor into with-timeout and with-retry forms, or thunking HOFs if that's more functional.

api/test/wfl/tools/workloads.clj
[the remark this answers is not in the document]
tbl3rd: ¯\\_(ツ)_/¯ We could check 30 times instead of 30 thousand times.

tbl3rd (approved):
Weird.  I see some old changes that don't show in GitHub.
And some of the integration tests fail for me.  Looking …

api/src/wfl/module/covid.clj
```diff
@@ -207,17 +209,55 @@
...
+                :rows
+                flatten))
+          (make-snapshots [now-obj row-ids]
+            (let [shards (partition-all 500 row-ids)
+                  compact-now (.format now-obj (DateTimeFormatter/ofPattern "YYYYMMDD'T'HHMMSS"))]
+              (doall (map #(go-create-snapshot! compact-now dataset dataset_table %) shards))))
+          (write-snapshots-creation-jobs [tx now row-ids snapshots-creation-jobs]
+            (jdbc/insert-multi! tx
+                                details
+                                (map (fn [id] {:snapshot_creation_job_id id
```
okotsopoulos, line 247: I'm sorry for missing this in your last PR, but I believe that `snapshot_id` was inadvertently removed from the source details schema, it is still in the LucidChart:

https://lucid.app/lucidchart/63bc90dd-5db6-4611-adef-be87d0e8b6e3/edit?page=7Wftl_-jXJbR#

I can add it back in a small PR but have a few questions:
- Will you be responsible for updating the `snapshot_id` in the source details record? 
- In checking the source queue for available snapshots, I need to find the first unconsumed record with an available snapshot.  Let's talk about this when you're back so that we don't cover the same ground.

api/src/wfl/module/covid.clj
```diff
@@ -207,17 +209,55 @@
...
+                 dataset
+                 dataset_table
+                 table_column_name
+                 [last_checked now]
+                 [:datarepo_row_id])
+                :rows
+                flatten))
+          (make-snapshots [now-obj row-ids]
+            (let [shards (partition-all 500 row-ids)
+                  compact-now (.format now-obj (DateTimeFormatter/ofPattern "YYYYMMDD'T'HHMMSS"))]
```
okotsopoulos, line 242: I will be using this too to generate meaningfully unique-enough snapshot reference names.  Can you break out into a utility please?

okotsopoulos: Actually, as I think about it more, I am using the snapshot name as the base of my snapshot reference name.  So if you're already tagging the name with a timestamp, I don't think I need to do the same (it will get long and confusing).

If you can give me a concrete example of what a snapshot name will be under this code, that would help my development!

api/src/wfl/service/datarepo.clj
rexwangcc: >Is job-done still used somewhere?

Ah good point, it's not used anymore, so I will probably remove it. (The peek/pop will likely have to implement something similar though)

api/src/wfl/module/covid.clj
rexwangcc: Actually I think you found a bug! The way it works now is `(assoc :name now)` which does not give what I expect… It should be updated to `(update s :name #(str % now))`, and a concrete example would be: 

```clojure
(go-create-snapshot! now (datarepo/dataset "85efdfea-52fb-4698-bee6-eef76104a7f4")
    "sample" ["1" "2" "3"])

; => 

{:contents [{:datasetName "zerotest_partition",
             :mode "byRowId",
             :rowIdSpec {:tables [{:columns #{"vcf" "datarepo_row_id" "id" "vcf_index"},
                                   :rowIds ["1" "2" "3"],
                                   :tableName "sample"}]}}],
 :description "This is a test dataset for Zero that uses date partitioning.",
 :name "zerotest_partition202105133T150557",
 :profileId "XXXXX"}
```

okotsopoulos: Ahhh!  Great, thanks for looking at this closer and for providing the concrete example.

api/src/wfl/module/covid.clj
rexwangcc: AH nice catch, I completely missed that field somehow… I will add it back in a later commit!

re your questions, I think we should defo chat cauz I'm not sure who should be checking and updating the source queue for available snapshots now.

api/src/wfl/module/covid.clj
okotsopoulos: Caught something else:
```
YYYYMMDD'T'HHMMSS
```

Note that `MM` is the month and it's used in both the month and minute spots, think what you want instead is:
```
YYYYMMDD'T'HHmmSS
```

Source:
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

api/src/wfl/module/covid.clj
rexwangcc: Fixed

api/src/wfl/module/covid.clj
rexwangcc: Thanks for catching, based on https://help.sumologic.com/03Send-Data/Sources/04Reference-Information-for-Sources/Timestamps%2C-Time-Zones%2C-Time-Ranges%2C-and-Date-Formats, I also updated `SS` to `ss`

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -84,6 +99,29 @@
...
+;; null value in column "id" of relation "terradatareposourcedetails_000000001" violates not-null constraint
+(deftest test-update-tdr-source
+  (let [{:keys [source]}
+        (workloads/create-workload!
+         (workloads/covid-workload-request {} {} {}))]
+    (with-redefs-fn
+      {#'covid/create-snapshots mock-create-snapshots
+       #'covid/find-new-rows mock-find-new-rows
+       #'covid/check-tdr-job mock-check-tdr-job}
+      #(#'covid/update-tdr-source source))
```
ehigham, line 113: Can you cal multi-method instead please? I want to make sure it's plumbed in correctly!

ehigham (approved):

api/test/wfl/integration/modules/covid_test.clj
rexwangcc: Ah sure, will update!

api/src/wfl/module/covid.clj
```diff
@@ -183,17 +183,120 @@
...
+   from the `_tdr-job-metadata` map, update timestamp with real now."
+  [{:keys [details] :as _source}
+   {:keys [id job_status snapshot_id] :as _tdr-job-metadata}]
+  (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+    (jdbc/update! tx details {:snapshot_creation_job_status job_status
+                              :snapshot_id snapshot_id
+                              :updated  (OffsetDateTime/now)}
+      ;; we don't have the primary key reference of details table here,
+      ;; so use job-id for where clause, since job-id is unique in TDR,
+      ;; this shall be fine
```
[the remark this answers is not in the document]
rexwangcc: I was thinking, ideally this SQL might use the primary key in the details table to update the records than the `job-id`, but that requires us to persist the reference to primary key `id`s when performing the query (for pending jobs), and I found it a bit fussy and redundant. This comment was more for myself to justify using `job-id` is qualified for this operation. I can remove it if you think it's confusing!

api/src/wfl/module/covid.clj
```diff
@@ -183,17 +183,120 @@
...
+  "Check for new data in TDR, create new snapshots, insert
+   resulting job creation ids into database and update the
+   timestamp for next time using transaction TX."
+  [source]
+  ;; attempt to snapshot new rows in TDR
+  (let [utc-now-obj     (OffsetDateTime/now (ZoneId/of "UTC"))
+        utc-now         (.format utc-now-obj (DateTimeFormatter/ofPattern "yyyy-MM-dd HH:mm:ss"))
+        row-ids         (vec (find-new-rows source utc-now))
+        tdr-job-ids     (create-snapshots source utc-now-obj row-ids)]
+    (write-snapshots-creation-jobs source utc-now-obj row-ids tdr-job-ids)
```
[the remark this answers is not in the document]
rexwangcc: Yes, nice catch! I found this wrong this morning, which in fact records:

```clojure
[{:datarepo_row_ids [1 2 3 4 5 6 7 8 9 10], :snapshot_creation_job_id "job-1"},
 {:datarepo_row_ids [1 2 3 4 5 6 7 8 9 10], :snapshot_creation_job_id "job-2"}]
```

as opposed to 

```clojure
[{:datarepo_row_ids [1 2 3 4 5], :snapshot_creation_job_id "job-1"},
 {:datarepo_row_ids [6 7 8 9 10], :snapshot_creation_job_id "job-2"}]
```

I have updated the upstream function to return `[shards job-ids]` so this function will pick up and write the correct information to the DB.

api/src/wfl/module/covid.clj
```diff
@@ -183,17 +183,120 @@
...
-      (update-last-checked source now))))
+(defn ^:private update-tdr-source
+  "Check for new data in TDR, create new snapshots, insert
+   resulting job creation ids into database and update the
+   timestamp for next time using transaction TX."
+  [source]
+  ;; attempt to snapshot new rows in TDR
+  (let [utc-now-obj     (OffsetDateTime/now (ZoneId/of "UTC"))
+        utc-now         (.format utc-now-obj (DateTimeFormatter/ofPattern "yyyy-MM-dd HH:mm:ss"))
+        row-ids         (vec (find-new-rows source utc-now))
```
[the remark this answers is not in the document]
rexwangcc: As `: datarepo_row_ids` is a `TEXT[]` in DB, the postgres (or JDBC) wants something other than a lazy seq, otherwise it throws. Same reason for why I had to use `doall` in another place. Given what we discussed in https://github.com/broadinstitute/wfl/pull/371#discussion_r633655546, I have moved the `vec` here to the actual place that requires it.

api/src/wfl/module/covid.clj
```diff
@@ -183,17 +183,120 @@
...
+           (jdbc/query tx)))))
+
+(defn ^:private check-tdr-job
+  "Check TDR job status for `job-id`, return a map with job-id,
+   snapshot_id and job_status if job has failed or succeeded, otherwise nil."
+  [job-id]
+  (when-let [job-metadata (datarepo/get-job-metadata-when-done job-id)]
+    (let [{:keys [id job-status] :as result} job-metadata]
+      ;; TODO: catch but not throw at (= job-status "succeeded")
+      (if (= job-status "succeeded")
```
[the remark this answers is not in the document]
rexwangcc: Yeah, I was thinking about your original "compound error` thrown at the end, and realized it requires more work here. I'd prefer we log this failure and our monitoring system (perhaps log-based stackdriver) can pickup by keywords and notify us when this happens.

When it comes to `store the state of the job so we don't always check the failed jobs`, yes, that's why I was struggling to not throw but rather log here. In this case, the `job-metadata` map already stores the failure status which will then be written to the DB. Since `get-pending-tdr-jobs` uses `WHERE snapshot_creation_job_status = 'running'`, we would always only check for running jobs not `succeeded` nor `failed`.

api/src/wfl/module/covid.clj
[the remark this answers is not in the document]
ehigham: nice

api/src/wfl/module/covid.clj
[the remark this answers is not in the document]
ehigham: Right, but are you ever writing 'failed' to the 'snapshot_creation_job_status`?

api/src/wfl/module/covid.clj
```diff
@@ -180,25 +180,135 @@
...
+(defn ^:private update-tdr-source
+  "Check for new data in TDR, create new snapshots, insert
+   resulting job creation ids into database and update the
+   timestamp for next time using transaction TX."
+  [source]
+  ;; attempt to snapshot new rows in TDR
+  (let [utc-now-obj           (OffsetDateTime/now (ZoneId/of "UTC"))
+        utc-now               (.format utc-now-obj (DateTimeFormatter/ofPattern "yyyy-MM-dd HH:mm:ss"))
+        row-ids               (find-new-rows source utc-now)
+        [shards tdr-job-ids]  (create-snapshots source utc-now-obj row-ids)]
```
ehigham, line 305: Can you compose these three into something like the following as these temporary names aren't used anywhere else

```clojure
(defn ^:private find-and-snapshot-new-rows [souce utc-now]
  (let [date-format (DateTimeFormatter/ofPattern "yyyy-MM-dd HH:mm:ss")]
    (->> (.format utc-now date-format)
         (find-new-rows source)
         (create-snapshots souce utc-now))))
```

ehigham: Also, if you're using this for big-query, don't you mean "yyyy-MM-ddTHH:mm:ss"

ehigham: https://console.cloud.google.com/bigquery?authuser=1&project=wfl-dev&ws=!1m5!1m4!4m3!1stdr-prod-covid19-viralseq!2sdatarepo_CDC_COVID19_Surveillance_v1!3sflowcells!1m0&j=bq:us-central1:bquxjob_69e0cf3b_1797b6d1bfa&page=queryresults

api/src/wfl/module/covid.clj
```diff
@@ -180,25 +180,135 @@
...
+  [{:keys [dataset dataset_table] :as _source} now-obj row-ids]
+  (let [shards (partition-all 500 row-ids)
+        compact-now (.format now-obj (DateTimeFormatter/ofPattern "YYYYMMdd'T'HHmmss"))
+        job-ids (doall (map-indexed (fn [idx shard]
+                                      (go-create-snapshot!
+                                       (format "_%s_%s" compact-now idx)
+                                       dataset
+                                       dataset_table
+                                       shard)) shards))]
+    [shards job-ids]))
```
ehigham, line 236:
```clojure
(let [dt-format   (DateTimeFormatter/ofPattern "YYYYMMdd'T'HHmmss")
      compact-now (.format now-obj dt-format)]
  (letfn [(create-snapshot [idx shard]
            [shard (go-create-snapshot! 
                     (format "_%s_%s" compact-now idx)
                      dataset
                      dataset_table
                      shard)])]
    (->> row-ids
        (partition-all 500)
        (map-indexed create-snapshot)
        vec)))
```

api/src/wfl/module/covid.clj
rexwangcc: updated

api/src/wfl/module/covid.clj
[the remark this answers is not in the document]
rexwangcc: updated

api/src/wfl/module/covid.clj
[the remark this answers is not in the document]
rexwangcc: updated

api/src/wfl/module/covid.clj
rexwangcc: I think I will keep the `create-snapshot` a separate function for now, for the convenience of testing, if you don't mind.

PR #376, opened 2021-04-26
[GH-1215] Import snapshots within COVID workload

okotsopoulos:
### Purpose

My component of the COVID update loop: importing TDR snapshots to the workspace.  I consider it safe to merge, but is subject to change if/when the data model evolves further and we connect the pieces of the update loop.

Directly upstream from me:
- [Workflow creation](https://broadinstitute.atlassian.net/browse/GH-1213)
- [Polling for TDR snapshots](https://broadinstitute.atlassian.net/browse/GH-1222)

### Changes

- Data model change: a `snapshot_reference_id` is linked to a workspace and should be associated with the executor, not the source (TDR in our use case).
- `covid/get-imported-snapshot-reference` -  nil or snapshot reference from Rawls for `snapshot_reference_id` in executor details instance (to be created from `TerraExecutorDetails`).
- `covid/import-snapshot!` - import snapshot to workspace, writing to DB if successful
- added integration tests (incomplete coverage)

### Review Instructions

- Robust testing would be easiest with workflow creation fleshed out.  I cannot presently run `covid/import-snapshot!` and write to the database for this reason (more info in my inline comment on `covid_test.clj`).  Is waiting for this functionality the right move?  If not, would appreciate mobbing / pairing to build out a lightweight integration test set-up.
- I pass around objects that I expect will be DB records and destructure them.  We may elect to pass the needed parameters directly.  For now, I err towards overcommunication.

### DIFF
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -11,6 +11,7 @@
             [wfl.service.firecloud :as firecloud]
             [wfl.service.google.storage :as gcs]
             [wfl.service.postgres :as postgres]
+            [wfl.service.rawls :as rawls]
             [wfl.util :as util]
             [wfl.wfl :as wfl]))
 
@@ -26,6 +27,51 @@
 (defn start-covid-workload!
   [])
 
+(defn ^:private snapshot-imported?
+  "Check if TDR snapshot has been imported to Terra Workspace.
+
+  TODOs / QUESTIONS:
+  - Do we need additional / altered arguments?
+  - Should we have access to loaded workload from outer scope (I think so),
+    or should every check in the update loop load the workload anew?
+  - Do we check workload for presence of snapshot ID? How?
+    (I think it could be related to items per aou. items = table name?)
+  - Could one workload be associated with many snapshots?  Are we checking
+    for all if so, or only for a specified snapshot?
+  "
+  [tx {:keys [id] :as workload}]
+  (let [loaded (workloads/load-workload-for-id tx id)]
+    (:snapshot_reference_id loaded)))
+
+(defn ^:private import-snapshot!
+  "Import TDR snapshot to Terra Workspace.
+
+  TODOs / QUESTIONS:
+  - Do we need additional / altered arguments?
+  - How should we name the snapshot reference?
+  - Where do we get workspace and snapshot ID? From workload?
+    From an element within the workload?
+  - How do we store new reference ID in our transaction recording?
+    Associated with the workload, or something else?
+  - Is this a JDBC update or insert?
+  - Error handling?
+  "
+  [tx {:keys [id] :as workload}]
+  (let [name (util/randomize "placeholder-snapshot-ref-name")
+        workspace "workspace from workload?"
+        snapshot-id (:snapshot_id workload) ;; Is this correct, or need to go deeper?
+        reference-id (rawls/create-snapshot-reference workspace snapshot-id name)]
+    (jdbc/update! tx :workload {:snapshot_reference_id reference-id} ["id = ?" id])))
+
+(comment "Proposed workload update loop as described in Spike doc for reference"
+         (cond
+           (workflow-finished?) (mark-as-done!)
+           (submission-launched?) (monitor-workflow!)
+           (snapshot-imported?) (launch-submission!)
+           (snapshot-created?) (import-snapshot!)
+           (snapshot-creation-job-exist?) (monitor-snapshot-status!)
+           :else (check-for-new-data-in-TDR!)))
+
 ;; TODO: implement progressive (private) functions inside this update loop
 (defn update-covid-workload!
   "Use transaction `tx` to batch-update `workload` statuses."
@@ -33,6 +79,11 @@
   (letfn [(update! [{:keys [id] :as workload}]
             (postgres/batch-update-workflow-statuses! tx workload)
             (postgres/update-workload-status! tx workload)
+            ;; Do we wish to save this loaded workload for our
+            ;; to-be-defined workload update loop?
+            ;; All of the checkers want to look at a loaded
+            ;; workflow, and the update loop conditional makes it
+            ;; such that we only do one task per round.
             (workloads/load-workload-for-id tx id))]
     (if (and started (not finished)) (update! workload) workload)))
 
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -13,7 +13,8 @@
             [wfl.service.postgres :as postgres]
             [wfl.service.rawls :as rawls]
             [wfl.util :as util]
-            [wfl.wfl :as wfl]))
+            [wfl.wfl :as wfl])
+  (:import [java.time OffsetDateTime]))
 
 (def pipeline "Sarscov2IlluminaFull")
 
@@ -27,55 +28,36 @@
 (defn start-covid-workload!
   [])
 
-(defn ^:private snapshot-error-map
-  "Return default ExceptionInfo map for snapshot-related exceptions."
-  [{:keys [executor items] :as _workload} {:keys [id snapshot_id] :as _record}]
-  {:workspace (:workspace executor)
-   :table items
-   :id id
-   :snapshot_id snapshot_id})
-
-(defn ^:private try-get-snapshot-reference
-  "Attempt to fetch snapshot reference with id SNAPSHOT_REFERENCE_ID
-  from EXECUTOR-derived workspace."
-  [{:keys [executor] :as workload} {:keys [snapshot_reference_id] :as record}]
-  (let [workspace (:workspace executor)]
-    (try
-      (rawls/get-snapshot-reference workspace snapshot_reference_id)
-      (catch Throwable cause
-        (throw (ex-info "Rawls unable to fetch snapshot reference"
-                        (util/deep-merge
-                         (snapshot-error-map workload record)
-                         {:snapshot_reference_id snapshot_reference_id})
-                        cause))))))
-
 (defn ^:private snapshot-imported?
-  "Check if snapshot with id SNAPSHOT_ID in TDR has been imported to
-  workspace and can be successfully fetched."
-  [workload {:keys [snapshot_id snapshot_reference_id] :as record}]
-  {:pre [(some? snapshot_id)]}
-  (and (some? snapshot_reference_id)
-       (some? (try-get-snapshot-reference workload record))))
+  "Check if snapshot SNAPSHOT_ID in TDR has been imported to WORKSPACE
+  as SNAPSHOT_REFERENCE_ID and can be successfully fetched."
+  [{:keys [snapshot_id snapshot_reference_id] :as _source_details}
+   {:keys [workspace] :as _executor}]
+  (letfn [(fetch [] (rawls/get-snapshot-reference workspace
+                                                  snapshot_reference_id))]
+    (some? (and snapshot_id snapshot_reference_id (util/do-or-nil (fetch))))))
 
 (defn ^:private import-snapshot!
-  "Import snapshot with id SNAPSHOT_ID from TDR to EXECUTOR-derived workspace.
-  Use transaction TX to update ITEMS table with resulting reference id."
-  [tx {:keys [executor items] :as workload} {:keys [id snapshot_id] :as record}]
-  (let [workspace (:workspace executor)
-        reference-id (rawls/create-snapshot-reference workspace snapshot_id)]
-    (when (nil? reference-id)
+  "Import snapshot SNAPSHOT_ID from TDR to WORKSPACE.
+  Use transaction TX to update DETAILS table with resulting reference id.
+  Throw if import fails."
+  [tx
+   {:keys [workload_id] :as _workload}
+   {:keys [details] :as _source}
+   {:keys [details_id snapshot_id] :as source_details}
+   {:keys [workspace] :as _executor}]
+  (letfn [(create! [] (rawls/create-snapshot-reference workspace snapshot_id))]
+    (if-let [reference-id (util/do-or-nil (:referenceId (create!)))]
+      ((jdbc/update! tx
+                     details
+                     {:snapshot_reference_id reference-id}
+                     ["id = ?" details_id])
+       (jdbc/update! tx
+                     :workload
+                     {:updated (OffsetDateTime/now)}
+                     ["id = ?" workload_id]))
       (throw (ex-info "Rawls unable to create snapshot reference"
-                      (snapshot-error-map workload record))))
-    (jdbc/update! tx items {:snapshot_reference_id reference-id} ["id = ?" id])))
-
-(comment "Proposed workload update loop as described in Spike doc for reference"
-         (cond
-           (workflow-finished?) (mark-as-done!)
-           (submission-launched?) (monitor-workflow!)
-           (snapshot-imported?) (launch-submission!)
-           (snapshot-created?) (import-snapshot!)
-           (snapshot-creation-job-exist?) (monitor-snapshot-status!)
-           :else (check-for-new-data-in-TDR!)))
+                      (assoc source_details :workspace workspace))))))
 
 ;; TODO: implement progressive (private) functions inside this update loop
 (defn update-covid-workload!
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -28,36 +28,32 @@
 (defn start-covid-workload!
   [])
 
-(defn ^:private snapshot-imported?
-  "Check if snapshot SNAPSHOT_ID in TDR has been imported to WORKSPACE
-  as SNAPSHOT_REFERENCE_ID and can be successfully fetched."
-  [{:keys [snapshot_id snapshot_reference_id] :as _source_details}
-   {:keys [workspace] :as _executor}]
-  (letfn [(fetch [] (rawls/get-snapshot-reference workspace
-                                                  snapshot_reference_id))]
-    (some? (and snapshot_id snapshot_reference_id (util/do-or-nil (fetch))))))
+(defn ^:private get-imported-snapshot-reference
+  "Nil or the snapshot reference for SNAPSHOT_REFERENCE_ID in WORKSPACE."
+  [{:keys [workspace] :as _executor}
+   {:keys [snapshot_reference_id] :as _executor_details}]
+  (when snapshot_reference_id
+    (util/do-or-nil (rawls/get-snapshot-reference workspace
+                                                  snapshot_reference_id))))
 
 (defn ^:private import-snapshot!
-  "Import snapshot SNAPSHOT_ID from TDR to WORKSPACE.
-  Use transaction TX to update DETAILS table with resulting reference id.
-  Throw if import fails."
+  "Import snapshot with SNAPSHOT_ID to WORKSPACE.
+  Use transaction TX to update DETAILS table with resulting reference id."
   [tx
-   {:keys [workload_id] :as _workload}
-   {:keys [details] :as _source}
-   {:keys [details_id snapshot_id] :as source_details}
-   {:keys [workspace] :as _executor}]
-  (letfn [(create! [] (rawls/create-snapshot-reference workspace snapshot_id))]
-    (if-let [reference-id (util/do-or-nil (:referenceId (create!)))]
-      ((jdbc/update! tx
-                     details
-                     {:snapshot_reference_id reference-id}
-                     ["id = ?" details_id])
-       (jdbc/update! tx
-                     :workload
-                     {:updated (OffsetDateTime/now)}
-                     ["id = ?" workload_id]))
-      (throw (ex-info "Rawls unable to create snapshot reference"
-                      (assoc source_details :workspace workspace))))))
+   workload
+   {:keys [snapshot_id] :as _source_details}
+   {:keys [workspace details] :as _executor}
+   executor_details]
+  (let [refid (-> (rawls/create-snapshot-reference workspace snapshot_id)
+                  :referenceId)]
+    (jdbc/update! tx
+                  details
+                  {:snapshot_reference_id refid}
+                  ["id = ?" (:id executor_details)])
+    (jdbc/update! tx
+                  :workload
+                  {:updated (OffsetDateTime/now)}
+                  ["id = ?" (:id workload)])))
 
 ;; TODO: implement progressive (private) functions inside this update loop
 (defn update-covid-workload!
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -29,17 +29,17 @@
 
 (defn ^:private snapshot-error-map
   "Return default ExceptionInfo map for snapshot-related exceptions."
-  [{:keys [items sink] :as _workload} {:keys [id snapshot_id] :as _record}]
-  {:workspace (:workspace sink)
+  [{:keys [executor items] :as _workload} {:keys [id snapshot_id] :as _record}]
+  {:workspace (:workspace executor)
    :table items
    :id id
    :snapshot_id snapshot_id})
 
 (defn ^:private try-get-snapshot-reference
   "Attempt to fetch snapshot reference with id SNAPSHOT_REFERENCE_ID
-  from SINK-derived workspace."
-  [{:keys [sink] :as workload} {:keys [snapshot_reference_id] :as record}]
-  (let [workspace (:workspace sink)]
+  from EXECUTOR-derived workspace."
+  [{:keys [executor] :as workload} {:keys [snapshot_reference_id] :as record}]
+  (let [workspace (:workspace executor)]
     (try
       (rawls/get-snapshot-reference workspace snapshot_reference_id)
       (catch Throwable cause
@@ -51,17 +51,17 @@
 
 (defn ^:private snapshot-imported?
   "Check if snapshot with id SNAPSHOT_ID in TDR has been imported to
-  SINK-derived workspace and can be successfully fetched."
+  workspace and can be successfully fetched."
   [workload {:keys [snapshot_id snapshot_reference_id] :as record}]
   {:pre [(some? snapshot_id)]}
   (and (some? snapshot_reference_id)
        (some? (try-get-snapshot-reference workload record))))
 
 (defn ^:private import-snapshot!
-  "Import snapshot with id SNAPSHOT_ID from TDR to workspace.
+  "Import snapshot with id SNAPSHOT_ID from TDR to EXECUTOR-derived workspace.
   Use transaction TX to update ITEMS table with resulting reference id."
-  [tx {:keys [items sink] :as workload} {:keys [id snapshot_id] :as record}]
-  (let [workspace (:workspace sink)
+  [tx {:keys [executor items] :as workload} {:keys [id snapshot_id] :as record}]
+  (let [workspace (:workspace executor)
         name (util/randomize "placeholder-snapshot-ref-name")
         reference-id (rawls/create-snapshot-reference workspace snapshot_id name)]
     (when (nil? reference-id)
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -27,41 +27,48 @@
 (defn start-covid-workload!
   [])
 
-(defn ^:private snapshot-imported?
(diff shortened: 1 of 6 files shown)


okotsopoulos:
Failed tests not a result of my changes, ticketed here:

https://broadinstitute.atlassian.net/browse/GH-1291

api/src/wfl/module/covid.clj
```diff
@@ -23,6 +24,61 @@
...
+                         {:snapshot_reference_id snapshot_reference_id})
+                        cause))))))
+
+(defn ^:private snapshot-imported?
+  "Check if snapshot with id SNAPSHOT_ID in TDR has been imported to
+  SINK-derived workspace and can be successfully fetched."
+  [workload {:keys [snapshot_id snapshot_reference_id] :as record}]
+  {:pre [(some? snapshot_id)]}
+  (and (some? snapshot_reference_id)
+       (some? (try-get-snapshot-reference workload record))))
```
okotsopoulos, line 58: Short-circuiting means that we'll try to get the snapshot reference from Rawls only if a `snapshot_reference_Id` is defined.

If it's defined in the DB, we'd hope that we can fetch the snapshot reference successfully… but there might be the possibility that a snapshot is successfully imported in the last run of the update loop and corrupted / deleted during the sleep.

If that happens and the fetch fails, should we…
- Throw the error in `try-get-snapshot-reference` as we do now, halting the processing?
- Log the error but do not throw so the snapshot import retries gracefully?

api/src/wfl/module/covid.clj
```diff
@@ -23,6 +24,61 @@
...
+
+(defn ^:private import-snapshot!
+  "Import snapshot with id SNAPSHOT_ID from TDR to workspace.
+  Use transaction TX to update ITEMS table with resulting reference id."
+  [tx {:keys [items sink] :as workload} {:keys [id snapshot_id] :as record}]
+  (let [workspace (:workspace sink)
+        name (util/randomize "placeholder-snapshot-ref-name")
```
okotsopoulos, line 65: As we know, naming things is hard…

This will be the name of the snapshot reference in the workspace.  It needs to be unique, but AFAIK is only rendered in the UI.  We never pass it to Rawls.

- Any preferences for what this name should be?
- Do we want to punt name generation to `rawls/create-snapshot-reference`?

api/src/wfl/module/covid.clj
```diff
@@ -23,6 +24,61 @@
...
+        reference-id (rawls/create-snapshot-reference workspace snapshot_id name)]
```
okotsopoulos, line 66: If this chokes, I think that an error is thrown and we never get to our more detailed thrown error below.

Is there a Clojure one-liner like Scala's [`Try… getOrElse…`](https://stackoverflow.com/questions/40212369/try-and-getorelse-scala/40212461) that can handle thrown exceptions in assignment gracefully?  Or a better way?

api/src/wfl/module/covid.clj
```diff
@@ -23,6 +24,61 @@
...
+    (when (nil? reference-id)
+      (throw (ex-info "Rawls unable to create snapshot reference"
+                      (util/deep-merge (snapshot-error-map workload record)
+                                       {:name name}))))
+    (jdbc/update! tx items {:snapshot_reference_id reference-id} ["id = ?" id])))
```
okotsopoulos, line 71: Our schema def'n for `Sarscov2IlluminaFull` user-defined type has an `updated` column:
```
            updated                     timestamptz  -- status update time
```

Do we wish to update this value every time we modify row(s) in the table or only when workflow status is updated?

api/src/wfl/module/covid.clj
```diff
@@ -23,6 +24,61 @@
...
+  {:workspace (:workspace sink)
+   :table items
+   :id id
+   :snapshot_id snapshot_id})
+
+(defn ^:private try-get-snapshot-reference
+  "Attempt to fetch snapshot reference with id SNAPSHOT_REFERENCE_ID
+  from SINK-derived workspace."
+  [{:keys [sink] :as workload} {:keys [snapshot_reference_id] :as record}]
+  (let [workspace (:workspace sink)]
```
okotsopoulos, line 42: I repeat this assignment a lot 😬 is there a way to destructure even further to get this element of `sink` above?

api/src/wfl/module/covid.clj
```diff
@@ -23,6 +24,61 @@
...
+  [{:keys [items sink] :as _workload} {:keys [id snapshot_id] :as _record}]
+  {:workspace (:workspace sink)
+   :table items
+   :id id
+   :snapshot_id snapshot_id})
+
+(defn ^:private try-get-snapshot-reference
+  "Attempt to fetch snapshot reference with id SNAPSHOT_REFERENCE_ID
+  from SINK-derived workspace."
+  [{:keys [sink] :as workload} {:keys [snapshot_reference_id] :as record}]
```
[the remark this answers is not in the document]
ehigham: The workspace is given to the executor.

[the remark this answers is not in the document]
okotsopoulos: Thanks, oversight.  Saw a workspace in the sink but overlooked the one in the executor.

api/src/wfl/module/covid.clj
ehigham: In general, you can use `get-in` for this. 
Example:
```clojure
(let [workspace (get-in workload [:sink :workspace])]
  ..)
```

api/src/wfl/module/covid.clj
ehigham: If the snapshot has a snapshot reference, then isn't it already imported?

api/src/wfl/module/covid.clj
ehigham: Can we use the name of the snapshot instead of randomising?

api/src/wfl/module/covid.clj
okotsopoulos: Yes.  But this guards against the (probably low) chance of the snapshot reference being deleted in between import and this check, since they happen in different runs of the update loop.

This may be an overoptimization: I can revert to a simple check for the reference id if so.

api/src/wfl/module/covid.clj
ehigham: Nope - you can write one though:
```clojure
(defn on-error [thunk handler]
  (try
    (thunk)
    (catch Throwable t
      (handler t)))))

(let [reference-id (on-error #(create-snapshot-reference ...) #(add-error-context context %))]
  ..)
````

File: api/src/wfl/module/covid.clj (hunk 2 above)
Comment (okotsopoulos) [reply api/src/wfl/module/covid.clj:65]: Rawls permits us to create multiple snapshot references for the same source snapshot as long as the snapshot reference names are unique.

If we want to allow for this possibility, we can use the snapshot name as the base name with a random suffix.

Comment (okotsopoulos) [reply api/src/wfl/module/covid.clj:65]: Here's an example snapshot created in this manner, where the source snapshot has name "sarscov2_illumina_full_inputs":
```
{:cloningInstructions "COPY_NOTHING",
 :description "",
 :name "sarscov2_illumina_full_inputs78aa97140e6f402dae0ecba77a563a01",
 :reference {:instanceName "terra", :snapshot "7cb392d8-949b-419d-b40b-d039617d2fc7"},
 :referenceId "f7b7eb31-cdc8-4eb4-a42b-18355dc716ee",
 :referenceType "DATA_REPO_SNAPSHOT",
 :workspaceId "e9d053b9-d79f-40b7-b701-904bf542ec2d"}
```

File: api/src/wfl/module/covid.clj [hunk 7]
```diff
@@ -23,6 +24,61 @@
...
+(defn start-covid-workload!
+  [{:keys [submission-id workspace] :as workload}])
+
+(defn ^:private snapshot-error-map
+  "Return default ExceptionInfo map for snapshot-related exceptions."
+  [{:keys [executor items] :as _workload} {:keys [id snapshot_id] :as _record}]
+  {:workspace (:workspace executor)
+   :table items
+   :id id
+   :snapshot_id snapshot_id})
```
tbl3rd, line 36: Why rename `items` to `table` here?
Is `record` a database row?

api/src/wfl/module/covid.clj
```diff
@@ -23,6 +24,61 @@
...
+  [{:keys [executor] :as workload} {:keys [snapshot_reference_id] :as record}]
+  (let [workspace (:workspace executor)]
+    (try
+      (rawls/get-snapshot-reference workspace snapshot_reference_id)
+      (catch Throwable cause
+        (throw (ex-info "Rawls unable to fetch snapshot reference"
+                        (util/deep-merge
+                         (snapshot-error-map workload record)
+                         {:snapshot_reference_id snapshot_reference_id})
+                        cause))))))
```
tbl3rd, line 50: Why catch and rethrow here?
Why `(assoc :snapshot_reference_id …)` here instead of in `snapshot-error-map`?

api/src/wfl/module/covid.clj
```diff
@@ -23,6 +24,61 @@
...
+
+(defn ^:private snapshot-imported?
+  "Check if snapshot with id SNAPSHOT_ID in TDR has been imported to
+  workspace and can be successfully fetched."
+  [workload {:keys [snapshot_id snapshot_reference_id] :as record}]
+  {:pre [(some? snapshot_id)]}
+  (and (some? snapshot_reference_id)
+       (some? (try-get-snapshot-reference workload record))))
```
tbl3rd, line 58: Maybe just try on each pass through the loop, and log failures?
Why `some?`?  Or maybe lift it if you want `false` instead of `nil`?

api/src/wfl/module/covid.clj
```diff
@@ -23,6 +24,61 @@
...
+
+(defn ^:private import-snapshot!
+  "Import snapshot with id SNAPSHOT_ID from TDR to EXECUTOR-derived workspace.
+  Use transaction TX to update ITEMS table with resulting reference id."
+  [tx {:keys [executor items] :as workload} {:keys [id snapshot_id] :as record}]
+  (let [workspace (:workspace executor)
+        name (util/randomize "placeholder-snapshot-ref-name")
```
tbl3rd, line 65: `"placeholder-snapshot-ref-name"` could blame `"wfl"`
to make the source clearer to anyone who finds it later.

api/src/wfl/module/covid.clj
tbl3rd: Clojurists generally just `try` like savages.

api/src/wfl/module/covid.clj
tbl3rd:
```clojure
(util/deep-merge (snapshot-error-map workload record)
                 {:name name})
```
is probably just
```clojure
(assoc (snapshot-error-map workload record) :name name)
```

tbl3rd (commented):
I hope this helps.

api/src/wfl/module/covid.clj
okotsopoulos: Oh nice, I like this better here.

api/src/wfl/module/covid.clj
okotsopoulos: Sounds like this will be imminently deprecated anyway, but my best guess was that `items` stored the name of the COVID instance table.  But now realizing that the renaming makes it harder to track down any issue, so will remove the rename.

I did intend for `record` to be a database row within the COVID instance table, following @rexwangcc 's example (but that also may be deprecated).

api/src/wfl/module/covid.clj
okotsopoulos: That's an interesting thought.  I've since updated the code to generate the snapshot reference name from the TDR snapshot name in `rawls/create-snapshot-reference` but we could supplement that naming to tie it back to `"wfl"` without much trouble.

api/src/wfl/module/covid.clj
okotsopoulos: I caught and rethrew because the original exception from Rawls API didn't share information about the request parameters (that I could see).

I also (try to) use this error map when we create a snapshot reference, if that fails we would not have a `:snapshot_reference_id` available.  But I think I can still put it in the map, since it should be in the underlying table schema.

api/src/wfl/module/covid.clj
okotsopoulos: I'd be interested in talking more with you about this.  This other thread might help provide context:

https://github.com/broadinstitute/wfl/pull/376#discussion_r622657060

api/src/wfl/module/covid.clj
tbl3rd: > Our schema def'n for `Sarscov2IlluminaFull` user-defined type has an `updated` column:
> 
> ```
>             updated                     timestamptz  -- status update time
> ```
> 
> Do we wish to update this value every time we modify row(s) in the table or only when workflow status is updated?

That `status update time` comment may be copypasta because it is on every `updated` column in the schema
For example, in `ContinuousWorkloadInstance` it cannot be just for `status` because there is none in that table.
I think it represents the last time WFL "considered" that workload, and probably should be (ahem) updated
whenever anything in that tree of workload tables is changed -- including workflow status when that applies.

api/src/wfl/module/covid.clj
okotsopoulos: I ended up getting what I needed from `util/do-or-nil`.

api/src/wfl/module/covid.clj
okotsopoulos: Indeed, this is deprecated.

api/src/wfl/module/covid.clj
okotsopoulos: That makes sense to me, thanks for confirming.  I think I've handled this now: changes in the data model meant that I need to update the source details instance with the snapshot refid, *and* the workload table with the updated time.

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -0,0 +1,67 @@
...
+           (let [source-details [sd-base sd-snapshot sd-reference]]
+             (run! go source-details)))))
+    (testing "Snapshot, reference, and successful fetch returns true"
+      (with-redefs-fn
+        {#'rawls/get-snapshot-reference mock-rawls-snapshot-reference}
+        #(is (true? (snapshot-imported? sd-reference executor)))))))
+
+(deftest test-import-snapshot
+  (let [import-snapshot! #'covid/import-snapshot!]
+    #_(testing "Successful create writes to db"
```
okotsopoulos, line 52: I ran into a roadblock trying to test this.

`covid/import-snapshot` updates the active source detail instance within the DB.  This should be created as `TerraDataRepoSourceDetails` during workload creation (not yet in `develop` branch).  I wasn't able to find similar examples in our test code.

I tried adding additional fixtures to create the source detail instance and add row(s) in the state we'd expect at this method call, but ran into issues.  I believe workload creation will make it easier to write this test, so if that functionality is close to completion my preference is to wait for it rather than hammer at this.

Other thoughts welcome.

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -0,0 +1,67 @@
...
+   :name "snapshot",
+   :reference {:instanceName "terra", :snapshot snapshot-id},
+   :referenceId reference-id,
+   :referenceType "DATA_REPO_SNAPSHOT",
+   :workspaceId "e9d053b9-d79f-40b7-b701-904bf542ec2d"})
+
+(defn ^:private mock-throw [& _] (throw (ex-info "mocked throw" {})))
+
+(deftest test-snapshot-imported
+  (let [snapshot-imported? #'covid/snapshot-imported?]
```
okotsopoulos, line 38: Workaround to test private methods.  This works, but LMK if we prefer to go about another way.

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -0,0 +1,67 @@
...
+            [clojure.test :as clj-test]
+            [wfl.tools.fixtures :as fixtures]
+            [wfl.jdbc :as jdbc]
+            [wfl.module.covid :as covid]
+            [wfl.service.rawls :as rawls])
+  (:import [clojure.lang ExceptionInfo]))
+
+(clj-test/use-fixtures :once fixtures/temporary-postgresql-database)
+
+(def workload {:id 1})
```
okotsopoulos, line 12: All of these indiv. `def`s have me thinking that there must be a better way to set up values for testing.  Maybe an `.edn` file?

I'm inclined to wait for the data model to stabilize and see how others use testing data.

api/src/wfl/module/covid.clj
```diff
@@ -27,6 +28,37 @@
...
   [])
 
+(defn ^:private snapshot-imported?
+  "Check if snapshot SNAPSHOT_ID in TDR has been imported to WORKSPACE
+  as SNAPSHOT_REFERENCE_ID and can be successfully fetched."
+  [{:keys [snapshot_id snapshot_reference_id] :as _source_details}
+   {:keys [workspace] :as _executor}]
+  (letfn [(fetch [] (rawls/get-snapshot-reference workspace
+                                                  snapshot_reference_id))]
+    (some? (and snapshot_id snapshot_reference_id (util/do-or-nil (fetch))))))
```
okotsopoulos, line 38: Thanks @tbl3rd for the lesson on `some?`s.

api/src/wfl/module/covid.clj
```diff
@@ -27,6 +28,37 @@
...
+(defn ^:private import-snapshot!
+  "Import snapshot SNAPSHOT_ID from TDR to WORKSPACE.
+  Use transaction TX to update DETAILS table with resulting reference id.
+  Throw if import fails."
+  [tx
+   {:keys [workload_id] :as _workload}
+   {:keys [details] :as _source}
+   {:keys [details_id snapshot_id] :as source_details}
+   {:keys [workspace] :as _executor}]
+  (letfn [(create! [] (rawls/create-snapshot-reference workspace snapshot_id))]
```
okotsopoulos, line 49: I punted naming the snapshot reference to `rawls`, which will generate the name using the TDR snapshot name as its base.

api/src/wfl/module/covid.clj
```diff
@@ -27,6 +28,37 @@
...
+    (if-let [reference-id (util/do-or-nil (:referenceId (create!)))]
+      ((jdbc/update! tx
```
ehigham, line 51: If you want to run a sequence of effects, you can use `do`:
```clojure
(do (jdbc/update! tx details {:snapshot_reference_id reference-id} ["id = ?" details_id])
    (jdbc/update! tx :workload {:updated (OffsetDateTime/now)} ["id = ?" workload_id])))
```

api/src/wfl/module/covid.clj
ehigham: I'm not sure why this needs to do a null check on the snapshot_id. Furthermore, this doesn't check if that snapshot has been imported as that snapshot reference. It just checks if they're both non-nil and that the snapshot reference exists.
Maybe you can just check if the snapshot reference exists?

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -0,0 +1,67 @@
...
+   :workspaceId "e9d053b9-d79f-40b7-b701-904bf542ec2d"})
+
+(defn ^:private mock-throw [& _] (throw (ex-info "mocked throw" {})))
+
+(deftest test-snapshot-imported
+  (let [snapshot-imported? #'covid/snapshot-imported?]
+    (testing "Missing snapshot, missing reference, or failed fetch returns false"
+      (with-redefs-fn
+        {#'rawls/get-snapshot-reference mock-throw}
+        #(letfn [(go [sd] (is (false? (snapshot-imported? sd executor))))]
```
ehigham, line 42: You can inline `#'covid/snapshot-imported?` here - no need to allocate it in the `let` above

api/src/wfl/module/covid.clj
```diff
@@ -27,6 +28,37 @@
...
+  "Import snapshot SNAPSHOT_ID from TDR to WORKSPACE.
+  Use transaction TX to update DETAILS table with resulting reference id.
+  Throw if import fails."
+  [tx
+   {:keys [workload_id] :as _workload}
+   {:keys [details] :as _source}
+   {:keys [details_id snapshot_id] :as source_details}
+   {:keys [workspace] :as _executor}]
+  (letfn [(create! [] (rawls/create-snapshot-reference workspace snapshot_id))]
+    (if-let [reference-id (util/do-or-nil (:referenceId (create!)))]
```
ehigham, line 50: Why are you swallowing the exception here? If there's an error, we probably want to handle it further up the stack?

api/src/wfl/module/covid.clj
ehigham: The snapshot already has a name (it must be named when created). If the executor is given the snapshot object (i.e. that returned by `datarepo/get-snapshot` which contains the name) you can just use that.

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -0,0 +1,67 @@
...
+   :description "test importing a snapshot into a workspace",
+   :name "snapshot",
+   :reference {:instanceName "terra", :snapshot snapshot-id},
+   :referenceId reference-id,
+   :referenceType "DATA_REPO_SNAPSHOT",
+   :workspaceId "e9d053b9-d79f-40b7-b701-904bf542ec2d"})
+
+(defn ^:private mock-throw [& _] (throw (ex-info "mocked throw" {})))
+
+(deftest test-snapshot-imported
```
ehigham, line 37: Is rawls is mocked out here,? If so can you use a dummy workspace and/or snapshot to make it clear that these are arbitrary?

api/test/wfl/integration/modules/covid_test.clj
ehigham: Likewise in your other test

api/test/wfl/integration/modules/covid_test.clj
ehigham: I wouldn't worry too much about it for now. This is easy enough to grep though and change as needed.
I generally move constants shared between tests into the test tools but we can refactor this later.

ehigham: I wouldn't worry too much about it for now. This is easy enough to grep though and change as needed.
I generally move constants shared between tests into the test tools but we can refactor this later.

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -0,0 +1,67 @@
+(ns wfl.integration.modules.covid-test
+  (:require [clojure.test :refer :all]
+            [clojure.test :as clj-test]
+            [wfl.tools.fixtures :as fixtures]
+            [wfl.jdbc :as jdbc]
+            [wfl.module.covid :as covid]
+            [wfl.service.rawls :as rawls])
+  (:import [clojure.lang ExceptionInfo]))
```
tbl3rd, line 8: Docstring? `"Test the Sarscov2IlluminaFull COVID pipeline."`

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -0,0 +1,67 @@
...
+    (testing "Missing snapshot, missing reference, or failed fetch returns false"
+      (with-redefs-fn
+        {#'rawls/get-snapshot-reference mock-throw}
+        #(letfn [(go [sd] (is (false? (snapshot-imported? sd executor))))]
+           (let [source-details [sd-base sd-snapshot sd-reference]]
+             (run! go source-details)))))
+    (testing "Snapshot, reference, and successful fetch returns true"
+      (with-redefs-fn
+        {#'rawls/get-snapshot-reference mock-rawls-snapshot-reference}
+        #(is (true? (snapshot-imported? sd-reference executor)))))))
```
tbl3rd, line 48: Those assertions become just `(is …)` and `(is (not …))` if you take my earlier suggestion.

tbl3rd (approved):
Merging this will help me.

api/test/wfl/integration/modules/covid_test.clj
okotsopoulos: Yes, all rawls calls are mocked.  Wasn't sure if it was overkill to also use a dummy workspace if we never interact with it.  But I can add it to make that behavior clear.

api/src/wfl/module/covid.clj
okotsopoulos: Yes, that's what I'm doing.  See change to `rawls/create-snapshot-reference` which calls `datarepo/get-snapshot` to get the name if none is specified.

api/src/wfl/module/covid.clj
okotsopoulos: @tbl3rd (summary from our Slack conversation): I like this suggestion and am implementing it.  It communicates the same information without throwing out the snapshot reference.

@ehigham yeah, you're right that the `snapshot_id` check is unneeded, I will remove.  Are you okay with me swallowing any error thrown by the fetch in this case?

api/src/wfl/service/rawls.clj
```diff
@@ -23,7 +24,8 @@
 
 (defn create-snapshot-reference
   "Link a Terra Data Repo snapshot with id SNAPSHOT-ID to a fully-qualified
-  Terra WORKSPACE as NAME, optionally described by DESCRIPTION."
+  Terra WORKSPACE as NAME (with the snapshot name as its base if unspecified),
+  optionally described by DESCRIPTION."
   ([workspace snapshot-id name description]
    (-> (workspace-api-url workspace "snapshots")
        (http/post {:headers      (auth/get-auth-header)
```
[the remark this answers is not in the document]
okotsopoulos: >Not sure what (with the snapshot name as its base if unspecified) means.
Does its refer to the snapshot, the name? What might be unspecified?

Good call out, I see how this is unclear.  When creating a snapshot reference, we may specify the reference's name. If unspecified, we generate a unique reference name with the snapshot's name as its base.  Does that clarify?

>This is the rawls namespace, so can we assume the TDR here?
(Not actually sure, but I think so. Are there other varieties of snapshot?)

Yes, I think so too.  I was skittish about distinguishing between snapshots (TDR) and snapshot references (workspace) because the Rawls Swagger docs do not do a great job of distinguishing between them IMO.

api/src/wfl/module/covid.clj
okotsopoulos: I throw an exception if reference creation fails below, but you are right that I swallow the original exception:
 ```
(throw (ex-info "Rawls unable to create snapshot reference"
                      (assoc source_details :workspace workspace)))
```
Would you prefer no fussing with exceptions at this point in the code if we think it will be handled further up?  Or instead of swallowing, I could intercept and rethrow with additional information in the map on the executor, similar to how I approached this earlier on in this PR's life. 

At this point I lean towards the "no fuss" approach until things get stitched together.  Sorry about all the exception back-and-forth :(

File: api/src/wfl/module/covid.clj [hunk 23]
```diff
@@ -27,6 +28,37 @@
...
+      ((jdbc/update! tx
+                     details
+                     {:snapshot_reference_id reference-id}
+                     ["id = ?" details_id])
+       (jdbc/update! tx
+                     :workload
+                     {:updated (OffsetDateTime/now)}
+                     ["id = ?" workload_id]))
+      (throw (ex-info "Rawls unable to create snapshot reference"
+                      (assoc source_details :workspace workspace))))))
```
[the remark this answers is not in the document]
okotsopoulos: >That ((jdbc/update! …)) looks suspect.

What is suspicious about it?  Would it be addressed if I follow this guidance on chaining a sequence of effects?
https://github.com/broadinstitute/wfl/pull/376#discussion_r626169664

Or if the concern is around exception handling, similar concerns / discussion here:
https://github.com/broadinstitute/wfl/pull/376#discussion_r626173102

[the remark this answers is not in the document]
tbl3rd: Yes.  Follow Ed's advice.
That `((` means apply the result of the first`update!` to the result of the second `update!`.
(Probably not what you want to do.)
`let` and `when` have implicit `do`: they take a sequence of expressions and can execute them for side-effects.
Or you can `if` and then use an explicit `do`.  FTR: I like what's there now.

api/src/wfl/service/rawls.clj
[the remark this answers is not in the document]
tbl3rd: What's there is good enough.

PR #377, opened 2021-04-27
Update documentation and infrastructure for release candidates

ehigham:
Update documentation and infrastructure for release candidates
- creating release candiates
- bashing release candidates

Restore version override in cli.py.
Re-tag latest docker images in cli.py so that version can be overwritten.

Update IMAGES target in Makefile to add the :latest tag. This is the image that will be tagged when we run cli.py tag-and-push-images.
Clean up :latest images on distclean target.


okotsopoulos (approved):
Looks good to me.

PR #378, opened 2021-04-27
[GH-1293] fix test-create-submissions-for-entity-set

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1293

Simply getting the submission again is not sufficient to guarantee that
the workflow has been queued for execution. Poll instead.
This is the same issue as https://broadinstitute.atlassian.net/browse/GH-1212

### DIFF
--- a/api/test/wfl/integration/firecloud_test.clj
+++ b/api/test/wfl/integration/firecloud_test.clj
@@ -31,15 +31,14 @@
   (util/bracket
    #(firecloud/create-submission workspace method-configuration entity)
    #(firecloud/abort-submission workspace %)
-   #(let [workflow (-> (firecloud/get-submission workspace %) :workflows first)]
+   #(let [get-workflows (fn [] (:workflows (firecloud/get-submission workspace %)))
+          workflow      (first (get-workflows))]
       (is (= (second entity) (get-in workflow [:workflowEntity :entityName])))
       (is (#{"Queued" "Submitted"}
            ;; GH-1212: `get-submission` can return the workflow before it has
            ;; been queued. In such cases :status is nil so try again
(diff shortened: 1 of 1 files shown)


api/test/wfl/integration/firecloud_test.clj
```diff
@@ -31,15 +31,14 @@
...
       (is (= (second entity) (get-in workflow [:workflowEntity :entityName])))
       (is (#{"Queued" "Submitted"}
            ;; GH-1212: `get-submission` can return the workflow before it has
            ;; been queued. In such cases :status is nil so try again
-           (or (:status workflow) (-> (firecloud/get-submission workspace %)
-                                      :workflows
-                                      first
-                                      :status)))))))
+           (or (:status workflow)
+               (util/poll (comp :status first get-workflows))))))))
```
okotsopoulos, line 41: (Commenting only to confirm my own understanding)

- We pass a function that is itself a composition of functions to `poll` so that `poll` can call it anew for each round.
- If the passed-in function evaluates to false, `poll` will keep trying.  A `nil` status evaluates to boolean false.

okotsopoulos (approved):
Thanks for taking care of this.  I was worried that we might run into a similar issue with `wfl.integration.rawls-test/test-create-snapshot-reference`.  If we do, I feel like I understand `poll` well enough to refactor it myself.

Approving as the two failing tests are existing failures ticketed in [GH-1291](https://broadinstitute.atlassian.net/browse/GH-1291).

api/test/wfl/integration/firecloud_test.clj
ehigham: Exactly

PR #383, opened 2021-05-03
Add multimethods for source, executor and sink operations

ehigham:
This PR is a demonstration for how the source, executor and sink objects could be implemented.
I've added a skeleton implementations for tdr-source


rexwangcc (approved):
LGTM, thanks!

PR #384, opened 2021-05-04
[GH-1295] Use Rawls to Import Workflow Outputs

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1295

firecloud's [flexibleImportEntities](https://firecloud-orchestration.dsde-prod.broadinstitute.org/#/Entities/flexibleImportEntities) has a size limit on the TSV file you POST. Unfortunately, one `sarscov2_illumina_full` workflow's outputs exceeds this. We can work around this issue by using Rawl's [batchUpsert](https://rawls.dsde-prod.broadinstitute.org/#/entities/batch_upsert_entities). This is slightly different as it takes a list of operations on how to construct the entity rather than the entity serialised to TSV.

In this PR, I've demonstrated how we can use Rawls to import the workflow's outputs into the workspace. I've also updated the demo to do this for a workflow that has already passed.
A few questions remain:
- I'm not sure how name the entity we're creating. It should be the `flowcell_id` but this is not a workflow output. I suspect we'll have pass this a workflow input (modify the workflow) and modify the `sink` section of the workflow-request to use that output as the name of the entity.
- In the future, if we are copying to a different workspace we'll need to get the workflow description to fetch the files to copy (see `data-repo-test/test-ingest-pipeline-outputs` for an example of how files are extracted from outputs by traversing the outputs type). I'm not sure how to get this from a method configuration...

UPDATE
I've asked Christine to update the WDL to thread the `flowcell_id` through the WDL so we can pick it up as an output and use it as the primary key in the flowcell table. I don't think this should hold us back from merging this PR.


rexwangcc (approved):
LGTM, also thanks for updating the demo script!

PR #385, opened 2021-05-06
[GH-1215] COVID `update-terra-executor` progress

okotsopoulos:
### Purpose

Main focus was fleshing out `update-terra-executor` to include snapshot import and sketch the update pattern.  As part of this work I got a better understanding of how we intend to use queues to pass items from source to executor and executor to sink.

Full functionality depends on [Rex's PR](https://github.com/broadinstitute/wfl/pull/371) for:
- New snapshot naming convention
- Reintroduction of `snapshot_id` to source details table

### Changes

- Added `update-terra-executor` and required helpers: coerce an available snapshot from `source` queue to a snapshot reference, add entry point for submission creation (@tbl3rd) and write new workflows to executor details instance.
- Refined queue peek and pop for source and executor based on @ehigham 's clarification and mobbing discussion with @rexwangcc .  Notably, queue peeks return a meaningful object (snapshot for source, workflow for executor) rather than a database record.
- Removed now-deprecated code from earlier update loop design.
- Added integration test.

### Review Instructions

- Would appreciate a double-check on how I've generated new record ids when writing to an executor details instance. 
- @tbl3rd - would like to hear your thoughts on `update-terra-executor` submission creation.

### In the future
- Expand `update-terra-executor` to check and update status of active workflows.
- Add unit tests for new helper methods as desired.

### DIFF
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -276,6 +276,15 @@
         (update :fromSource edn/read-string))
     (throw (ex-info "Invalid executor_items" details))))
 
+(defn ^:private import-snapshot-v2!
+  "Import snapshot with SNAPSHOT_ID to WORKSPACE.
+  Update EXECUTOR with resulting reference id."
+  [{:keys [workspace :as executor]} snapshot-id]
+  (let [reference (rawls/create-snapshot-reference workspace snapshot-id)
+        reference-id (:referenceId reference)]
+    (assoc executor :snapshot_reference_id reference-id)
+    reference-id))
+
 (defn ^:private update-terra-executor [source executor]
   "
   IN PROGRESS
@@ -289,28 +298,25 @@
   - Can we assume that source, executor, and sink are loaded with their
     details if applicable?  (Common method?)
   "
-  (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-    (letfn [(find-new-snapshot  [source now]
-              (:snapshotId (peek-queue! source)))
-            (import-snapshot [executor snapshot-id]
-              (-> (rawls/create-snapshot-reference (:workspace executor) snapshot-id)
-                  :referenceId)
-                                      ;; Also update the executor with reference ID, or handle in helper?
-              )
-            (create-submissions  [executor snapshot-reference]
-                                      ;; Tom to implement:
-                                      ;; Create submissions from a snapshot reference.
-                                      ;; Push each submission to executor queue --
-                                      ;; i.e. write to executor (and corresponding details) tables.
-              )
-            (update-last-checked [executor now]
-                                      ;; Update in DB too?  Separate helper?
-              (assoc executor :updated now))]
-      (let [now         (OffsetDateTime/now)
-            snapshot-id (find-new-snapshot source now)]
-        (when snapshot-id
-          (create-submissions executor (import-snapshot executor snapshot-id)))
-        (update-last-checked executor now)))))
+  (letfn [(find-new-snapshot  [source now]
+            (:snapshotId (peek-queue! source)))
+          (import-snapshot [executor snapshot-id]
+            ;; Ultimately may not need this new fn defn:
+            (import-snapshot-v2! executor snapshot-id))
+          (create-submissions  [executor snapshot-reference-id]
+            ;; Tom to implement:
+            ;; Create submissions from a snapshot reference.
+            ;; Push each submission to executor queue --
+            ;; i.e. write to executor (and corresponding details) tables.
+            )
+          (update-last-checked [executor now]
+            ;; Update in DB too?  Separate helper?
+            (assoc executor :updated now))]
+    (let [now         (OffsetDateTime/now)
+          snapshot-id (find-new-snapshot source now)]
+      (when snapshot-id
+        (create-submissions executor (import-snapshot executor snapshot-id)))
+      (update-last-checked executor now))))
 
 (defn ^:private peek-terra-executor-queue [{:keys [queue] :as _executor}]
   (let [query "SELECT * FROM %s WHERE NOT visited ORDER BY id ASC LIMIT 1"]
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -48,10 +48,10 @@
   [{:keys [details] :as queue}]
   (if-let [{:keys [id] :as _item} (peek-queue! queue)]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-        (let [now (OffsetDateTime/now)]
-          (jdbc/update! tx details {:consumed now
-                                    :updated  now}
-                        ["id = ?" id])))
+      (let [now (OffsetDateTime/now)]
+        (jdbc/update! tx details {:consumed now
+                                  :updated  now}
+                      ["id = ?" id])))
     (throw (ex-info "No items in queue" {:queue queue}))))
 
 ;; interfaces
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -259,50 +259,53 @@
                     (->> (jdbc/insert! tx terra-executor-table)))]
     [terra-executor-type (-> items first :id str)]))
 
-(defn import-snapshot!
+(defn ^:private import-snapshot!
   "Return snapshot reference for ID imported to WORKSPACE as NAME."
   [{:keys [workspace] :as _executor}
    {:keys [name id]   :as _snapshot}]
   (rawls/create-snapshot-reference workspace id name))
 
-(defn from-source
+(defn ^:private from-source
   "Coerce SOURCE-ITEM to form understood by EXECUTOR via FROMSOURCE."
   [{:keys [fromSource] :as executor}
    source-item]
   (cond (= "importSnapshot" fromSource) (import-snapshot! executor source-item)
         :else (throw (ex-info "Unknown fromSource" {:executor executor}))))
 
+(defn ^:private create-submission!
+  "TO IMPLEMENT:
+  Create submission from REFERENCE."
+  [executor reference]
+)
+
+(defn ^:private write-workflows!
+  "Write WORKFLOWS to DETAILS table."
+  [{:keys [details]                :as _executor}
+   {:keys [referenceId]            :as _reference}
+   {:keys [submissionId workflows] :as _submission}]
+  (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+    (let [now      (OffsetDateTime/now)
+          id-start (inc (postgres/table-max tx details :id))
+          to-rows  (fn [idx {:keys [status workflowId] :as _workflow}]
+                     {:id                    (+ id-start idx)
+                      :snapshot_reference_id referenceId
+                      :rawls_submission_id   submissionId
+                      :workflow_id           workflowId
+                      :workflow_status       status
+                      :updated               now})]
+      (jdbc/insert-multi! tx details (map-indexed to-rows workflows)))))
+
 (defn ^:private update-terra-executor
   "Create new submission from new SOURCE snapshot if available,
   writing its workflows to DETAILS.
   Return EXECUTOR."
-  [source
-   {:keys [details] :as executor}]
-  (letfn [(create-submission! [executor reference]
-            ;; Tom to implement:
-            ;; Create submission from snapshot reference.
-            )
-          (write-workflows!
-            [{:keys [referenceId]            :as _reference}
-             {:keys [submissionId workflows] :as _submission}
-             now]
-            (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-              (->> workflows
-                   (map (fn [{:keys [status workflowId] :as _workflow}]
-                          ;; FIXME: must also specify id:
-                          ;; no autoincrement for TABLEs created from TYPEs.
-                          {:snapshot_reference_id referenceId
-                           :rawls_submission_id   submissionId
-                           :workflow_id           workflowId
-                           :workflow_status       status
-                           :updated               now}))
-                   (jdbc/insert-multi! tx details))))]
-    (if-let [snapshot (peek-queue! source)]
-      (let [reference  (from-source executor snapshot)
-            submission (create-submission! executor reference)]
-        (write-workflows! reference submission (OffsetDateTime/now))
-        (pop-queue! source)))
-    executor))
+  [source executor]
+  (if-let [snapshot (peek-queue! source)]
+    (let [reference  (from-source executor snapshot)
+          submission (create-submission! executor reference)]
+      (write-workflows! executor reference submission)
+      (pop-queue! source)))
+  executor)
 
 (defn ^:private load-terra-executor [tx {:keys [executor_items] :as details}]
   (if-let [id (util/parse-int executor_items)]
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -275,8 +275,7 @@
 (defn ^:private create-submission!
   "TO IMPLEMENT:
   Create submission from REFERENCE."
-  [executor reference]
-)
+  [executor reference])
 
 (defn ^:private write-workflows!
   "Write WORKFLOWS to DETAILS table."
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -308,7 +308,7 @@
             )]
     (let [now      (OffsetDateTime/now) ; Use to update... something?
           snapshots (find-new-snapshots source)]
-      (when (seq? snapshots)
+      (when (seq snapshots)
         (create-submissions executor (import-snapshots! executor snapshots))))))
 
 (defn ^:private peek-terra-executor-queue [{:keys [queue] :as _executor}]
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -35,20 +35,23 @@
     record))
 
 (defn peek-queue!
-  "Get first unconsidered item in DETAILS table."
+  "Get first unconsumed item in DETAILS table."
   [{:keys [details] :as _queue}]
-  (let [query "SELECT * FROM %s WHERE NOT considered ORDER BY id ASC LIMIT 1"]
+  (let [query "SELECT * FROM %s WHERE NOT consumed ORDER BY id ASC LIMIT 1"]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
       (->> (format query details)
            (jdbc/query tx)
            (first)))))
 
 (defn pop-queue!
-  "Consider first unconsidered item in DETAILS table, or throw if empty."
+  "Consume first unconsumed item in DETAILS table, or throw if empty."
   [{:keys [details] :as queue}]
   (if-let [{:keys [id] :as _item} (peek-queue! queue)]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-      (jdbc/update! tx details {:considered true} ["id = ?" id]))
+        (let [now (OffsetDateTime/now)]
+          (jdbc/update! tx details {:consumed now
+                                    :updated  now}
+                        ["id = ?" id])))
     (throw (ex-info "No items in queue" {:queue queue}))))
 
 ;; interfaces
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -285,22 +285,21 @@
             ;; Create submission from snapshot reference.
             )
           (write-workflows!
-            [tx
-             {:keys [referenceId]            :as _reference}
+            [{:keys [referenceId]            :as _reference}
              {:keys [submissionId workflows] :as _submission}]
-            (->> workflows
-                 (map (fn [{:keys [status workflowId] :as _workflow}]
-                        {:snapshot_reference_id referenceId
-                         :rawls_submission_id   submissionId
-                         :workflow_id           workflowId
-                         :workflow_status       status
-                         :updated               (OffsetDateTime/now)}))
-                 (jdbc/insert-multi! tx details)))]
+            (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+              (->> workflows
+                   (map (fn [{:keys [status workflowId] :as _workflow}]
+                          {:snapshot_reference_id referenceId
+                           :rawls_submission_id   submissionId
+                           :workflow_id           workflowId
+                           :workflow_status       status
+                           :updated               (OffsetDateTime/now)}))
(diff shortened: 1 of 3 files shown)


api/src/wfl/module/covid.clj
```diff
@@ -180,3 +181,82 @@
...
+  (letfn [(check-request! [request] "Rhian to implement?")
+          (make-record    [request] "Rhian / Rex to implement?")]
+    (check-request! executor-request)
+    (let [foreign-key (make-record executor-request)]
+      (load-terra-executor {:executor-ptr foreign-key}))))
+
+(defn ^:private import-snapshot-v2!
+  "Import snapshot with SNAPSHOT_ID to WORKSPACE.
+  Update EXECUTOR with resulting reference id."
+  [{:keys [workspace :as executor]} snapshot-id]
```
ehigham, line 208:
```clojure
{:keys [workspace] :as executor}
```

api/src/wfl/module/covid.clj
```diff
@@ -180,3 +153,79 @@
...
+  (letfn [(check-request! [request] "Rhian to implement?")
+          (make-record    [request] "Rhian / Rex to implement?")]
+    (check-request! executor-request)
+    (let [foreign-key (make-record executor-request)]
+      (load-terra-executor {:executor-ptr foreign-key}))))
+
+(defn ^:private find-new-snapshots
+  "Return source details records for first member of SOURCE queue."
+  [source]
+  (if-let [source-entry (peek-queue! source)]
```
okotsopoulos, line 181: Another option is to bake in the details fetch into `peek-queue!`.

api/src/wfl/module/covid.clj
```diff
@@ -58,16 +31,24 @@
...
-(defmulti peek-queue!
-  "Peek the first object from the `queue`, if one exists."
-  (fn [queue] (:type queue)))
-
-(defmulti pop-queue!
-  "Pop the first object from the `queue`. Throws if none exists."
-  (fn [queue] (:type queue)))
+(defn peek-queue!
+  "Get first unconsidered item in DETAILS table."
+  [{:keys [details] :as _queue}]
```
okotsopoulos, line 36: I found that the queue operations didn't differ based on whether we were checking a source, executor, or sink queue.

As defined, these operations are specific to our model where a `details` table is the queue to check.  We could switch this back to a more generic `queue` check but would need to supplement source/executor/sink with this key-val at load.  Or we accept instead the table to monitor rather than a destructured map.  I think this works for now but can change if there are strong feelings.

api/src/wfl/module/covid.clj
```diff
@@ -263,6 +236,35 @@
...
+            ;; Tom to implement?
+            ;; Create submission from snapshot reference.
+            )
+          (push-to-executor! [source executor submission]
+            (pop-queue! source)
+            ;; Write new records to executor queue (:details executor)
+            ;; for each workflow in submission.
+            )
+          (update-executor! [executor]
+            (assoc executor :updated (OffsetDateTime/now)))]
```
okotsopoulos, line 261: Not sure if we want to do this anymore, we may be happy with setting `:updated` on the queue objects.

api/src/wfl/module/covid.clj
```diff
@@ -58,16 +31,27 @@
...
-  "Pop the first object from the `queue`. Throws if none exists."
-  (fn [queue] (:type queue)))
+(defn peek-queue!
+  "Get first unconsumed item in DETAILS table."
+  [{:keys [details] :as _queue}]
+  (let [query "SELECT * FROM %s WHERE NOT consumed ORDER BY id ASC LIMIT 1"]
+    (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+      (->> (format query details)
+           (jdbc/query tx)
+           (first)))))
```
tbl3rd: `WHERE consumed IS NULL`?  Or does SQL agree with Clojure on truthiness?

api/src/wfl/module/covid.clj
```diff
@@ -58,16 +31,27 @@
...
+(defn pop-queue!
+  "Consume first unconsumed item in DETAILS table, or throw if empty."
+  [{:keys [details] :as queue}]
+  (if-let [{:keys [id] :as _item} (peek-queue! queue)]
+    (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+      (let [now (OffsetDateTime/now)]
+        (jdbc/update! tx details {:consumed now
+                                  :updated  now}
+                      ["id = ?" id])))
+    (throw (ex-info "No items in queue" {:queue queue}))))
```
tbl3rd, line 50: SELECT and UPDATE are in different transactions,
which is _probably_ OK here.

api/src/wfl/module/covid.clj
```diff
@@ -195,24 +183,22 @@
...
-    (throw (ex-info "No snapshots in queue" {:source source}))))
+(defn ^:private create-tdr-source [tx id request]
+  (let [create  "CREATE TABLE %s OF TerraDataRepoSourceDetails (PRIMARY KEY (id))"
+        details (format "TerraDataRepoSourceDetails_%09d" id)
+        _       (jdbc/execute! tx [(format create details)])
+        items   (-> (select-keys request (keys tdr-source-serialized-fields))
+                    (set/rename-keys tdr-source-serialized-fields)
+                    (assoc :details details)
+                    (->> (jdbc/insert! tx tdr-source-table)))]
+    [tdr-source-type (-> items first :id)]))
```
tbl3rd, line 192: Suggestion:
```clojure
(defn ^:private create-tdr-source
  "Use TX and REQUEST to create a TerraDataRepoSourceDetails table named
  with ID.  Return the resulting table's ID."
  [tx id request]
  (let [create  "CREATE TABLE %s OF TerraDataRepoSourceDetails (PRIMARY KEY (id))"
        details (format "TerraDataRepoSourceDetails_%09d" id)]
    (jdbc/execute! tx [(format create details)])
    [tdr-source-type (-> request
                         (select-keys (keys tdr-source-serialized-fields))
                         (set/rename-keys tdr-source-serialized-fields)
                         (assoc :details details)
                         (->> (jdbc/insert! tx tdr-source-table))
                         first :id)]))
```
I notice this isn't really new though.

api/src/wfl/module/covid.clj
```diff
@@ -226,19 +212,7 @@
...
 
-(defn ^:private create-tdr-source [tx id request]
-  (let [create  "CREATE TABLE %s OF TerraDataRepoSourceDetails (PRIMARY KEY (id))"
-        details (format "TerraDataRepoSourceDetails_%09d" id)
-        _       (jdbc/execute! tx [(format create details)])
-        items   (-> (select-keys request (keys tdr-source-serialized-fields))
-                    (set/rename-keys tdr-source-serialized-fields)
-                    (assoc :details details)
-                    (->> (jdbc/insert! tx tdr-source-table)))]
-    [tdr-source-type (-> items first :id)]))
```
tbl3rd, line 237: Ah.

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -2,75 +2,11 @@
   "Test the Sarscov2IlluminaFull COVID pipeline."
   (:require [clojure.test :refer :all]
             [clojure.string :as str]
-            [wfl.debug :as debug]
-            [wfl.jdbc :as jdbc]
```
tbl3rd, line 6: Thanks.

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -2,75 +2,11 @@
...
-        #_(testing "Successful create writes to db"
-            (jdbc/with-db-transaction [tx (fixtures/testing-db-config)]
-              (with-redefs-fn {#'rawls/create-snapshot-reference mock-rawls-snapshot-reference}
-                #(#'covid/import-snapshot! tx workload source-details executor ed-base))))
-        (testing "Failed create throws"
-          (jdbc/with-db-transaction [tx (fixtures/testing-db-config)]
-            (with-redefs-fn {#'rawls/create-snapshot-reference mock-throw}
-              #(is (thrown-with-msg?
-                    ExceptionInfo #"mocked throw"
-                    (#'covid/import-snapshot! tx workload source-details executor ed-base))))))))))
```
tbl3rd, line 73: These multis calling multis in the same module confuse me,
but anticipate our tests will sort it out.

tbl3rd (approved):

api/src/wfl/module/covid.clj
```diff
@@ -263,6 +237,35 @@
                     (->> (jdbc/insert! tx terra-executor-table)))]
     [terra-executor-type (-> items first :id)]))
 
+(defn import-snapshot!
+  "Return snapshot reference for SNAPSHOT_ID imported to WORKSPACE."
+  [{:keys [workspace] :as _executor}
+   {:keys [snapshot_id] :as _snapshot}]
+  (rawls/create-snapshot-reference workspace snapshot_id))
```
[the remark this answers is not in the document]
ehigham: Note we need the name to update the method config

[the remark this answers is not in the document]
okotsopoulos: I'll see what Rex did in his open PR with appending the timestamp instead of a random suffix.  I'd be happier with that middle ground if it's feasible.

api/src/wfl/module/covid.clj
okotsopoulos: Ugh.  Yes, thanks for this sanity check.  Will fix.

api/src/wfl/module/covid.clj
okotsopoulos: Thanks for catching.  This is a holdover from when we thought we'd store this value as a boolean: 

https://www.postgresqltutorial.com/postgresql-boolean/

Will fix.

api/test/wfl/integration/modules/covid_test.clj
okotsopoulos: I'll actually be simplifying this multi-ception:
https://github.com/broadinstitute/wfl/pull/385#discussion_r631383843

api/src/wfl/module/covid.clj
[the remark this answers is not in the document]
okotsopoulos: Also, because this returns the whole snapshot reference, downstream functions will have access to the snapshot reference name.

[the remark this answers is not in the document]
ehigham: Great.

api/src/wfl/module/covid.clj
okotsopoulos: We could change the queue methods to accept transactions, but I think that's an okay change to save for later.

api/src/wfl/module/covid.clj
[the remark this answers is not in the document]
okotsopoulos: Closing the loop, I did end up using the name that the snapshot was created with as you originally requested.  When Rex's changes are merged, that name will have the formatted timestamp as its suffix, so a double timestamp suffix would be overkill.

api/src/wfl/module/covid.clj
```diff
@@ -226,21 +197,35 @@
...
-                    (assoc :details details)
-                    (->> (jdbc/insert! tx tdr-source-table)))]
-    [tdr-source-type (-> items first :id)]))
+(defn ^:private peek-tdr-source-queue
+  "Get first unconsumed snapshot in DETAILS table."
+  [{:keys [details] :as _source}]
+  (let [query "SELECT *
+               FROM %s
+               WHERE consumed IS NULL
+               AND snapshot_id IS NOT NULL
```
okotsopoulos, line 206: Flagging for @rexwangcc -- to be peekable, a member of the source queue (record in the `TerraDataRepoSourceDetails` instance) should have a defined `snapshot_id`.

From mobbing today, I'll rely on `snapshot_id` being defined for eligible records within your `update-tdr-source` loop.  Thank you!

api/src/wfl/module/covid.clj
```diff
@@ -226,21 +197,35 @@
...
+               ORDER BY id ASC LIMIT 1"]
+    (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+      (->> (format query details)
+           (jdbc/query tx)
+           first))))
+
+(defn ^:private pop-tdr-source-queue
```
okotsopoulos, line 213: While peek behavior changes based off of whether the queue is for source, executor, or sink, I think that popping could be consolidated to a single method without dispatching.  Will hold off on this for now though.

api/src/wfl/module/covid.clj
```diff
@@ -263,6 +248,49 @@
...
+  [{:keys [workspace]   :as _executor}
+   {:keys [snapshot_id] :as _snapshot}]
+  (rawls/create-snapshot-reference workspace snapshot_id))
+
+(defn from-source
+  "Coerce SOURCE-ITEM to form understood by EXECUTOR via FROMSOURCE."
+  [{:keys [fromSource] :as executor}
+   source-item]
+  (cond (= "importSnapshot" fromSource) (import-snapshot! executor source-item)
+        :else (throw (ex-info "Unknown fromSource" {:executor executor}))))
```
okotsopoulos, line 262: Flagging for @rhiananthony -- this validation should ideally take place inside `verify-executor!`.

api/src/wfl/module/covid.clj
```diff
@@ -263,6 +248,49 @@
...
+
+(defn ^:private update-terra-executor
```
okotsopoulos, line 264: Right now, this method may trigger a submission from a coerced member of the `source` queue.  We also want it to check the status for any active workflows in the `executor` queue and update them if necessary.  This is similar to how @rexwangcc is expanding `update-tdr-source`.

I'd like to save this for a future PR.

api/src/wfl/service/rawls.clj
```diff
@@ -37,7 +37,7 @@
   ([workspace snapshot-id name]
    (create-snapshot-reference workspace snapshot-id name ""))
   ([workspace snapshot-id]
-   (let [name (util/randomize (:name (datarepo/snapshot snapshot-id)))]
+   (let [name (:name (datarepo/snapshot snapshot-id))]
```
okotsopoulos, line 40: Expects changes in @rexwangcc 's [PR](https://github.com/broadinstitute/wfl/pull/371) to name TDR snapshots something like:
```
<snapshotBaseName>_<YYYYMMDD>T<HHmmSS>_<shardIndex>
```

This serves as a realistically unique name.  Known possibility for collision: importing the same snapshot as a reference multiple times (we wouldn't do this).

api/src/wfl/module/covid.clj
ehigham: Leave it as a multimethod. I'm using this for testing. If a bunch of instances share the same implementation, they should delegate.

api/src/wfl/module/covid.clj
ehigham: Ideally "this shouldn't happen" but doesn't hurt to guard against illegal states

api/src/wfl/module/covid.clj
```diff
@@ -271,8 +298,36 @@
         (update :fromSource edn/read-string))
     (throw (ex-info "Invalid executor_items" details))))
 
+(defn ^:private peek-terra-executor-queue
```
ehigham, line 301: Can your peek implementation return the workflow instead of the workflow id? Add a line to fetch the workflow from firecloud? This is because the executor is viewed as "a queue of succeeded workflows" by the sink.

api/src/wfl/service/rawls.clj
ehigham: It would throw so this can't happen.

ehigham (commented):
Please add an integration test at least.
You can see a way to mock the source queue in my PR.

api/src/wfl/module/covid.clj
okotsopoulos: To be clear, your sink implementation expects and would use the full workflow?  Because we call the `peek` within the `pop`, this change would result in two external calls with one of them unused.  Any appetite instead for calling the fetch from within your update?

For the analogous `peek-tdr-source-queue` implementation, my executor update loop would not make use of the full submission at any point, so I would leave that as is.

ehigham: > To be clear, your sink implementation expects and would use the full workflow? 
Yes

> Because we call the peek within the pop, this change would result in two external calls with one of them unused.
Then implement pop without peek?

> Any appetite instead for calling the fetch from within your update?
No.

okotsopoulos: Okay.

> What do submissions have to do with the implementation of the source queue?

I mistyped and meant snapshot, not submission, corrected above.  I know that you and Rex spoke yesterday about adding the full snapshot either to the source details table or the `peek-tdr-source-queue` implementation.  I am stating explicitly here that I do not need the full snapshot, because this situation is similar and I want to call out the difference in approach as a choice and not an oversight.

ehigham: I'm aware - you just need the ID and the name.

ehigham: We can either store the snapshot in the database or call the datarepo on peek.

okotsopoulos: Fulfilled your request and broke up the peek code so that we don't trigger an unneeded external call when popping, resolving this convo but LMK if you need something else here.

api/src/wfl/module/covid.clj
```diff
@@ -229,21 +202,41 @@
...
+  [source]
+  (if-let [{:keys [snapshot_id] :as _record} (peek-tdr-source-details source)]
+    (datarepo/snapshot snapshot_id)))
+
+(defn ^:private pop-tdr-source-queue
+  "Consume first unconsumed snapshot record in DETAILS table, or throw if none."
+  [{:keys [details] :as source}]
+  (if-let [{:keys [id] :as _record} (peek-tdr-source-details source)]
+    (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+      (let [now (OffsetDateTime/now)]
```
rexwangcc, line 229: It looks like we use vanilla `(OffsetDateTime/now)` across the DB updates? If so, I will update my code wherever used:
```
(.format (OffsetDateTime/now (ZoneId/of "UTC")) (DateTimeFormatter/ofPattern "yyyy-MM-dd HH:mm:ss"))
```
to `(OffsetDateTime/now)` too.

api/src/wfl/module/covid.clj
```diff
@@ -198,24 +173,22 @@
...
-      (first (jdbc/query tx (format query queue))))))
-
-(defn ^:private pop-tdr-source-queue [{:keys [queue] :as source}]
-  (if-let [{:keys [id] :as _snapshot} (peek-queue! source)]
-    (let [query "DELETE FROM %s WHERE id = ?"]
-      (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-        (jdbc/execute! tx [(format query queue) id])))
-    (throw (ex-info "No snapshots in queue" {:source source}))))
+(defn ^:private create-tdr-source [tx id request]
+  (let [create  "CREATE TABLE %s OF TerraDataRepoSourceDetails (PRIMARY KEY (id))"
```
rexwangcc, line 177: For my own understanding, I'm not sure if this will create the `id` field with auto increment feature (e.g. `SERIAL` or `GENERATED AS IDENTITY`), since it's not defined in our `CREATE TYPE TerraDataRepoSourceDetails` composition type?

tbl3rd: You could try an `"ALTER TABLE %s ALTER COLUMN id ADD GENERATED ALWAYS AS IDENTITY"`
after the `"CREATE TABLE OF %s"`.  That might do the trick.
See https://www.postgresql.org/docs/11/sql-altertable.html

api/src/wfl/service/rawls.clj
rexwangcc: Just a minor update, I corrected the timestamp to `<snapshotBaseName>_<YYYYMMdd>T<HHmmss>_<shardIndex>`.

api/src/wfl/module/covid.clj
okotsopoulos: That's how I've seen it used throughout, but if we want to enforce a different way we can break it out somewhere common.

api/src/wfl/module/covid.clj
okotsopoulos: I'll throw out there that I moved this code block but I didn't write it initially, tagging in @ehigham .

(Lesson learned: touch not what you don't understand / aren't willing to fix :) )

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -158,4 +162,5 @@
                (fn [] (seq (firecloud/list-entities workspace "flowcell"))))]
           (is (= name flowcell-id) "The test entity was not created"))))))
 
-(test-vars [#'test-update-terra-workspace-sink])
+(test-vars [#'test-update-terra-executor
```
ehigham, line 165: I left this in by mistake!

api/src/wfl/module/covid.clj
```diff
@@ -266,6 +259,53 @@
...
+  [executor reference])
+
+(defn ^:private write-workflows!
+  "Write WORKFLOWS to DETAILS table."
+  [{:keys [details]                :as _executor}
+   {:keys [referenceId]            :as _reference}
+   {:keys [submissionId workflows] :as _submission}]
+  (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+    (let [now      (OffsetDateTime/now)
+          id-start (inc (postgres/table-max tx details :id))
```
okotsopoulos, line 287: Want to flag how I've generated ids here, since we could be writing multiple rows (one submission can trigger many workflows).

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -119,6 +84,45 @@
...
+    (fn [workspace]
+      (let [snapshot {:name "test-snapshot-name"
+                      :id   (str (UUID/randomUUID))}
+            source (make-queue-from-list [snapshot])
+            executor (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+                       (->> {:name                       "Terra"
+                             :workspace                  workspace
+                             :methodConfiguration        "mc-namespace/mc-name"
+                             :methodConfigurationVersion 1
+                             :entity                     "snapshot"
```
ehigham, line 99: We don't need the entity anymore I don't think?

ehigham (approved):
This looks great. Thank you.

api/test/wfl/integration/modules/covid_test.clj
okotsopoulos: You're right, I pulled it from the spike doc but we never use it, will clean up!

api/test/wfl/integration/modules/covid_test.clj
okotsopoulos: Ahh!  Was going to ask about it, will remove.

PR #386, opened 2021-05-07
[GH-1301] Notify Workload Watchers when "User Visible" Exceptions Occur

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1301
Introduce `wfl.util.UserVisibleException` - a new type of exception that we should handle to throw and handle errors that users are meant to see.
Re-wire the background loop to handle `UserVisibleException` and also be more exception safe. Start to plumb in email notifications.
Awaiting https://github.com/broadinstitute/wfl/pull/387 as there's some duplicate DB plumbing.


api/src/wfl/jdbc.clj
```diff
@@ -135,3 +137,20 @@
...
+;; https://stackoverflow.com/a/25786990
+(extend-protocol clojure.java.jdbc/ISQLParameter
+  IPersistentVector
+  (set-parameter [v ^PreparedStatement stmt ^long i]
+    (let [conn          (.getConnection stmt)
+          meta          (.getParameterMetaData stmt)
+          [head & rest] (.getParameterTypeName meta i)]
+      (if-let [elem-type (when (= head \_) (apply str rest))]
+        (.setObject stmt i (.createArrayOf conn elem-type (to-array v)))
+        (.setObject stmt i v)))))
```
tbl3rd, line 151: Weird.  So `_` introduces a type name?

tbl3rd (approved):
Cool.  Was hoping I could watch this work in a test.
Maybe `UserException` is enough of a name?

api/src/wfl/jdbc.clj
ehigham: > Cool. Was hoping I could watch this work in a test.
> Maybe `UserException` is enough of a name?

I'm just introducing the type - what would you have the test verify?
Rex has tests for the protocol extensions in his PR so I didn't bother as I'm hoping he'll merge soon. https://github.com/broadinstitute/wfl/pull/371

PR #393, opened 2021-05-13
GH-1417: Remove the "skipped" nil UUID Cromwell status hack.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1417

### Changes

- Remove the synthetic string `"skipped"` status from cromwell support.
- Turns out `uuid-nil?` and so on supported only `"skipped"` so that's gone too.
- Fuss over specs a bit.
- Gleaned some lint that I noticed collecting.

### Review Instructions

- Ignore this for now because Ed doesn't want it yet.
- Ed says its OK to review this now.
- Diff this branch against `develop`?


ehigham (approved):
Thanks.

PR #396, opened 2021-05-14
[GH-1306] Add `workflows` multimethod for workloads

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1306
Add workflows multimethod
Change all locations where workflows are accessed via the `:workflows` keyword to use this function.
Implement `workflows` for COVID
- I've added a simple test to make sure the SQL is correct
- Awaiting executor implementation to test with more than 0 workflows.

### DIFF
--- a/api/src/wfl/module/batch.clj
+++ b/api/src/wfl/module/batch.clj
@@ -1,13 +1,77 @@
 (ns wfl.module.batch
   "Some utilities shared between batch workloads in cromwell."
-  (:require [wfl.api.workloads :as workloads]
-            [wfl.jdbc :as jdbc]
+  (:require [clj-http.client      :as http]
+            [clojure.string       :as str]
+            [wfl.api.workloads    :as workloads]
+            [wfl.auth             :as auth]
+            [wfl.jdbc             :as jdbc]
             [wfl.service.cromwell :as cromwell]
             [wfl.service.postgres :as postgres]
-            [wfl.util :as util]
-            [wfl.wfl :as wfl])
+            [wfl.util             :as util]
+            [wfl.wfl              :as wfl])
   (:import [java.time OffsetDateTime]
            [java.util UUID]))
+
+(defn ^:private cromwell-status
+  "`status` of the workflow with `uuid` in `cromwell`."
+  [cromwell uuid]
+  (-> (str/join "/" [cromwell "api" "workflows" "v1" uuid "status"])
+      (http/get {:headers (auth/get-auth-header)})
+      :body
+      util/parse-json
+      :status))
+
+(def ^:private finished?
+  "Test if a workflow `:status` is in a terminal state."
+  (set (conj cromwell/final-statuses "skipped")))
+
+(defn make-update-workflows [get-status!]
+  (fn [tx {:keys [items] :as workload}]
+    (letfn [(update! [{:keys [id uuid]} status]
+              (jdbc/update! tx items
+                            {:status status
+                             :updated (OffsetDateTime/now)
+                             :uuid uuid}
+                            ["id = ?" id]))]
+      (->> (workloads/workflows workload)
+           (remove (comp nil? :uuid))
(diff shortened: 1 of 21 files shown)


api/src/wfl/module/covid.clj
```diff
@@ -346,11 +351,21 @@
...
+         (->> (format "SELECT DISTINCT rawls_submission_id FROM %s" details)
+              (jdbc/query tx)))
+       (mapcat (comp :workflows #(firecloud/get-submission workspace %)))))
+
 (defoverload create-executor! terra-executor-name create-terra-executor)
 (defoverload update-executor! terra-executor-type update-terra-executor)
 (defoverload load-executor!   terra-executor-type load-terra-executor)
 (defoverload peek-queue!      terra-executor-type peek-terra-executor-queue)
 (defoverload pop-queue!       terra-executor-type pop-terra-executor-queue)
+(defoverload executor-workflows terra-executor-type terra-executor-workflows)
```
okotsopoulos, line 368: Awesome, I look forward to using this for [GH-1315](https://broadinstitute.atlassian.net/browse/GH-1315).

api/src/wfl/module/batch.clj
```diff
@@ -30,6 +94,18 @@
     (jdbc/update!        tx :workload {:items table} ["id = ?" id])
     [id table]))
 
+(defn pre-v0.4.0-load-workload-impl
```
okotsopoulos, line 97: What is this function's eventual fate, guessing it's here for backwards compatibility?  Is there a known condition that would allow us to delete it?

api/src/wfl/module/covid.clj
```diff
@@ -255,8 +255,10 @@
 
 (defn ^:private create-terra-executor [tx id request]
   (let [create  "CREATE TABLE %s OF TerraExecutorDetails (PRIMARY KEY (id))"
+        alter   "ALTER TABLE %s ALTER COLUMN id ADD GENERATED ALWAYS AS IDENTITY"
```
okotsopoulos, line 258: Thanks.

api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -85,78 +81,102 @@
...
+          (is (every? #(= submission-id (:rawls_submission_id %)) records))
+          (is (every? #(nil? (:consumed %)) records))
+          (verify-record-against-workflow running-record running-workflow 1)
+          (verify-record-against-workflow succeeded-record succeeded-workflow 2))))))
+
+(deftest test-peek-terra-executor-queue
+  (let [succeeded? #{"Succeeded"}
+        source     (make-queue-from-list [{:name "test-snapshot-name"
+                                           :id   (str (UUID/randomUUID))}])
+        executor   (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
```
okotsopoulos, line 121: Thanks for adding this test as well.  Nit: might be good to break out this assignment into a helper since we use it in the above test.

api/src/wfl/module/batch.clj
ehigham: yes I think there is one.
In the future I propose that we add a "InputsListSource" - a source that's literally a list of workflow inputs:
```json
...
"source": {
    "name": "Inputs List",
    "inputs": [{
        "sample_name": "Sample",
        ...
    },
    ...
    ]}},
...
```

This would allow us to "batch" workflows in the way we currently do but recast into this source->executor->sink model.

okotsopoulos (approved):
LGTM, this would help me to merge but you may like to wait for the approval of someone with more background knowledge on the historical code being altered.

api/test/wfl/integration/modules/covid_test.clj
ehigham: Sure. Nit: it's not an assignment, it's a binding :-p

PR #397, opened 2021-05-17
[GH-1314] expose workflows via `GET /api/v1/workload/{uuid}/workflows`

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1314
Adding a separate `workflows` selector will allow us to decouple the workload representation from serialisation artefacts that aren’t useful for a user.

In this PR, I'm changing
```
GET /api/v1/workloads
```
to only return workload metadata (everything but the workflows) and adding
```
GET /api/v1/workloads/{uuid}/workflows
```
to fetch the workflows in the workload.
Below is the swagger UI with the new endpoint.

Screenshot from 2021-05-17 15-52-24

### DIFF
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -80,12 +80,12 @@
   "Return the workflows managed by the workload."
   [request]
   (let [uuid (get-in request [:path-params :uuid])]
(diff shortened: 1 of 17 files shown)


rexwangcc (approved):
LGTM, thanks!

PR #402, opened 2021-05-19
GH-1307 GH-1312 Update method configuration using snapshot reference

okotsopoulos:
### Purpose

https://broadinstitute.atlassian.net/browse/GH-1307
https://broadinstitute.atlassian.net/browse/GH-1312

When creating submissions from snapshots, we have to modify the method configuration to specify the snapshot reference's name. Because Terra is an interactive tool, that same method config can be modified by a user at the same time.

I used the COVID demo script's method configuration updating as a starting point, to which I added additional checks for method configuration existence and version matching, as well as DB updating.

### Changes

In new method `update-method-configuration`:
1. GET the method configuration from Firecloud (presently includes verification that the method configuration exists, but should be moved to `verify-terra-executor!` when available).
2. Throw if the Firecloud-derived method configuration version doesn't match our DB record.
3. Update the method configuration with the snapshot reference name and POST to Firecloud.
4. Increment our DB record's method configuration version.

Supplemented integration tests to check that the snapshot reference name is POSTed to Firecloud as expected and that the executor object's version is incremented.

### Review Instructions

- I'm not convinced that someone unfamiliar with my new method would know what's going on, so fresh eyes to the space are especially welcome, as are suggestions to make it clearer.

### DIFF
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -422,7 +422,7 @@
   (cond (= "importSnapshot" fromSource) (import-snapshot! executor source-item)
         :else (throw (ex-info "Unknown fromSource" {:executor executor}))))
 
-(defn ^:private update-method-configuration
+(defn ^:private update-method-configuration!
   "Update METHODCONFIGURATION in WORKSPACE with snapshot reference NAME
   as :dataReferenceName via Firecloud.
   Update executor table record ID with incremented METHODCONFIGURATIONVERSION.
@@ -432,21 +432,20 @@
            methodConfiguration
            methodConfigurationVersion] :as executor}
    {:keys [name]                       :as _reference}]
-  (let [{:keys [methodConfigVersion] :as firecloud-mc}
+  (let [firecloud-mc
         (try
           (firecloud/get-method-configuration workspace methodConfiguration)
-          (catch Throwable t
+          (catch Throwable cause
             (throw (ex-info "Method configuration does not exist in workspace"
-                            {:executor executor
-                             :cause    (.getMessage t)}))))
+                            {:executor executor} cause))))
         inc-version (inc methodConfigurationVersion)]
-    (when-not (== methodConfigurationVersion methodConfigVersion)
+    (when-not (== methodConfigurationVersion (:methodConfigVersion firecloud-mc))
       (throw (ex-info (str "Firecloud method configuration version != expected: "
                            "possible concurrent modification")
                       {:executor            executor
                        :methodConfiguration firecloud-mc})))
     (-> firecloud-mc
-        (assoc :dataReferenceName name)
+        (assoc :dataReferenceName name :methodConfigVersion inc-version)
         (->> (firecloud/update-method-configuration workspace
                                                     methodConfiguration)))
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
@@ -517,7 +516,7 @@
   [source executor]
   (if-let [snapshot (peek-queue! source)]
     (let [reference        (from-source executor snapshot)
-          updated-executor (update-method-configuration executor reference)
+          updated-executor (update-method-configuration! executor reference)
           submission       (create-submission! updated-executor reference)]
       (write-workflows! updated-executor reference submission)
       (pop-queue! source)
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -422,6 +422,39 @@
   (cond (= "importSnapshot" fromSource) (import-snapshot! executor source-item)
         :else (throw (ex-info "Unknown fromSource" {:executor executor}))))
 
+(defn ^:private update-method-configuration
+  "Update METHODCONFIGURATION in WORKSPACE with snapshot reference NAME
+  as :dataReferenceName via Firecloud.
+  Update executor table record ID with incremented METHODCONFIGURATIONVERSION.
+  Return EXECUTOR with incremented METHODCONFIGURATIONVERSION."
+  [{:keys [id
+           workspace
+           methodConfiguration
+           methodConfigurationVersion] :as executor}
+   {:keys [name]                       :as _reference}]
+  (let [{:keys [methodConfigVersion] :as firecloud-mc}
+        (try
+          (firecloud/get-method-configuration workspace methodConfiguration)
+          (catch Throwable t
+            (throw (ex-info "Method configuration does not exist in workspace"
+                            {:executor executor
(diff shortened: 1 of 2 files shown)


api/src/wfl/module/covid.clj
```diff
@@ -422,6 +422,39 @@
...
+  as :dataReferenceName via Firecloud.
+  Update executor table record ID with incremented METHODCONFIGURATIONVERSION.
+  Return EXECUTOR with incremented METHODCONFIGURATIONVERSION."
+  [{:keys [id
+           workspace
+           methodConfiguration
+           methodConfigurationVersion] :as executor}
+   {:keys [name]                       :as _reference}]
+  (let [{:keys [methodConfigVersion] :as firecloud-mc}
+        (try
```
okotsopoulos, line 436: Method configuration existence check eventually to be moved upstream to `verify-terra-executor!` when available.

api/src/wfl/module/covid.clj
```diff
@@ -422,6 +422,39 @@
...
+          (firecloud/get-method-configuration workspace methodConfiguration)
+          (catch Throwable t
+            (throw (ex-info "Method configuration does not exist in workspace"
+                            {:executor executor
+                             :cause    (.getMessage t)}))))
+        inc-version (inc methodConfigurationVersion)]
+    (when-not (== methodConfigurationVersion methodConfigVersion)
```
okotsopoulos, line 443: Woof, names just similar enough to be confusing.

api/src/wfl/module/covid.clj
```diff
@@ -422,6 +422,39 @@
...
+      (throw (ex-info (str "Firecloud method configuration version != expected: "
+                           "possible concurrent modification")
+                      {:executor            executor
+                       :methodConfiguration firecloud-mc})))
+    (-> firecloud-mc
```
okotsopoulos, line 448: Pretty much copy-pasted from the [COVID demo script](https://github.com/broadinstitute/wfl/blob/develop/api/test/wfl/system/cdc_covid19_surveillance_demo.clj#L76).

api/src/wfl/module/covid.clj
```diff
@@ -481,12 +516,13 @@
   Return EXECUTOR."
   [source executor]
   (if-let [snapshot (peek-queue! source)]
-    (let [reference  (from-source executor snapshot)
-          submission (create-submission! executor reference)]
-      (write-workflows! executor reference submission)
-      (pop-queue! source)))
-  (update-terra-workflow-statuses! executor)
-  executor)
+    (let [reference        (from-source executor snapshot)
```
okotsopoulos, line 519: Had to reorg a bit to return the executor with incremented MC version in this case, or the original executor if we're only updating active-or-failed workflow statuses.  LMK if there's a better way to do this or if this is unclear.

api/src/wfl/module/covid.clj
```diff
@@ -422,6 +422,39 @@
...
+           methodConfiguration
+           methodConfigurationVersion] :as executor}
+   {:keys [name]                       :as _reference}]
+  (let [{:keys [methodConfigVersion] :as firecloud-mc}
+        (try
+          (firecloud/get-method-configuration workspace methodConfiguration)
+          (catch Throwable t
+            (throw (ex-info "Method configuration does not exist in workspace"
+                            {:executor executor
+                             :cause    (.getMessage t)}))))
```
ehigham, line 441: FYI ex-info takes a Throwable as its third argument, so no need to add the message here

api/src/wfl/module/covid.clj
```diff
@@ -422,6 +422,39 @@
...
+        inc-version (inc methodConfigurationVersion)]
+    (when-not (== methodConfigurationVersion methodConfigVersion)
+      (throw (ex-info (str "Firecloud method configuration version != expected: "
+                           "possible concurrent modification")
+                      {:executor            executor
+                       :methodConfiguration firecloud-mc})))
+    (-> firecloud-mc
+        (assoc :dataReferenceName name)
+        (->> (firecloud/update-method-configuration workspace
```
ehigham, line 450: I'd assoc the Inc-version here. I'm not sure if firecloud cares what version you send it though...

ehigham: Assoc it because I don't think we should rely on firecloud auto incrementing it

ehigham (approved):
Looks good! Thanks

api/src/wfl/module/covid.clj
```diff
@@ -481,12 +516,13 @@
...
   [source executor]
   (if-let [snapshot (peek-queue! source)]
-    (let [reference  (from-source executor snapshot)
-          submission (create-submission! executor reference)]
-      (write-workflows! executor reference submission)
-      (pop-queue! source)))
-  (update-terra-workflow-statuses! executor)
-  executor)
+    (let [reference        (from-source executor snapshot)
+          updated-executor (update-method-configuration executor reference)
```
ehigham, line 520: If be tempted to move the method configuration update into the create submission function, that way you don't need to juggle the "updated-executor"

okotsopoulos: Thanks for pointing this out.  Since `update-executor!` is meant to return an `executor` to feed into `update-sink!`, I thought that if we're altering the executor's method config version then we should ultimately return that updated executor.  But if we are content with only updating the corresponding DB record, then I'd be very happy to not have to juggle this!

api/src/wfl/module/covid.clj
okotsopoulos: Will take a closer look!

api/src/wfl/module/covid.clj
rexwangcc: `(= methodConfigurationVersion methodConfigVersion)`?

okotsopoulos: Unrelated to your comment, but I decided that this was a case where destructuring made things less clear rather than more clear, so tweaked this.

api/src/wfl/module/covid.clj
okotsopoulos: Done.

api/src/wfl/module/covid.clj
ehigham: Yep - I don't think we should rely on the source/executor object being up-to-date. That way, we can split these updates from sequential to parallel execution.

okotsopoulos: Great!  I will simplify this in my upcoming submission launch PR.

okotsopoulos: Addressed here: https://github.com/broadinstitute/wfl/pull/406

PR #406, opened 2021-05-19
GH-1316 Create submission within TerraExecutor update

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1316

Launch submissions from Rawls snapshot references after updating the method configuration with the reference's name and an incremented version.  I used the [COVID demo script](https://github.com/broadinstitute/wfl/blob/develop/api/test/wfl/system/cdc_covid19_surveillance_demo.clj#L82) as a starting point.

Because we already had code to write records to the executor details table from a snapshot, this PR was less involved than I anticipated.

### Changes

- Fully implemented `create-submission!` stub, pulling in existing `update-method-configuration!`.
- Per earlier PR feedback, simplified code which passed around updated executor with incremented method configuration version.  We don't expect the executor object passed from `update-executor! -> update-sink!` to be updated, what matters is that we update the underlying DB record with the incremented method configuration version so that the change propagates on executor load.
- Altered integration tests, now going to the `TerraExecutor` table to verify that method configuration version has been bumped.  Updated mocking to mock external calls only.

### Review Instructions

- No instructions.

### DIFF
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -469,8 +469,7 @@
 (defn ^:private update-method-configuration!
   "Update METHODCONFIGURATION in WORKSPACE with snapshot reference NAME
   as :dataReferenceName via Firecloud.
-  Update executor table record ID with incremented METHODCONFIGURATIONVERSION.
-  Return EXECUTOR with incremented METHODCONFIGURATIONVERSION."
+  Update executor table record ID with incremented METHODCONFIGURATIONVERSION."
   [{:keys [id
            workspace
            methodConfiguration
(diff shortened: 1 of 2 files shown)


api/src/wfl/module/covid.clj
```diff
@@ -625,13 +623,12 @@
...
-      (write-workflows! updated-executor reference submission)
-      (pop-queue! source)
-      (update-terra-workflow-statuses! updated-executor))
-    (update-terra-workflow-statuses! executor)))
+    (let [reference  (from-source executor snapshot)
+          submission (create-submission! executor reference)]
+      (write-workflows! executor reference submission)
+      (pop-queue! source)))
+  (update-terra-workflow-statuses! executor)
+  executor)
```
tbl3rd, line 631: when-let instead of if-let?
Passing an input through to the output always makes me think.

tbl3rd (approved):

api/src/wfl/module/covid.clj
```diff
@@ -535,8 +535,7 @@
 (defn ^:private update-method-configuration!
   "Update METHODCONFIGURATION in WORKSPACE with snapshot reference NAME
   as :dataReferenceName via Firecloud.
-  Update executor table record ID with incremented METHODCONFIGURATIONVERSION.
-  Return EXECUTOR with incremented METHODCONFIGURATIONVERSION."
+  Update executor table record ID with incremented METHODCONFIGURATIONVERSION."
```
rexwangcc, line 538: optional nit: or just 
```suggestion
  Update executor table record `id` with incremented `methodConfiguration`."
```
? Sad we have a mixture of docstring styles now :(

rexwangcc (approved):
I have a minor nit otherwise LGTM!

api/src/wfl/module/covid.clj
okotsopoulos: >when-let instead of if-let?
Thanks, good suggestion!

>Passing an input through to the output always makes me think.
Yeah, this seemed a bit funny to me too.  We may elect to lightly refactor `update-covid-workload` to avoid having to do this.

okotsopoulos: Switched to `when-let`.

api/src/wfl/module/covid.clj
okotsopoulos: I did a sweep of this namespace and adjusted all docstrings to backtick parameter names.

PR #410, opened 2021-05-20
GH-1298: TerraWorkspaceSink - verify target entity columns exist

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1298
- 
### Changes

- Added another check to wfl.covid/verify-terra-sink! and fixed the tests it broke.

### Review Instructions

- The system tests fail with a database error now. Will investigate on Monday.
- Please let me know whether I'm on the right track.

### DIFF
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -1,6 +1,7 @@
 (ns wfl.module.covid
   "Manage the Sarscov2IlluminaFull pipeline."
-  (:require [clojure.edn :as edn]
+  (:require [clojure.data :as data]
+            [clojure.edn :as edn]
             [clojure.set :as set]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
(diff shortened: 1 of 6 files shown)


api/src/wfl/module/covid.clj
```diff
@@ -958,17 +959,23 @@
...
+          entity-types (firecloud/list-entity-types workspace)
+          types        (-> entity-types keys set)]
+      (when-not (types entity-type)
+        (throw (UserException. "Entity not found"
+                               (util/make-map entityType types workspace))))
+      (let [attributes    (get-in entity-types [entity-type :attributeNames])
+            [missing _ _] (data/diff (set (vals fromOutputs)) (set attributes))]
+        (when (seq missing)
+          (throw (UserException. "Attributes missing for these outputs"
+                                 (util/make-map missing)))))))
```
ehigham, line 977: Might be useful to know what the actual attributes are too, not just the missing ones. Even better sorted lexicographically.

ehigham (approved):
Thank you. Looks good.

rexwangcc (approved):
LGTM

api/src/wfl/module/covid.clj
tbl3rd: Typically there will be lots of attributes but just a few missing ones,
but added and sorted.  We can revisit if `:missing` gets swamped.

PR #411, opened 2021-05-21
[GH-1327] Fix Start After Stop

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1327

In this change I'm fixing the strange behaviour were we can start a
covid workload after it was stopped. To do this, I've made it illegal to
stop a workload before it's been started.

I've done this because stop before start seems strange to me too. The
purpose of stop was to stop getting new data from the source but allow
all data in the workload pipeline to get flushed (ie. consume all
created snapshots and write all worklfows back). Stop was not meant for
[aborting a workload and cancelling/aborting all active prossessing](https://youtu.be/gO2hwvinFLE). If
there is a use case for such a thing (and i think there is) - I think we
should expose a new endpoint called `POST /api/v1/abort` or something.

Of course this PR is pretty small so if the team feels strongly about
stop before start I can find another solution/


rexwangcc (approved):
LGTM!

> Stop was not meant for
aborting a workload and cancelling/aborting all active prossessing. If
there is a use case for such a thing (and i think there is) - I think we
should expose a new endpoint called POST /api/v1/abort or something.

it makes me wonder if we should just use `pause` over `stop`.

Also note the test is failing at:

```
Syntax error (ClassNotFoundException) compiling at (wfl/module/batch.clj:1:1).
wfl.util.UserException
```

ehigham:
> LGTM!
> 
> > Stop was not meant for
> > aborting a workload and cancelling/aborting all active prossessing. If
> > there is a use case for such a thing (and i think there is) - I think we
> > should expose a new endpoint called POST /api/v1/abort or something.
> 
> it makes me wonder if we should just use `pause` over `stop`.
> 
> Also note the test is failing at:
> 
> ```
> Syntax error (ClassNotFoundException) compiling at (wfl/module/batch.clj:1:1).
> wfl.util.UserException
> ```
> 
> which is weird…

Oh bugger. This is caused by is writing all the wdl versions in prebuild before we've generated the class in build.

tbl3rd (approved):
This prevents a user from doing what that user wants to do,
but I like the clean-ups.

ehigham:
> This prevents a user from doing what that user wants to do,
> but I like the clean-ups.

Not sure what you're referring to - no user has asked for stopping before starting?

PR #418, opened 2021-05-26
[GH-1333] Fix NullPointerException in TDR Source

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1333
Caused by database changes not propagated to source code. Destructing
old keys for the source lead to null values.


okotsopoulos (approved):
The two test errors are due to TDR dev being full:

https://broadinstitute.slack.com/archives/CD4HBRFMG/p1622058450038500

PR #419, opened 2021-05-26
[GH-1336] Fix continual snapshot creation

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1336

Our BigQuery query was wrong as we were simply calling `toString` on a java.sql.Timestamp. This lead to queries like

```sql
SELECT (datarepo_row_id)
FROM `tdr-prod-workflow-launcher-dev.datarepo_ehigham_covid_bashing.flowcells`
WHERE last_modified_date BETWEEN '2021-05-26 15:31:18.287641' AND '2021-05-26T20:00:13'
```

The space in the interval was causing the problem and BigQuery interpreted that as a date. The fix is to ensure the lower and upper bound of our query-between interval is of the same type by coercing java.sql.Timestamp `last_checked` into a java.time.OffsetDateTime.

### DIFF
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -14,6 +14,7 @@
             [wfl.util :as util :refer [do-or-nil]]
             [wfl.wfl :as wfl])
   (:import [clojure.lang ExceptionInfo]
+           [java.sql Timestamp]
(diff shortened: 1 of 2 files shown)


rexwangcc (approved):
LGTM, thanks for fixing this!!

PR #421, opened 2021-05-27
[GH-1341] Fix accepting malformed labels in workload request

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1341
Enforce labels are of the form "name:value" where
- name starts with a letter followed by any combination of letters,
  numbers, underscores and dashes
- value is any non-blank string not containing `:`

I had to include labels in the batch workload request as the coercion
layer got confused when you gave it bad labels.

### DIFF
--- a/api/src/wfl/api/spec.clj
+++ b/api/src/wfl/api/spec.clj
@@ -106,7 +106,7 @@
 (s/def ::identifier string?)
 (s/def ::fromOutputs map?)
 (s/def ::fromSource string?)
-(s/def ::labels (s/* string?))
+(s/def ::labels (s/* util/label?))
 (s/def ::name string?)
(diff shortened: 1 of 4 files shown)


api/src/wfl/api/spec.clj
```diff
@@ -167,26 +169,25 @@
 
 (s/def ::covid-workload-request (s/keys :req-un [::executor
                                                  ::pipeline
+                                                 ::project
```
ehigham, line 172: this is actually required for now

api/src/wfl/module/covid.clj
```diff
@@ -130,7 +130,6 @@
    :executor
    :finished
    :labels
-   :pipeline
```
ehigham, line 133: We won't be sending back the pipeline anymore for this workload type.

PR #422, opened 2021-05-27
[GH-1324] Get or import snapshot

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1324

When retrying workflows, we're currently limited to trying the whole snapshot by creating a workload with that snapshot as the source. While an edge case, this always fails as the workload tries to import the snapshot into a workspace that already has the snapshot.

If snapshot import fails, we now try to return a snapshot reference in the workspace for the same snapshot id.

Note that this is intended as a stopgap until our App Services feature request is fulfilled: https://broadworkbench.atlassian.net/browse/AS-778

Over time, unbounded snapshot growth in the workspace will mean that in this edge case, paginating through all snapshots to find a match could get expensive.  They appear to be returned oldest to newest with no options for sorting.

### Changes

- Using [`cromwell/query`](https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/service/cromwell.clj#L180-L197) as a guide for lazy pagination (thanks @ehigham), introduced `rawls/get-snapshot-references` to lazily paginate through all snapshot references in a workspace.
- Added `rawls/create-or-get-snapshot-reference`: if importing the snapshot fails with a 409, finds the first existing snapshot reference matching the snapshot id.
- Supplemented automated testing.

### To Do

- Not within the scope of this ticket, but possible break-out of lazy pagination into a general utility.

### Review Instructions

- No instructions.


api/src/wfl/service/rawls.clj
```diff
@@ -22,10 +20,13 @@
       (http/get {:headers (auth/get-auth-header)})
       util/response-body-json))
 
+(def ^:private snapshot-endpoint
```
okotsopoulos, line 23: Preemptive to make updating to v2 endpoints easier.

api/src/wfl/service/rawls.clj
```diff
@@ -37,14 +38,33 @@
...
 (defn get-snapshot-reference
-  "Return the snapshot reference in fully-qualified Terra WORKSPACE with REFERENCE-ID."
+  "Return the snapshot reference in `workspace` with `reference-id`."
   [workspace reference-id]
-  (get-workspace-json workspace "snapshots" reference-id))
+  (get-workspace-json workspace snapshot-endpoint reference-id))
+
+(defn get-snapshot-references
+  "Lazily returns the snapshot references in `workspace`
+  with `limit` resources per page (default: 100)."
```
okotsopoulos, line 47: I welcome other suggestions for a different page size default.

docs/md/dev-process.md
```diff
@@ -17,10 +17,6 @@ without logging in first because it is bundled with the UI and not the API.
     server and access via the UI. See the development tips below for more
     information.
 
-!!! tip
```
okotsopoulos, line 20: We still have this tip, see above.

docs/md/dev-process.md
```diff
@@ -170,7 +166,10 @@ shared database in `gotc-dev` or *gasp* production.
         thus you could use `pq start` or `pq stop` to easily spin up and turn down
         the db.
 
-3. [Optional] Create wfl DB.
```
okotsopoulos, line 173: @ehigham this documentation change doesn't read right to me in the context of setting up a local Postgres, maybe the spec tip should go somewhere else?

api/src/wfl/service/rawls.clj
ehigham: Hard to tell without profiling and real uses. I tend to favour powers of 2 because much two-ness. 100 should be fine.

api/src/wfl/service/rawls.clj
```diff
@@ -37,14 +39,60 @@
...
+   (get-snapshot-references workspace 100)))
+
+(defn create-or-get-snapshot-reference
+  "Return first snapshot reference for `snapshot-id` in `workspace`,
+  creating it as `name` with `description` if it does not yet exist."
+  ([workspace snapshot-id name description]
+   (try
+     (create-snapshot-reference workspace snapshot-id name description)
+     (catch ExceptionInfo cause
+       (let [{:keys [status] :as _data} (.getData cause)]
```
ehigham, line 71: `ex-data` is the clojure selector for this!

okotsopoulos: Thank you thank you!

api/src/wfl/service/rawls.clj
```diff
@@ -37,14 +39,60 @@
...
+         (cond
+           (= 409 status)
+           ;; On a duplicate resource exception, try to fetch the resource.
+           (if-let [reference
+                    (->> (get-snapshot-references workspace)
+                         (filter #(= (get-in % [:reference :snapshot])
+                                     snapshot-id))
+                         first)]
+             reference
```
ehigham, line 80: It would be nice to move this into a helper function to break the line count a bit. Then your catch expression could read something like
```clojure
(def ^:private reference-creation-failed-message
  (str/join " " ["Could not create snapshot reference and"
                 "found no snapshot reference matching the"
                 "snapshot id."]))

...

(when-not (== 409 status)
  (throw cause))
(or (get-reference-if-exists workspace snapshot-id)
    (throw (ex-info reference-creation-failed-message
                    {:workspace   workspace
                     :snapshot-id snapshot-id
                     :name        name}
             cause)))
```

ehigham (approved):
Looks good. Thanks.

api/src/wfl/service/rawls.clj
okotsopoulos: Good suggestion, will implement.

docs/md/dev-process.md
ehigham: The code that generates the swagger page doesn't work. I don't know what to do tbh. Anything involving that part of the code makes me tear my hair out so I think your comment is appropriate!

okotsopoulos: Ohhh misunderstanding, I didn't introduce this comment but I thought maybe you did!  Some git weirdness… maybe @tbl3rd ?

api/src/wfl/module/covid.clj
```diff
@@ -686,18 +686,14 @@
...
-  [{:keys [fromSource] :as executor}
-   source-item]
-  (cond (= "importSnapshot" fromSource) (import-snapshot! executor source-item)
-        :else (throw (ex-info "Unknown fromSource" {:executor executor}))))
+  [{:keys [workspace fromSource] :as executor}
+   {:keys [name id]              :as _source-item}]
+  (cond (= "importSnapshot" fromSource)
+        (rawls/create-or-get-snapshot-reference workspace id name)
+        :else
+        (throw (ex-info "Unknown fromSource" {:executor executor}))))
```
tbl3rd, line 696: `throw` from `when`.
```clojure
(defn ^:private from-source
  "Coerce `source-item` to form understood by `executor` via `fromSource`."
  [{:keys [workspace fromSource] :as executor}
   {:keys [name id]              :as _source-item}]
  (when-not (= "importSnapshot" fromSource)
    (throw (ex-info "Unknown fromSource" {:executor executor})))
  (rawls/create-or-get-snapshot-reference workspace id name))
```

api/src/wfl/service/rawls.clj
```diff
@@ -37,14 +39,62 @@
    (create-snapshot-reference workspace snapshot-id name "")))
```
tbl3rd, line 39: Does this create an empty description? (-Do we care?-)

api/src/wfl/service/rawls.clj
```diff
@@ -37,14 +39,62 @@
...
 
 (defn get-snapshot-reference
-  "Return the snapshot reference in fully-qualified Terra WORKSPACE with REFERENCE-ID."
+  "Return the snapshot reference in `workspace` with `reference-id`."
   [workspace reference-id]
-  (get-workspace-json workspace "snapshots" reference-id))
+  (get-workspace-json workspace snapshot-endpoint reference-id))
```
tbl3rd, line 44: Thank you.

api/src/wfl/service/rawls.clj
```diff
@@ -37,14 +39,62 @@
...
+
+(defn get-snapshot-references
+  "Lazily returns the snapshot references in `workspace`
+  with `limit` resources per page (default: 100)."
```
tbl3rd, line 48: "Return a lazy sequence of snapshot references in …"
Should the caller know or care about the page size?

api/src/wfl/service/rawls.clj
```diff
@@ -37,14 +39,62 @@
...
+             (-> (workspace-api-url workspace snapshot-endpoint)
+                 (http/get {:headers      (auth/get-auth-header)
+                            :query-params {:offset offset
+                                           :limit  limit}})
+                 (util/response-body-json))
+             total    (+ offset (count resources))]
+         (lazy-cat resources (when (seq resources) (page total)))))]
+     (util/lazy-unchunk (page 0))))
+  ([workspace]
+   (get-snapshot-references workspace 100)))
```
tbl3rd, line 62: Do we need `lazy-unchunk`?
Cannot hurt much, I guess.

api/src/wfl/service/rawls.clj
```diff
@@ -37,14 +39,62 @@
...
+  "Return first snapshot reference for `snapshot-id` in `workspace`."
+  [workspace snapshot-id]
+  (->> (get-snapshot-references workspace)
+       (filter #(= (get-in % [:reference :snapshot]) snapshot-id))
+       first))
+
+(def ^:private reference-creation-failed-message
+  (str/join " " ["Could not create snapshot reference and"
+                 "found no snapshot reference matching the"
+                 "snapshot id."]))
```
tbl3rd, line 74: Thank you.  Better yet, respect English clause structure.
```clojure
(str/join " " ["Could not create snapshot reference"
               "and found no snapshot reference"
               "matching the snapshot id."])
```

api/test/wfl/integration/rawls_test.clj
```diff
@@ -20,23 +19,38 @@
...
+              (is (empty? rest))
+              (verify-snapshots snapshots names)))
+          (testing "Create or get"
+            (let [snapshot (rawls/create-or-get-snapshot-reference workspace
+                                                                   snapshot-id
+                                                                   (first names))]
+              (verify-snapshot [snapshot (first names)])))
+          (testing "Create already exists"
+            (is (thrown-with-msg? ExceptionInfo #"clj-http: status 409"
+                                  (make-reference (first names))))))))))
```
tbl3rd, line 53: We could simplify verification by moving some letfns inside the let,
but then we'd have to untangle the reference-ids binding from the Create test.

docs/md/dev-process.md
tbl3rd: Got me.  The docs get moved around a lot.

tbl3rd (approved):

docs/md/dev-process.md
okotsopoulos: Thanks!  I found its rightful home and will fix.

api/src/wfl/module/covid.clj
okotsopoulos: It's likely down the road that we may support multiple variants for `fromSource`, so the `cond` seemed appropriate for that future.

api/src/wfl/service/rawls.clj
okotsopoulos: It does, it's optional all the way to Rawls but figured the support may help in manual testing.

api/src/wfl/service/rawls.clj
okotsopoulos: I think so, our current use case for this method is to find the first reference matching a snapshot ID.  Not very many snapshot references to look through now, but if we retain them indefinitely as is our current plan we could experience some savings.  (Bigger than this use case, I was hoping down the road we might generalize this to other cases where lazy pagination could be an asset, like [`datarepo/list-snapshots`](https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/service/datarepo.clj#L163-L182).)

api/src/wfl/service/rawls.clj
okotsopoulos: I guess we don't need to expose it, but I did find it helpful in testing to confirm my understanding of the pagination!

okotsopoulos:
Failed tests stem from TDR dev being full, known issue unrelated to code changes.  Will merge anyway.

PR #426, opened 2021-06-01
[GH-1348] Remove :pipeline from COVID workload request

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1348
The `:pipeline` attribute of the workload request is meaningless for COVID-type workloads. Note that these workloads aren't even specific to COVID-19 processing, rather a generalisation.
In this PR, I've removed the `:pipeline` attribute. To do this, all workload operations on maps with a `nil` `:pipeline` will be implemented by the functions in the covid namespace.


ehigham:
> Have to admit that `defmethod` on `nil` makes me nervous.

Agreed. I plan to remove/refactor these in a future release because they're not fit for purpose anymore.

PR #428, opened 2021-06-02
[GH-1354] Enforce `snapshotReaders` are email addresses

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1354
This guards against erroring while creating snapshots if a user
makes a typo.


okotsopoulos (approved):
Looks good.

PR #432, opened 2021-06-04
[GH-1359] Overwrite Workspace Entity On Sink

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1359
When sinking a resubmitted workflow, the TerraWorkspaceSink would
append to the existing entity in the workspace. This leaves the entity
in an even more wrong state than the reason the sample was reanalysed.
Our fix (as agreed with cloreth) is to clobber the entity in the
workspace by deleting the entity if it exists then upserting the new one.

### DIFF
--- a/api/test/wfl/integration/modules/covid_test.clj
+++ b/api/test/wfl/integration/modules/covid_test.clj
@@ -554,19 +554,20 @@
          (is (== 1 (covid/queue-length! executor)))
          (is (not (covid/done? executor)))))))
 
+(def ^:private fake-entity-type "flowcell")
+(def ^:private fake-entity-name "test")
+
 (deftest test-update-terra-workspace-sink
-  (let [flowcell-id "test"
-        entity-type "flowcell"
-        workflow    {:uuid    "2768b29e-c808-4bd6-a46b-6c94fd2a67aa"
+  (let [workflow    {:uuid    "2768b29e-c808-4bd6-a46b-6c94fd2a67aa"
                      :status  "Succeeded"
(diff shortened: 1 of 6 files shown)


api/test/wfl/integration/modules/covid_test.clj
```diff
@@ -554,43 +554,96 @@
...
                            (zipmap [:sink_type :sink_items])
                            (covid/load-sink! tx)))]
-    (letfn [(verify-upsert-request [workspace [[name type _]]]
+    (letfn [(verify-upsert-request [workspace [[type name _]]]
               (is (= "workspace-ns/workspace-name" workspace))
-              (is (= name flowcell-id))
-              (is (= type entity-type)))]
+              (is (= type fake-entity-type))
+              (is (= name fake-entity-name)))
+            (throw-if-called [fname & args]
```
okotsopoulos, line 582: Nice, I was wondering how to achieve something like this in testing, will bank this.

okotsopoulos (approved):
Looks good.  Noted that the test failure is related to arrays (related to what @rexwangcc was working on Friday?) and not this changeset.  Wish we could have one central definition for entity [type name] structure, but alas!

ehigham:
> Looks good. Noted that the test failure is related to arrays (related to what @rexwangcc was working on Friday?) and not this changeset.

Created https://broadinstitute.atlassian.net/browse/GH-1360

> Wish we could have one central definition for entity [type name] structure, but alas!

I've tried to improve consistency here at least. I agree one definition would be better.

PR #439, opened 2021-06-09
GH-1371: Lint more.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1371

### Changes

- Added an alias for the eastwood linter.
- Ran kibit and eastwood and kondo and acquiesced to most of their demands.
- Tweaked a couple of other things noticed along the way. 
- Changed `lint` target to `format` and run it during builds.
- New `lint` target runs all 3 linters with their configuration files.

### Review Instructions

- The linters still complain.
- Please run them and let me know what you think.
- We can probably suppress some of the remaining complaints.
- We may not want to keep all of these changes.

### DIFF
--- a/api/src/wfl/module/copyfile.clj
+++ b/api/src/wfl/module/copyfile.clj
@@ -101,9 +101,8 @@
 (defn create-copyfile-workload!
   "Use transaction TX to add the workload described by REQUEST."
   [tx {:keys [items common] :as request}]
-  (letfn [(nil-if-empty [x] (when (seq x) x))
-          (merge-to-json [shared specific]
-            (json/write-str (nil-if-empty (util/deep-merge shared specific))))
+  (letfn [(merge-to-json [shared specific]
+            (json/write-str (not-empty (util/deep-merge shared specific))))
           (serialize [workflow id]
             (-> workflow
                 (assoc :id id)
--- a/api/src/wfl/module/copyfile.clj
+++ b/api/src/wfl/module/copyfile.clj
@@ -22,7 +22,7 @@
   (cromwell/submit-workflow
    url
    workflow-wdl
-   (-> inputs (util/prefix-keys (str pipeline ".")))
+   (util/prefix-keys inputs (str pipeline "."))
    options
    labels))
 
@@ -101,7 +101,7 @@
 (defn create-copyfile-workload!
   "Use transaction TX to add the workload described by REQUEST."
   [tx {:keys [items common] :as request}]
-  (letfn [(nil-if-empty [x] (if (empty? x) nil x))
+  (letfn [(nil-if-empty [x] (when (seq x) x))
           (merge-to-json [shared specific]
             (json/write-str (nil-if-empty (util/deep-merge shared specific))))
           (serialize [workflow id]
--- a/api/src/wfl/module/covid.clj
+++ b/api/src/wfl/module/covid.clj
@@ -386,8 +386,9 @@
    (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
      (update-last-checked tx source now))))
 
-(defn ^:private timestamp-to-offsetdatetime [^Timestamp t]
+(defn ^:private timestamp-to-offsetdatetime
   "Parse the Timestamp `t` into an `OffsetDateTime`."
+  [^Timestamp t]
   (OffsetDateTime/ofInstant (.toInstant t) (ZoneId/of "UTC")))
 
 (def ^:private bigquery-datetime-format
@@ -517,7 +518,7 @@
         alter   "ALTER TABLE %s ALTER COLUMN id ADD GENERATED ALWAYS AS IDENTITY"
         details (format "%s_%09d" tdr-snapshot-list-type id)]
     (jdbc/db-do-commands tx [(format create details) (format alter details)])
-    (jdbc/insert-multi! tx details (map #(-> {:item (pr-str %)}) snapshots))
+    (jdbc/insert-multi! tx details (map #(hash-map :item (pr-str %)) snapshots))
     [tdr-snapshot-list-type details]))
 
 (defn ^:private load-tdr-snapshot-list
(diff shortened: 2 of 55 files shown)


api/src/wfl/module/copyfile.clj
```diff
@@ -101,7 +101,7 @@
 (defn create-copyfile-workload!
   "Use transaction TX to add the workload described by REQUEST."
   [tx {:keys [items common] :as request}]
-  (letfn [(nil-if-empty [x] (if (empty? x) nil x))
+  (letfn [(nil-if-empty [x] (when (seq x) x))
```
ehigham, line 104: that's just `not-empty`

api/test/wfl/unit/workloads_test.clj
```diff
@@ -5,13 +5,13 @@
...
 
 (deftest test-saved-before?
-  "in version 0.4.0, wgs workloads were serialized using CromwellWorkload table"
-  (is (not (workloads/saved-before? "0.4.0" {:version "0.4.0"})))
-  (is (not (workloads/saved-before? "0.4.0" {:version "1.0.0"})))
-  (is (not (workloads/saved-before? "0.4.0" {:version "0.4.1"})))
-  (is (workloads/saved-before? "0.4.0" {:version "0.3.8"}))
-  (is (workloads/saved-before? "0.4.0" {:version "0.0.0"}))
-  (is (thrown? ExceptionInfo (workloads/saved-before? "0.4" {:version "0.0.0"}))))
+  (testing "in version 0.4.0, wgs workloads were serialized using CromwellWorkload table"
```
ehigham, line 8: I really don't like the "testing blah blah blah" in tests like this. It adds nothing. Rename the test if you think it's not clear.
Also, this change was not limited to WGS

ehigham (approved):
OK i guess? Not sure how much value some of these changes add.

ehigham:
If the linters complain perhaps you could fix that and have them run on the lint target?

api/src/wfl/module/copyfile.clj
okotsopoulos: Clojure says that `(seq x)` is the preferred idiom for testing a non-empty collection: https://clojuredocs.org/clojure.core/seq#example-542692d5c026201cdc327079

ehigham: Good to know.

rhiananthony (approved):
👍🏼

api/src/wfl/module/copyfile.clj
tbl3rd: I like `not-empty` for this.

api/test/wfl/unit/workloads_test.clj
tbl3rd: I just wrapped what looked like an attempt at a docstring.
My goal was to reduce reports from the linter without changing behavior.
Except where a linter discovered a problem: then I tried to fix the problem.

api/src/wfl/module/copyfile.clj
tbl3rd: FYI: `seq` returns a sequence on the collection, but `not-empty` returns the original collection.

api/src/wfl/module/covid.clj
```diff
@@ -25,20 +25,20 @@
 ;; interfaces
```
okotsopoulos, line 25: Many / most of these multimethods moved FYI:

https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/stage.clj

api/src/wfl/module/covid.clj
```diff
@@ -50,59 +50,59 @@
...
+  "Update the `sink` with the `executor`."
+  (fn [_executor sink] (:type sink)))
 
 (defmulti load-sink!
-  "Use `tx` to load the workload sink with `sink_type`."
-  (fn [tx workload] (:sink_type workload)))
+  "Use `tx` to load the `workload` sink with `sink_type`."
+  (fn [_tx workload] (:sink_type workload)))
 
 ;; Processing "stage" operations
```
okotsopoulos, line 103: This one too:

https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/stage.clj

api/src/wfl/module/covid.clj
tbl3rd: Don't worry about it.  I'll just lint again before merging.

rexwangcc (approved):

api/test/wfl/tools/workflows.clj
```diff
@@ -42,8 +42,8 @@
    type object))
 
 (defn foldl
-  "Use the workflow `type` to left-fold the `object`, calling `f` on the values
-   with primitive types with the current state."
+  "Use ternary function `f` to reduce `object` tree of `type` starting
+  from `init` while preserving the types of intermediate reductions."
```
tbl3rd, line 46: @ehigham (-How wrong is this new docstring?-)

ehigham: While preserving the types of intermediate reductions? what does that mean?

ehigham: You could change `f` to take [result-so-far [type value]]` so it looks like "reduce" if that helps.

ehigham (approved):

api/test/wfl/tools/workflows.clj
tbl3rd: The string tags are the "types".
Reductions are the intermediate result subtrees such as might be returned by `clojure.core/reductions`.
This differs from a generic left fold (or `reduce`) in that it treats one of the parameters differently.
Perhaps in your terminology: this is a left fold over a tree with an implicit lens.
(FWIW: I don't care where the grouping happens. This doesn't have to look like `reduce`.)

tbl3rd: Mostly, I just want some mention of the roles of the parameters to `foldl`,
and to cast it in Clojure terms so it is familiar to the team (or me, at least).
Please give me a thumbs up, a docstring more acceptable to you, or a "revert it".
Thanks.

PR #446, opened 2021-06-11
Removing the arrays module and the corresponding integration test

rhiananthony:
### Purpose
The purpose of this PR is to remove the unused arrays module and the corresponding integration test.

This is linked to https://broadinstitute.atlassian.net/browse/GH-1360

### Review Instructions
Please verify that this is not an unnatural unholy thing that I am doing.


okotsopoulos:
For me, cleanup like this has always been deceptive.  You've deleted the namespace which is great, but there are still references to `wfl.module.arrays` in the repo.  I could see this driving a build breakage, though I didn't look too closely in this PR's build logs.

The layer beyond this is removing any code which was only used with the `arrays` module, including test resources, specs, etc.  Can be harder to track down and wouldn't break anything to leave in, but add to cruft over time and are easiest to clean up while cleaning up the main code.

ehigham:
> For me, cleanup like this has always been deceptive. You've deleted the namespace which is great, but there are still references to `wfl.module.arrays` in the repo. I could see this driving a build breakage, though I didn't look too closely in this PR's build logs.
> 
> The layer beyond this is removing any code which was only used with the `arrays` module, including test resources, specs, etc. Can be harder to track down and wouldn't break anything to leave in, but add to cruft over time and are easiest to clean up while cleaning up the main code.

+1 Sadly it's not enough to just delete those files. Delete them and all references to them, including all related test tools (workload requests etc)

ehigham (approved):
yeah boiii

docs/md/modules-aou-arrays.md
```diff
@@ -1,4 +1,4 @@
-# Arrays module
+# All Of Us Arrays module
```
okotsopoulos, line 1: Thank you!

okotsopoulos (approved):

PR #430, opened 2021-06-03
Documenting a Workload

rhiananthony:
### Purpose
The purpose of this PR is to create documentation to describe a workload. 

I tried to create a generic workload, which I linked to the main readme. I also added a specific workload which was related to the covid work. Please let me know if this is incorrect.

This work is for GH-1330

### Changes
New documentation

### Review Instructions
Please review content of the documentation for omissions, please review grammar, and please review language for readability.

PR #449, opened 2021-06-15
[GH-1347] Document the workload `Sink` and `Terra Workspace` sink implementation

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1347

In this PR, I've stated work on documenting the `Sink` model as well as its implementations. The style I've taken is a "User Guide" giving detailed instructions on how to create specific sink types and then a "Developer Guide" telling how to add new sink implementations. It's definitely not a comprehensive guide (I've omitted a bunch of things like DB schemas etc that I don't think are particularly interesting) so I welcome suggestions for where it could be improved/clarified!

My intention is to submit this then do a follow-up PR to link it into @rhiananthony's work in https://github.com/broadinstitute/wfl/pull/430
Thanks.


docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
+# Sink
+The workload `Sink` models the terminal stage of a processing pipeline. In a 
+typical workload configuration, a `Sink` can be used to write workflow outputs
+to a desired location in the cloud.
```
okotsopoulos, line 4: Re: "in the cloud" - any chance of us expanding support to on-prem applications in the future?

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Terra Workspace Sink
+You can write workflow outputs to a Terra Workspace using the `Terra Workspace`
+sink. A typical `Terra Workspace` sink configuration in the workload request
+looks like:
+```json
+{
+  "name": "Terra Workspace",
+  "workspace": "{workspace-namespace}/{workspace-name}",
+  "entityType": "{name}",
```
okotsopoulos, line 18: Having `{name}` as the value here is a bit confusing to me given the `name` key above, would qualify.

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+
+## User Guide
+You can configure the type of `Sink` used in your workload by changing the
+`sink` attribute of your workload request.
+
+### Terra Workspace Sink
+You can write workflow outputs to a Terra Workspace using the `Terra Workspace`
+sink. A typical `Terra Workspace` sink configuration in the workload request
+looks like:
+```json
```
okotsopoulos, line 14: As a newer person I'd find it helpful in addition to this generic sink configuration to include a specific example which can then be referenced in the subsections below, maybe with screenshots showing the relation to the Terra UI.

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+#### `identifier`
+The `identifier` is the name of a pipeline output that should be used as the 
+name of each newly created entity. 
+
+Example - Let's say the pipeline you're running has an output called
+"sample_name" that uniquely identifies the inputs and outputs to that pipeline.
+By setting `"identifier": "sample_name"` in the sink configuration, entities
+will be created using the "sample_name" as the entity name.
+
+Note that when two sets of pipeline outputs share the same "identifier" value,
```
okotsopoulos, line 58: Is this section worth flagging as a warning?

Stepping outside of documentation, do we have sufficient visibility into an identifier collision to flag as a `UserException`?

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+the first set of outputs will be overwritten by the second in the workspace.
+
+#### `fromOutputs`
+`fromOutputs` configures how to create new entities from pipeline outputs by 
+mapping the output names to attributes in the `entityType`. Note that all 
+attribute names must exist in the entityType before the workload  creation.
+
+`fromOutputs` allows a small about of flexibility in how to construct an entity
```
okotsopoulos, line 66: Nit: "a small amount"

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+- `"attribute": ["output0", "output2"]`
+  Make an attribute from a list of pipeline outputs.
+
+- `"attribute": "$value"`
+  Use the literal "value" for an attribute.
+
+## Developer Guide
+A sink is one satisfying the `Sink` protocol as below:
+```clojure
+(defprotocol Sink
```
okotsopoulos, line 81: Sorry to say I'm confused by the use of a `defprotocol` here when we don't do this in the code :( with more distance from our work in the space, I'd see the mismatch between code and docs and assume that the documentation is out of date.

If a protocol-based approach helps in documentation, is it also something we can implement?  I'm not that familiar with them, doesn't look like they're widely used in `wfl`.

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+  (update-sink!
+    ^Sink
+    [^Queue upstream ;; The queue to sink outputs from
```
okotsopoulos, line 84: Is this type hint intentionally generic as opposed to `^Executor` to leave open the possibility of pipelines with a non-executor queue feeding into the sink?  Do we have a known example in mind?

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+     ^Sink  sink     ;; This sink instance
+    ]
+    "Update the internal state of the `sink`, consuming objects from the
+     upstream `Queue`, performing any external effects as required.
+     Notes:
+     - The `Sink` and `Queue` are parameterised types and the `Queue`'s 
+       parameterisation must be convertible to the `Sink`s.
+     - Where possible, `Sink` implementations should minimise blocking and use
```
okotsopoulos, line 92: RE: "… should minimise blocking", I'd find it helpful to make an explicit reference to avoiding external calls within DB transactions.

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+       an internal queue to schedule and manage long-running computations."))
+```
+
+!!! info
+    The `Sink` protocol is currently implemented by the `update-sink!`
+    multimethod.
+
+To be used in a workload, a `Sink` implementation should satisfy the processing
+`stage` protocol and the `to-edn` multimethod in addition to the following
```
okotsopoulos, line 101: What's the `stage` protocol?

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+## Developer Guide
+A sink is one satisfying the `Sink` protocol as below:
+```clojure
+(defprotocol Sink
+  (update-sink!
+    ^Sink
+    [^Queue upstream ;; The queue to sink outputs from
+     ^Sink  sink     ;; This sink instance
+    ]
+    "Update the internal state of the `sink`, consuming objects from the
```
okotsopoulos, line 87: The docstring should go above the args, but I'm not sure if it also goes above the type hint for the return value.  @tbl3rd?

docs/md/sink.md
ehigham: I certainly hope not.

docs/md/sink.md
ehigham: good catch

docs/md/sink.md
okotsopoulos: Sounds good, let's manifest the future that we want to see.

docs/md/sink.md
ehigham: This is by design

docs/md/sink.md
ehigham: thanks

docs/md/sink.md
ehigham: We pass in `Queue` in tests. It should actually be a `Queue` as per Liskov.

docs/md/sink.md
ehigham: `validate` and `done?`

ehigham: You moved them into the `stage` namespace.

docs/md/sink.md
ehigham: Nope: https://clojuredocs.org/clojure.core/defprotocol

tbl3rd: > (I think) the docstring should go above the args, but I'm not sure if it also goes above the type hint for the return value. @tbl3rd?

Metadata generally applies to the immediate next expression.

docs/md/sink.md
ehigham: I actually think we should use `defprotocol` because it forces us to define a data model for workload/source/sink/executor that's different from what a user sees and what's in the database.
The multi-methods are generic functions but the protocols define the operations you can perform on the set of objects of that type.

docs/md/sink.md
okotsopoulos: TIL!

docs/md/sink.md
okotsopoulos: Ah right, to smooth retries.

docs/md/sink.md
```diff
@@ -0,0 +1,132 @@
...
+   workload record as `sink_type` and `sink_items`.
+   Notes:
+   - This is a factory method registered for workload creation.
+   - The `Sink` type string must match a value of the `sink` enum in the
+     database schema.   
+   - This multimethod is type-dispatched on the `:name` association in the
+     `request`."
+  (fn ^[^String ^String] 
+      [^Connection         transaction  ;; JDBC Connection
+       ^long               workload-id  ;; ID of the workload being created
```
okotsopoulos, line 118: Should the workload-id be hinted as a UUID instead?

ehigham: Nope - it's a long.

docs/md/sink.md
okotsopoulos: Nice.  I agree this would be an improvement on implementation!

I'd feel good about approving this PR if we have a plan for imminent refactoring to `defprotocols` (maybe expand the scope of the "break apart COVID ns" tickets).  Otherwise I'm not in favor of releasing documentation that reflects an implementation change that we want to pursue but haven't.

docs/md/sink.md
okotsopoulos: Right.  Was confused since that isn't a protocol (yet!)

docs/md/sink.md
ehigham: Sure - I'm keen to do it!

ehigham: I also like the protocols for illustrative purposes as they group together a set of operations and call that a thing whereas there's no way to really do this with multi-methods.

docs/md/sink.md
```diff
@@ -0,0 +1,133 @@
...
+A sink is one satisfying the `Sink` protocol as below:
+```clojure
+(defprotocol Sink
+  (update-sink!
+    ^Sink
+    [^Queue upstream ;; The queue to sink outputs from
+     ^Sink  sink     ;; This sink instance
+    ]
+    "Update the internal state of the `sink`, consuming objects from the
+     upstream `Queue`, performing any external effects as required.
```
okotsopoulos, line 89: Nit: think this should be "… from the `upstream` queue".

tbl3rd (approved):
I believe all the comments are resolved.

rhiananthony (approved):

docs/md/sink.md
ehigham: Ill do a follow-up PR

docs/md/sink.md
```diff
@@ -0,0 +1,133 @@
...
+| `workspace`   | The Terra Workspace to write pipeline outputs to.            |
+| `entityType`  | The entity type in the `workspace` to write outputs to.      |
+| `identifier`  | Selects the output that will be used as the entity name.     |
+| `fromOutputs` | Mapping from outputs to attribute names in the `entityType`. |
+
+#### `workspace`
+The workspace is a `"{workspace-namespace}/{workspace-name}"` string as it
+appears in the URL path in the Terra UI. The workspace must exist prior to
+workload creation. You must ensure that `workflow-launcher@firecloud.org` is
+a workspace "Writer" in order to write entities to the workspace.
```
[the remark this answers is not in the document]
ehigham: I'm not going to document how terra works.

PR #451, opened 2021-06-17
GH-1346 Document workload Executor and Terra implementation

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1346

Added `Executor` documentation to WFL's GitHub Pages site, to be connected with @rhiananthony 's work on higher level workload documentation once merged: https://github.com/broadinstitute/wfl/pull/430

The best way to take in this changeset is via a local docs instance:

```
make docs
. derived/.venv/docs/bin/activate && ( cd derived/docs; python3 -m mkdocs serve )
```

And go to `http://127.0.0.1:8000/executor.html`.

### Review Instructions

- I'll add comments inline on spots I'd like to call out.
- I especially welcome feedback from folks less familiar with the executor update loop.
- Note that future work on retry support should be coupled with doc changes here.
- Thanks @ehigham for your in-progress work on documenting sources and sinks.  Related, please keep eyes open for copypasta.


docs/md/executor.md
```diff
@@ -0,0 +1,163 @@
...
+{
+  "name": "Terra",
+  "workspace": "wfl-dev/CDC_Viral_Sequencing",
+  "methodConfiguration": "wfl-dev/sarscov2_illumina_full",
+  "methodConfigurationVersion": 1,
+  "fromSource": "importSnapshot"
+}
+```
+
+![](./assets/terra-method-configuration.png)
```
okotsopoulos, line 46: To satisfy my preference for a concrete example, I found a concrete request with a Terra screenshot provided enough specificity without referencing back to it throughout the doc.

docs/md/executor.md
```diff
@@ -0,0 +1,163 @@
...
+in the Terra UI.
+
+Prerequisites:
+
+- The workspace must exist prior to workload creation.
+- `workflow-launcher@firecloud.org` must be a workspace "Writer" in order to
+  import snapshots to the workspace.
+- The workspace must be compatible with any downstream processing stage that
+  consumes its workflows.
+  For example, we cannot sink a workflow's outputs to an outside workspace.
```
okotsopoulos, line 69: I put this in because I was pretty sure it was true.

Is it?
If it is, is this something we should verify at workload creation?

docs/md/executor.md
```diff
@@ -0,0 +1,163 @@
...
+from an upstream `Source` into a data type understood by the executor.
+
+Prerequisites:
+
+- Must be one of the following supported coercion instructions.
+
+###### `importSnapshot`
+Workflow-launcher should import Terra Data Repository (TDR) snapshots
+into `workspace` as snapshot references,
+updating `methodConfiguration` with the reference as its root entity type.
```
okotsopoulos, line 106: The way things are now, the `Terra` executor assumes that it will always be dealing with a snapshot reference when triggering a submission.  This method configuration update happens outside of our conditional coercion from the `Source`:

https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/module/covid.clj#L659-L666

I wasn't sure if we should anticipate a future where a `Terra` executor is meant to trigger a submission from a table in the workspace rather than a snapshot.

docs/md/executor.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Developer Guide
```
okotsopoulos, line 108: I agree with @rexwangcc that I'd prefer to see the Developer Guide(s) separate.

I could see separate User Guide files for source, executor and sink since the number of implementations may grow.  But I'd be in favor of grouping the developer-facing information in a single guide.

docs/md/executor.md
```diff
@@ -0,0 +1,163 @@
...
+| `fromSource`                 | Instruction to coerce an output from an upstream `Source` to a type understood by this `executor`.|
+
+#### `workspace`
+A `{workspace-namespace}/{workspace-name}` string as it appears in the URL path
+in the Terra UI.
+
+Prerequisites:
+
+- The workspace must exist prior to workload creation.
+- `workflow-launcher@firecloud.org` must be a workspace "Writer" in order to
```
ehigham, line 65: It actually has to be a workspace owner

docs/md/executor.md
ehigham: I don't think it is. We can write outputs to another workspace no problem. The only issue is that files wouldn't be copied to the new workspace so that is technically a bug!
The idea is that the executor care about what's downstream of it.

docs/md/executor.md
ehigham: I think we will be expanding this in the future. It's likely we'll have other sources (for eMerge processing for example) that don't offer snapshots to the executor.

docs/md/executor.md
```diff
@@ -0,0 +1,163 @@
...
+
+###### `importSnapshot`
+Workflow-launcher should import Terra Data Repository (TDR) snapshots
+into `workspace` as snapshot references,
+updating `methodConfiguration` with the reference as its root entity type.
+
+## Developer Guide
+An executor is a `Queue` that satisfies the `Executor` protocol below:
+```clojure
+(defprotocol Executor
```
ehigham, line 111: There's also `workflows` to fetch the workflows this executor has processed

rexwangcc (approved):

docs/md/executor.md
okotsopoulos: Thanks for clarifying!  I will remove this example.

Is the file copy issue a bug on our side? I can file a ticket.

docs/md/executor.md
okotsopoulos: Got it, will add.  I am stuck on the return type hint though, do you have an idea of what it should be?

docs/md/executor.md
okotsopoulos: I will leave this as is for now, and future work should involve looking at existing docs to update accordingly (… I hope!)

rhiananthony (approved):

docs/md/executor.md
```diff
@@ -0,0 +1,169 @@
...
+    "Consume items from the `upstream` queue and enqueue
+     to the `executor` queue for consumption by a later processing stage,
+     performing any external effects as necessary.
+     Implementations should avoid maintaining in-memory state and making long-
+     running external calls, favouring internal queues to manage such tasks
+     asynchronously between invocations.  Note that the `Executor` and `Queue`
+     are parameterised types and the `Queue`'s parameterisation must be
+     convertible to the `Executor`s.")
+  (executor-workflows
+    ^WhatAmI?                 ;; TODO: type hint for workflow list
```
okotsopoulos, line 125: Stuck on this, would love some help.

docs/md/executor.md
ehigham: That would be really helpful. Thanks.

docs/md/executor.md
tbl3rd: FTR: `^PersistentVector` is about all one can do here.

okotsopoulos: D'oh… I didn't push this before merging like I thought!

Will make a new PR to add.

PR #453, opened 2021-06-22
[GH-1294] Migrate to use datasets in prod TDR

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1294

### Changes

- Migrate to use [dataset](https://data.terra.bio/datasets/4a5d30fe-1f99-42cd-998b-a979885dea00/query) in production TDR.
- Update some tests so they pass.
- Update the default ENV values accordingly.

### Review Instructions

- No instructions.

### DIFF
--- a/api/src/wfl/environment.clj
+++ b/api/src/wfl/environment.clj
@@ -34,7 +34,7 @@
    "WFL_COOKIE_SECRET"
    #(-> "secret/dsde/gotc/dev/zero" vault-secrets :cookie_secret)
    "WFL_TDR_URL"
-   #(-> "https://jade.datarepo-dev.broadinstitute.org")
+   #(-> "https://data.terra.bio")
    "WFL_OAUTH2_CLIENT_ID"
    #(-> "secret/dsde/gotc/dev/zero" vault-secrets :oauth2_client_id)
    "WFL_POSTGRES_PASSWORD"
@@ -52,7 +52,7 @@
    "WFL_CROMWELL_URL"
    #(-> "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org")
    "WFL_TDR_DEFAULT_PROFILE"
-   #(-> "390e7a85-d47f-4531-b612-165fc977d3bd")
+   #(-> "6370f5a1-d777-4991-8200-ceab83521d43")
    "WFL_TDR_SA"
    #(-> "jade-k8-sa@broad-jade-dev.iam.gserviceaccount.com")
    "WFL_WFL_URL"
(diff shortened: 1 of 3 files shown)


api/src/wfl/environment.clj
```diff
@@ -52,7 +52,7 @@
    "WFL_CROMWELL_URL"
    #(-> "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org")
    "WFL_TDR_DEFAULT_PROFILE"
-   #(-> "390e7a85-d47f-4531-b612-165fc977d3bd")
+   #(-> "6370f5a1-d777-4991-8200-ceab83521d43")
    "WFL_TDR_SA"
    #(-> "jade-k8-sa@broad-jade-dev.iam.gserviceaccount.com")
```
tbl3rd, line 57: The service account does not change?

tbl3rd (approved):

api/src/wfl/environment.clj
rexwangcc: Nice catch! I actually meant to remove it since it's not used anywhere anymore, somehow I forgot to commit that change…

api/test/wfl/integration/datarepo_test.clj
```diff
@@ -78,19 +78,17 @@
             (is (= 1 row_count))
             (is (= 0 bad_row_count))))))))
 
-(def ^:private testing-dataset "85efdfea-52fb-4698-bee6-eef76104a7f4")
+(def ^:private testing-dataset "4a5d30fe-1f99-42cd-998b-a979885dea00")
```
ehigham, line 81: While you're here, it might help us identify datasets if you include their name here (i.e. make this a map with a `:name` as well as an `:id`)

ehigham (approved):

api/test/wfl/integration/datarepo_test.clj
ehigham: The same for the snapshot too - I think the names are easier to remember (and not delete) than the uuids

rexwangcc: fair enough, a map sounds a great idea! working on it now!

api/src/wfl/service/datarepo.clj
```diff
@@ -266,3 +266,8 @@
    (->> (map name columns)
         util/to-comma-separated-list
         (query-table-between-impl dataset table between interval))))
+
+(snapshot "0ef4bc30-b8a0-4782-b178-e6145b777404")
+;"workflow_launcher_testing_dataset7561609c9bb54ca6b34a12156dc947c1"
+(dataset "4a5d30fe-1f99-42cd-998b-a979885dea00")
+;"workflow_launcher_testing_dataset"
```
tbl3rd, line 273: May wanna wrap these in a Rich comment or sumpin.

rexwangcc: My bad, I didn't mean to check those in 🤦 … Will remove them in a follow up PR!!

PR #454, opened 2021-06-22
[GH-1334] Move `source` Interface and Implementation into new namespace

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1334
Move `source` interface, helper functions and implementations into new `wfl.source` namespace
Move some source-specific integration tests to a new `source-test` namespace
Remove IO from datarepo namespace


rhiananthony (approved):

tbl3rd (approved):
I didn't check that all the moved code was copied correctly.

tbl3rd (commented):
Also missed the last couple of commits.

makerules/modules.mk
```diff
@@ -38,11 +37,10 @@ all: $(MAKE_TARGETS)
...
 $(PREBUILD):    $(shell eval $(MKDIR) $(DERIVED_MODULE_DIR))
-$(FORMAT):      $(PREBUILD)
-$(BUILD):       $(FORMAT)
+$(LINT):        $(PREBUILD)
+$(BUILD):       $(PREBUILD)
 $(UNIT):        $(BUILD)
 $(INTEGRATION): $(BUILD)
-$(LINT):        $(BUILD)
 $(IMAGES):      $(BUILD)
 $(SYSTEM):      $(BUILD)
```
tbl3rd, line 45: Retain the `format` target for easy reformatting?
I'm OK with removing the dependencies though.

api/src/wfl/tsv.clj
```diff
@@ -77,7 +77,7 @@
   [table]
   (let [writer (StringWriter.)]
     (write table writer)
-    (.toString writer)))
+    (str writer)))
```
tbl3rd, line 80: Kinda prefer `.toString` here as well, but OK if that works.
(I know some linter complains.)

makerules/modules.mk
ehigham: That's what the clojure alias is for - make isn't for general scripting

tbl3rd: Then maybe restore this comment?

# Run `clojure -M:format` in this directory when this fails.

Even better: have a format check failure describe the remedy.

tbl3rd: Love markdown.

PR #455, opened 2021-06-23
[GH-1335] Move `Executor` Code to new Namespace

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1335
Doc-strings from https://github.com/broadinstitute/wfl/pull/451

### DIFF
--- a/api/src/wfl/service/firecloud.clj
+++ b/api/src/wfl/service/firecloud.clj
@@ -6,7 +6,9 @@
             [wfl.auth          :as auth]
             [wfl.environment   :as env]
             [wfl.util          :as util])
-  (:import [java.util UUID]))
+  (:import [clojure.lang ExceptionInfo]
+           [java.util UUID]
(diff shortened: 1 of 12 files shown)


rexwangcc (approved):
Looks like the tests are complaining about `UserException`

ehigham:
> Looks like the tests are complaining about `UserException`

Yeah - the prebuild expects that class to be generated but it is actually generated in build. Solution was generate `UserException` on prebuild.

api/src/wfl/service/firecloud.clj
```diff
@@ -20,9 +22,18 @@
       (http/get {:headers (auth/get-auth-header)})
       util/response-body-json))
 
-(defn workspace [workspace]
+(defn workspace-or-throw
```
okotsopoulos, line 25: Nice, I moved this here locally too.

PR #461, opened 2021-06-24
[GH-1401] Add `retry` Attribute to TerraExecutorDetails Type and Instances Thereof

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1401
Verified that this change affects all existing tables created from that type.
When workflows are requested, only the non-retried workflows are returned.
Note that unconsumed workflows are still visible to downstream. Note sure what the correct behaviour should be here.

### DIFF
--- a/api/src/wfl/executor.clj
+++ b/api/src/wfl/executor.clj
@@ -328,7 +328,50 @@
                (firecloud/get-workflow-outputs workspace submission workflow))))]
     (map from-record records)))
 
+;; terra-executor-workflows and terra-executor-workflows-by-status do not return
+;; workflows that are being or have been retied. Why?
+;;
+;; TL/DR: It's simpler.
+;;
+;; The truth is I'm not sure of a *good* reason to do this.
+;;
+;; Some context/terminology:
+;; Workflows are be serialised to the DB as a kind of flattened list of lists,
+;; where there `HEAD` of each inner list is the most recent workflow attempt and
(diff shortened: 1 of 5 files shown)


rexwangcc (approved):
LGTM! (updated the [chart](https://lucid.app/lucidchart/invitations/accept/inv_39e2a1d7-c5c8-4cfc-86a4-7ea92c24a452) accordingly)

api/src/wfl/executor.clj
```diff
@@ -332,7 +332,10 @@
   [tx {:keys [details] :as executor}]
   (when-not (postgres/table-exists? tx details)
     (throw (ex-info "Missing executor details table" {:table details})))
-  (let [query "SELECT * FROM %s WHERE workflow IS NOT NULL ORDER BY id ASC"]
+  (let [query "SELECT * FROM %s
+               WHERE workflow IS NOT NULL
+               AND   retry    IS NULL
```
okotsopoulos, line 337: Can you explain why we want to exclude records with a populated `retry` from these fetches?

ehigham:
> LGTM! (updated the [chart](https://lucid.app/lucidchart/invitations/accept/inv_39e2a1d7-c5c8-4cfc-86a4-7ea92c24a452) accordingly)

Thank you!

api/src/wfl/executor.clj
ehigham: Added a long comment

PR #464, opened 2021-06-25
Fix some namespace issues

rfricke-asymmetrik:
### Purpose
There were a few residual namespace issues resulting from splitting the specs out of the spec.clj file. This rectifies them.

### Changes
Fix straggler namespace issues

- No changes.

### Review Instructions

- No instructions.

### DIFF
--- a/api/src/wfl/api/spec.clj
+++ b/api/src/wfl/api/spec.clj
@@ -8,8 +8,9 @@
             [wfl.module.aou       :as aou]
             [wfl.module.copyfile  :as copyfile]
             [wfl.module.sg        :as sg]
-            [wfl.module.wgs       :as wgs]
             [wfl.module.xx        :as xx]
+            [wfl.module.wgs       :as wgs]
+            [wfl.source           :as source]
             [wfl.module.all       :as all]))
 
 (s/def ::workload-query (s/and (s/keys :opt-un [::all/uuid ::all/project])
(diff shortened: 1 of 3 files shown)


api/src/wfl/api/spec.clj
```diff
@@ -39,9 +39,28 @@
 
 (s/def ::workflow  (s/keys :req-un [::inputs]
                            :opt-un [::all/status ::all/updated ::all/uuid ::all/options]))
+
+;; This is the wrong thing to do. See [1] for more information.
+;; As a consequence, I've included the keys for a covid pipeline as optional
+;; inputs for batch workloads so that these keys are not removed during
+;; coercion.
+;; [1]: https://github.com/metosin/reitit/issues/494
+(s/def :batch/workload-request
```
ehigham, line 48: What motivated this move? It seemed to be in the right place before? Was there some dependency error?

ehigham (approved):

api/src/wfl/api/spec.clj
rfricke-asymmetrik: I attempted to move items, workload-inputs, inputs and workflow into the batch module, but in doing so, created a circular dependency as aou references batch in its implementation. Couldn't require spec in batch either as spec also requires batch.

ehigham: oh noes :'(

tbl3rd: Yes, the prior commit looked OK to me too FWIW.

tbl3rd (approved):

PR #466, opened 2021-06-28
[GH-1409] Add Status Query String to workflows GET request

rfricke-asymmetrik:
### Purpose
Make it easier to filter workflows by status. Also in this PR is work to bring the cromwell statuses up to date.

- https://broadinstitute.atlassian.net/browse/GH-1409
- https://broadinstitute.atlassian.net/browse/GH-1410

### Changes

- Added new spec for a workflow query string
- Added query string to the route for GET workflows
- Modified the get-workflows handler function to call workflows-by-status if status is provided

### Review Instructions

- The workload/:uuid/workflows endpoint should work with status both provided and not provided. It should be verified that the endpoint returns results both when provided a status and not as well as verifying that it returns workflows with the correct status if it is provided.

### DIFF
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -66,11 +66,11 @@
   [request]
   (log/info (select-keys request [:request-method :uri :parameters]))
   (let [uuid (get-in request [:path-params :uuid])
-        {:keys [status]} (get-in request [:parameters :query])]
+        status (get-in request [:parameters :query :status])]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-      (->> (let [w (workloads/load-workload-for-uuid tx uuid)]
-             (cond status (workloads/workflows-by-status tx w status)
-                   :else  (workloads/workflows tx w)))
+      (->> (let [workload (workloads/load-workload-for-uuid tx uuid)]
+             (if status (workloads/workflows-by-status tx workload status)
+                 (workloads/workflows tx workload)))
            (mapv util/to-edn)
            succeed))))
 
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -69,8 +69,7 @@
         status (get-in request [:parameters :query :status])]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
       (->> (let [workload (workloads/load-workload-for-uuid tx uuid)]
-             (if status (workloads/workflows-by-status tx workload status)
-                 (workloads/workflows tx workload)))
+             (if status (workloads/workflows-by-status tx workload status) (workloads/workflows tx workload)))
            (mapv util/to-edn)
            succeed))))
 
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -65,10 +65,12 @@
   "Return the workflows managed by the workload."
   [request]
   (log/info (select-keys request [:request-method :uri :parameters]))
-  (let [uuid (get-in request [:path-params :uuid])]
+  (let [uuid (get-in request [:path-params :uuid])
+        {:keys [status]} (get-in request [:parameters :query])]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-      (->> (workloads/load-workload-for-uuid tx uuid)
-           (workloads/workflows tx)
+      (->> (let [w (workloads/load-workload-for-uuid tx uuid)]
+             (cond status (workloads/workflows-by-status tx w status)
+                   :else  (workloads/workflows tx w)))
            (mapv util/to-edn)
            succeed))))
 
--- a/api/src/wfl/module/all.clj
+++ b/api/src/wfl/module/all.clj
@@ -112,7 +112,7 @@
 (s/def ::pipeline string?)
 (s/def ::project string?)
 (s/def ::release string?)
-(s/def ::status (set (conj (distinct (into cromwell/statuses rawls/statuses)) "skipped")))
+(s/def ::status (set (concat cromwell/statuses rawls/statuses ["skipped"])))
 (s/def ::started ::timestamp)
 (s/def ::stopped ::timestamp)
 (s/def ::updated ::timestamp)
--- a/api/src/wfl/module/all.clj
+++ b/api/src/wfl/module/all.clj
@@ -6,7 +6,8 @@
             [wfl.service.google.storage :as gcs]
             [wfl.util :as util]
             [wfl.wfl :as wfl]
-            [wfl.service.cromwell :as cromwell])
+            [wfl.service.cromwell :as cromwell]
+            [wfl.service.rawls :as rawls])
   (:import [java.util UUID]))
 
 (defn throw-when-output-exists-already!
@@ -111,7 +112,7 @@
 (s/def ::pipeline string?)
 (s/def ::project string?)
 (s/def ::release string?)
-(s/def ::status (set (conj cromwell/statuses "skipped")))
+(s/def ::status (set (conj (distinct (into cromwell/statuses rawls/statuses)) "skipped")))
 (s/def ::started ::timestamp)
 (s/def ::stopped ::timestamp)
 (s/def ::updated ::timestamp)
--- a/api/src/wfl/service/cromwell.clj
+++ b/api/src/wfl/service/cromwell.clj
@@ -20,10 +20,7 @@
   ["Aborting"
    "On Hold"
    "Running"
-   "Submitted"
-   "Queued",
-   "Launching",
-   "Unknown"])
+   "Submitted"])
 
 (def statuses
   "All the statuses a Cromwell workflow can have."
--- a/api/src/wfl/service/cromwell.clj
+++ b/api/src/wfl/service/cromwell.clj
@@ -18,9 +18,11 @@
 (def active-statuses
   "The statuses an active Cromwell workflow can have."
   ["Aborting"
-   "On Hold"
    "Running"
-   "Submitted"])
+   "Submitted"
+   "Queued",
+   "Launching",
+   "Unknown"])
 
 (def statuses
   "All the statuses a Cromwell workflow can have."
--- a/api/src/wfl/service/cromwell.clj
+++ b/api/src/wfl/service/cromwell.clj
@@ -18,6 +18,7 @@
 (def active-statuses
   "The statuses an active Cromwell workflow can have."
   ["Aborting"
+   "On Hold"
    "Running"
    "Submitted"
    "Queued",
--- a/api/src/wfl/service/rawls.clj
+++ b/api/src/wfl/service/rawls.clj
@@ -17,11 +17,11 @@
 
 (def active-statuses
   "The active statuses the Rawls workflow can have."
-  ["Queued"
+  ["Aborting"
    "Launching"
-   "Submitted"
+   "Queued"
    "Running"
-   "Aborting"
+   "Submitted"
    "Unknown"])
 
 (def statuses
--- a/api/src/wfl/service/rawls.clj
+++ b/api/src/wfl/service/rawls.clj
@@ -9,6 +9,25 @@
             [wfl.util             :as util])
   (:import [clojure.lang ExceptionInfo]))
 
+(def final-statuses 
+  "The final statuses the Rawls workflow can have."
+  ["Aborted"
+   "Failed"
+   "Succeeded"])
+
+(def active-statuses
+  "The active statuses the Rawls workflow can have."
+  ["Queued"
+   "Launching"
+   "Submitted"
+   "Running"
+   "Aborting"
+   "Unknown"])
+
+(def statuses
+  "All the statuses the Rawls workflow can have."
+  (into final-statuses active-statuses))
+
 (defn ^:private rawls-url [& parts]
   (let [url (util/de-slashify (env/getenv "WFL_RAWLS_URL"))]
     (str/join "/" (cons url parts))))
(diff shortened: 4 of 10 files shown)


api/src/wfl/api/handlers.clj
```diff
@@ -65,10 +65,12 @@
...
   [request]
   (log/info (select-keys request [:request-method :uri :parameters]))
-  (let [uuid (get-in request [:path-params :uuid])]
+  (let [uuid (get-in request [:path-params :uuid])
+        {:keys [status]} (get-in request [:parameters :query])]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-      (->> (workloads/load-workload-for-uuid tx uuid)
-           (workloads/workflows tx)
+      (->> (let [w (workloads/load-workload-for-uuid tx uuid)]
+             (cond status (workloads/workflows-by-status tx w status)
```
ehigham, line 72: Now I'm not so happy I made this two separate operations.

ehigham: ^ nothing you can do i don't think, just commenting on my mistakes.

api/src/wfl/api/handlers.clj
```diff
@@ -65,10 +65,12 @@
...
+                   :else  (workloads/workflows tx w)))
```
rexwangcc, line 73: I feel like this is similar to `if`, is there a special consideration behind using `cond`?

api/test/wfl/tools/endpoints.clj
```diff
@@ -46,6 +46,13 @@
       (http/get {:headers (auth/get-auth-header)})
       util/response-body-json))
 
+(defn get-workflows-by-status
+  "Query v1 api for all workflows managed by `_workload` with a specific status."
+  [{:keys [uuid] :as _workload} status]
+  (-> (wfl-url "/api/v1/workload/" uuid "/workflows")
```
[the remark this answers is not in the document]
ehigham: Since you're calling the same endpoint, it might be better to augment the existing method to take an optional status parameter. The signature could look like
```clojure
[{:keys [uuid] :as workload} & [status]] 
```
(this is one way to have one optional parameter in your function declarations)

ehigham (approved):

api/src/wfl/service/cromwell.clj
```diff
@@ -18,9 +18,11 @@
 (def active-statuses
   "The statuses an active Cromwell workflow can have."
   ["Aborting"
-   "On Hold"
    "Running"
-   "Submitted"])
+   "Submitted"
+   "Queued",
+   "Launching",
+   "Unknown"])
```
rexwangcc, line 25: I'm not sure if we want to update them in-place, since WFL needs to support both Cromwell and Terra. From what I can tell, the new set of statuses are still specific to Terra?

rexwangcc (commented):
Thank you for updating these! I left my 2 cents.

api/src/wfl/api/handlers.clj
okotsopoulos: (Not PR feedback but sharing as a FYI!)

I know we talked about this in stand-up a bit, but here's how I usually choose between `if`, `when`, and `cond`.

`if` - test one condition with a fallback (the "else" branch).
```
(if true "yay" "womp")
=> "yay"
(if false "yay" "womp")
=> "womp"
```

`if` - test one condition with `nil` as the fallback.
```
(when true "yay")
=> "yay"
(when false "yay")
=> nil
```

`cond` - test multiple conditions.  I introduced a `cond` with a single test case in `executor.clj` but it was because I anticipated a future in which we would be supporting other conditions.
```
(let [check "yay"]
  (cond (= check "yay") "party"
        (= check "womp") "stay home"
        :else "try again"))
=> "party"
(let [check "womp"]
  (cond (= check "yay") "party"
        (= check "womp") "stay home"
        :else "try again"))
=> "stay home"
(let [check "something"]
  (cond (= check "yay") "party"
        (= check "womp") "stay home"
        :else "try again"))
=> "try again"
```

Oftentimes when I'm using an `if` or a `when` I end up preferring `if-let` or `when-let`.  They're good ones to look at in the cheat sheet:
https://clojuredocs.org/clojure.core/if-let
https://clojuredocs.org/clojure.core/when-let

api/src/wfl/api/handlers.clj
```diff
@@ -65,10 +65,12 @@
   "Return the workflows managed by the workload."
   [request]
   (log/info (select-keys request [:request-method :uri :parameters]))
-  (let [uuid (get-in request [:path-params :uuid])]
+  (let [uuid (get-in request [:path-params :uuid])
+        {:keys [status]} (get-in request [:parameters :query])]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-      (->> (workloads/load-workload-for-uuid tx uuid)
-           (workloads/workflows tx)
+      (->> (let [w (workloads/load-workload-for-uuid tx uuid)]
```
okotsopoulos, line 71: Nit: I don't like using `w` as a variable name for a workload, and choose readability over compactness.  This is because we work primarily with workloads and workflows, and get them mixed up frequently even with their full names :)

api/test/wfl/tools/endpoints.clj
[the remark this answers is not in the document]
okotsopoulos: Another FYI is that ^^ is called a variadic function, the other way I've specified optional params is with multi-arity function definitions.  Some more info on both kinds here:

https://theburningmonk.com/2013/09/clojure-multi-arity-and-variadic-functions/

api/src/wfl/api/handlers.clj
```diff
@@ -65,10 +65,12 @@
   "Return the workflows managed by the workload."
   [request]
   (log/info (select-keys request [:request-method :uri :parameters]))
-  (let [uuid (get-in request [:path-params :uuid])]
+  (let [uuid (get-in request [:path-params :uuid])
+        {:keys [status]} (get-in request [:parameters :query])]
```
okotsopoulos, line 69: I think this binding can be made more clear.  Would this work instead?
```
status (get-in request [:parameters :query :status]
```

https://clojuredocs.org/clojure.core/get-in

api/src/wfl/api/handlers.clj
ehigham: I see your point though in this context I'd argue it's pretty clear as he's calling workload operations on w.

api/test/wfl/tools/endpoints.clj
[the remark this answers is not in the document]
rfricke-asymmetrik: Thank you! I went with the variadic route for this simply so that I didn't have duplicated code. Also @ehigham, I added a utils function to trim slashes from the front and back of a url part and added that in a call to map in wfl-url, that way we should be able to leave the endpoint calls as they are and the wfl-url function will clean it up for us. I think this also helps make the url parts more readable with the slashes included.

api/src/wfl/api/handlers.clj
rfricke-asymmetrik: I changed this to a an if. Thank you for the advice.

api/src/wfl/api/handlers.clj
rfricke-asymmetrik: For clarity, I went ahead and updated to workload.

api/src/wfl/api/handlers.clj
rfricke-asymmetrik: Seems like it :)

api/src/wfl/service/cromwell.clj
rfricke-asymmetrik: I added the statuses from the rawls swagger into the rawls service namespace, so now there are statuses in cromwell and rawls.

api/src/wfl/api/handlers.clj
```diff
@@ -65,10 +65,12 @@
...
   (log/info (select-keys request [:request-method :uri :parameters]))
-  (let [uuid (get-in request [:path-params :uuid])]
+  (let [uuid (get-in request [:path-params :uuid])
+        status (get-in request [:parameters :query :status])]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-      (->> (workloads/load-workload-for-uuid tx uuid)
-           (workloads/workflows tx)
+      (->> (let [workload (workloads/load-workload-for-uuid tx uuid)]
+             (if status (workloads/workflows-by-status tx workload status)
+                 (workloads/workflows tx workload)))
```
tbl3rd, line 73: Maybe move the then onto its own line even with the else?

api/src/wfl/module/all.clj
```diff
@@ -111,7 +112,7 @@
 (s/def ::pipeline string?)
 (s/def ::project string?)
 (s/def ::release string?)
-(s/def ::status (set (conj cromwell/statuses "skipped")))
+(s/def ::status (set (conj (distinct (into cromwell/statuses rawls/statuses)) "skipped")))
```
tbl3rd, line 115: Maybe: `(s/def ::status (set (concat cromwell/statuses rawls/statuses ["skipped"])))`
`set` will `distinct` for you.
(I wish we could get rid of `"skipped"`.)

api/src/wfl/service/rawls.clj
```diff
@@ -9,6 +9,25 @@
...
+   "Succeeded"])
+
+(def active-statuses
+  "The active statuses the Rawls workflow can have."
+  ["Queued"
+   "Launching"
+   "Submitted"
+   "Running"
+   "Aborting"
+   "Unknown"])
```
tbl3rd, line 25: These look like they're ordered by time?

api/test/wfl/system/v1_endpoint_test.clj
```diff
@@ -38,6 +38,10 @@
   (run! verify-succeeded-workflow (endpoints/get-workflows workload))
   (workloads/postcheck workload))
 
+(defn ^:private verify-workflows-by-status
+  [workload status]
+  (run! #(is (= (:status %) status)) (endpoints/get-workflows workload status)))
```
tbl3rd, line 43: You can move this down to where it is called.

api/test/wfl/system/v1_endpoint_test.clj
```diff
@@ -296,3 +300,11 @@
...
           (is (inst? started))
           (is (inst? stopped)))))))
+
+(deftest ^:parallel test-workflows-by-status
+  (testing "Get workflows by status"
+    (let [workload (first (endpoints/get-workloads))
+          workflows (endpoints/get-workflows workload)]
+      (->> (map #(get % :status) workflows)
+           (distinct)
+           (run! #(verify-workflows-by-status workload %))))))
```
tbl3rd, line 310: I *think* `#(get % :status)` is just `:status` here.

api/test/wfl/tools/endpoints.clj
```diff
@@ -10,7 +10,7 @@
   "The WFL server URL to test."
   [& parts]
   (let [url (util/de-slashify (env/getenv "WFL_WFL_URL"))]
-    (str/join "/" (cons url parts))))
+    (str/join "/" (cons url (map #(util/trim-slashes %) parts)))))
```
tbl3rd, line 13: @rexwangcc: This `wfl-url` is gettin craazy.
We shouldn't even need `de-slashify` since we control `WFL_WFL_URL`.
Make sure that `WFL_WFL_URL` doesn't end in `/` wherever we set it.
Then just `(str/join "/" (cons (env/getenv "WFL_WFL_URL") parts))`
What am I missing?

api/test/wfl/tools/endpoints.clj
```diff
@@ -41,9 +41,9 @@
 
 (defn get-workflows
   "Query v1 api for all workflows managed by `_workload`."
-  [{:keys [uuid] :as _workload}]
+  [{:keys [uuid] :as _workload} & [status]]
   (-> (wfl-url "/api/v1/workload/" uuid "/workflows")
-      (http/get {:headers (auth/get-auth-header)})
+      (http/get (merge {:headers (auth/get-auth-header)} (when status {:query-params {:status status}})))
       util/response-body-json))
```
tbl3rd, line 47: Maybe just:
``` clojure
(defn get-workflows
  "Query v1 api for all workflows managed by `_workload`."
  [{:keys [uuid] :as _workload} & [status]]
  (-> (wfl-url "api" "v1" "workload" uuid "workflows")
      (http/get (merge {:headers (auth/get-auth-header)}
                       (when status {:query-params {:status status}})))
      util/response-body-json))
```

BTW: I like the `(merge … (when …))`. It never occurs to me to do that.

api/test/wfl/unit/utils_test.clj
```diff
@@ -74,3 +74,11 @@
...
                       [:membership "flowcell_set_id" membership]]]
       (run! go parameters))))
+
+(deftest test-trim-slashes
+  (is (= "wfl" (util/trim-slashes "/wfl")))
+  (is (= "wfl" (util/trim-slashes "/wfl/")))
+  (is (= "wfl" (util/trim-slashes "wfl/")))
+  (is (= "wfl" (util/trim-slashes "wfl")))
+  (is (= "" (util/trim-slashes "/")))
+  (is (= "" (util/trim-slashes "///"))))
```
tbl3rd, line 84: We can get rid of `util/trim-slashes` and its test if you agree with my comments above.

tbl3rd (approved):
… with some suggestions.

api/test/wfl/tools/endpoints.clj
rfricke-asymmetrik: @tbl3rd the trim-slashes was meant for the url parts, if they are written in with slashes, the join inserts double slashes like workload//:uuid//workflows

tbl3rd: Why would they have /s?
Doc, it hurts when I do this.
Then don't do that.

rfricke-asymmetrik: FWIW, I don't think the previous implementation really was breaking anything anyway. Was just something Ed pointed out.

PR #467, opened 2021-06-28
[GH-1301] Slack watchers about workload statuses.

rexwangcc:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1301

### Changes

- Add a Slack module that interacts with Slack API.
- Implement an `agent` queue that takes all notifications.
- Unit tests + integration test that sends to a real channel `#hornet-slack-app-testing`.

### Review Instructions

- I'm reading through the recents changes to the queue/executor services right now to figure out a better place for the notification loop to live in than the update loop, please comment if you have any insights!


api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,41 @@
...
+
+(defn ^:private api-url
+  "API URL for Slack."
+  [& parts]
+  (let [url "https://slack.com/api"]
+    (str/join "/" (cons url parts))))
+
+;; Slack Bot User token obtained from Vault
+(defonce ^:private token
+  (delay (:bot-user-token (#'env/vault-secrets "secret/dsde/gotc/dev/wfl/slack"))))
```
okotsopoulos, line 17: Is this path consistent across environments (I saw `…gotc/dev…` and thought of the prod env)?

api/src/wfl/server.clj
```diff
@@ -76,9 +77,22 @@
...
+                                 (str/join " "
+                                   ["NOTE: WorkFlow Launcher failed to update"
+                                    "a workload %s you are watching! Details"
+                                    "shown below: \n %s"]) uuid exception)
+        notify-with-slack      (fn [slack-channel-watcher]
+                                  (slack/post-message slack-channel-watcher slack-msg))
+        slack-channel-watcher? (fn [watcher]
+                                 (not (util/email-address? watcher)))
+        slack-channel-watchers (filter slack-channel-watcher? watchers)]
+    (log/info "notifying: " slack-channel-watchers)
```
ehigham, line 94: I'd use `split-with` to make the separation clearer: 
```clojure
(let [[channels _] (split-with slack-channel? watchers)]
  (run! notify-with-slack channels))
```

https://clojuredocs.org/clojure.core/split-with

ehigham: I also don't this we should be doing this in the "update loop" (or in an exception handler therein) - we should probably defer to another queue/execution service.

api/src/wfl/service/slack.clj
rexwangcc: This is a really good point! I think we will stick to one Slack token for all environments (we also have one slack Bot and one slack Org), but if it brings confusion, maybe we should move it to a different Vault path? (such as `secret/dsde/gotc/wfl`? ) I realized we have similar issues with other token/credentials as well…

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,41 @@
...
+
+(defn ^:private header
+  "Auth header for interacting with Slack API from `token`."
+  [token]
+  {:Authorization (str "Bearer " token)})
```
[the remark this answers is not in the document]
rexwangcc: ah nice catch… (lol that there's a real Github User behind @token …)

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,89 @@
...
+
+(defn start-notification-loop
+  "Return a future that listens at `agent` and
+   sends out notifications."
+  [agent]
+  (future (while true
+            (send-off agent send-notification)
+            (util/sleep-seconds 1))))
+
+(comment
```
okotsopoulos, line 86: Probably mean to take this block out :)

api/src/wfl/server.clj
```diff
@@ -76,9 +77,21 @@
...
+  ;; FIXME: add permission checks for slack-channel-watchers
   {:pre [(some? watchers)]}
-  (log/info (str/join " " ["notifying: " watchers])))
+  (let [slack-msg              (format
+                                (str/join " "
+                                          ["NOTE: WorkFlow Launcher failed to update"
+                                           "a workload %s you are watching! Details"
+                                           "shown below: \n %s"]) uuid exception)
+        slack-channel-watcher? (fn [watcher]
+                                 (not (util/email-address? watcher)))
```
ehigham, line 91:
```suggestion
        slack-channel-watcher? (complement util/email-address?)
```

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,84 @@
...
+  [channel message]
+  {:pre [(valid-channel-id? channel)]}
+  (let [url (api-url "chat.postMessage")
+        data {:channel channel
+              :text message}]
+    (-> (http/post url {:headers      (header @token)
+                        :content-type :application/json
+                        :body         (json/write-str data)})
+        :body
+        slack-api-raise-for-status)))
```
ehigham, line 54: How about removing the URL from the `let`?
```suggestion
  (let [data {:channel channel :text message}]
    (-> (api-url "chat.postMessage")
        (http/post {:headers      (header @token)
                    :content-type :application/json
                    :body         (json/write-str data)})
        :body
        slack-api-raise-for-status)))
```

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,84 @@
...
+
+;; Create the agent queue and attach a watcher
+;; FIXME: make the queue persistent
+(def notifier (agent (PersistentQueue/EMPTY)))
```
ehigham, line 58: Is this creating an agent/threads when the namespace loads? Maybe we want to move this into a function and have something manage the lifetime of the agent?

api/src/wfl/server.clj
tbl3rd: Does WFL support mixing Slack channels and Emails in the watchers list?

ehigham: The list currently supports emails - my fault for adding the potential for functionality without actual support for it. We'll need to keep that to avoid reitit errors when returning all workloads in the db. Suggest we extend watchers to contain both emails and extra stuff. See my comment below for one way of doing this.

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,84 @@
...
+      (pop queue))
+    queue))
+
+(defn start-notification-loop
+  "Return a future that listens at `agent` and
+   sends out notifications."
+  [agent]
+  (future (while true
+            (send-off agent send-notification)
+            (util/sleep-seconds 1))))
```
tbl3rd, line 84: Might want to bind this for future cancellation.

api/test/wfl/integration/slack_test.clj
```diff
@@ -0,0 +1,26 @@
...
+  (testing "notification can actually be sent to slack"
+    (slack/add-notification testing-agent (testing-slack-notification))
+    (await testing-agent)
+    (with-redefs-fn
+      {#'slack/start-notification-loop
+       #(future (send-off % #'slack/send-notification)
+                (util/sleep-seconds 1))
+       #'slack/post-message
+       post-message-assertion-wrapper}
+      #(slack/start-notification-loop testing-agent))))
```
tbl3rd, line 26: When I run `wfl.integration.slack-test/test-send-notification-to-a-slack-channel`, I see `Test ran without assertions. Did you forget an (is ...)?` but I don't understand why yet.
(Maybe Kaocha doesn't see the `is` in the scope of the `deftest`.)

api/src/wfl/server.clj
ehigham: I think that ship has sailed and would be more complicated to implement. We could make a breaking change of course!

api/src/wfl/service/slack.clj
ehigham: It allocates them regardless of whether or not you're using them?

tbl3rd: ¯\_(ツ)_/¯ That's how I read it.  You could jstack before and after to check if you care.

api/src/wfl/server.clj
tbl3rd: Who uses the email notification now?

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,84 @@
...
+  (and #(not (util/email-address? channel-id))
+       (str/starts-with? channel-id "C")))
+
+(defn ^:private slack-api-raise-for-status
+  "Slack API has its own way of reporting
+   statuses, so we need to parse the `body`
+   to raise for status."
+  [body]
+  (let [response (json/read-str body)]
+    (prn (get response "error"))
```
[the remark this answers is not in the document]
rexwangcc: oops that's for debugging forgot to remove it 🤦

api/test/wfl/integration/slack_test.clj
rexwangcc: Yeah, seems Kaocha cannot see it, I will change how it's tested!

api/src/wfl/server.clj
```diff
@@ -96,7 +93,8 @@
                   (catch UserException e
                     (log/warn (format "Error updating workload %s" uuid))
                     (log/warn e)
-                    (notify-watchers watchers uuid e))))))
+                    ;; need to deserialize the watchers array
+                    (slack/notify-watchers (map read-string watchers) uuid e))))))
```
ehigham, line 97: I think this ``read-string` should be moved into the `load-workload-impl` so that we don't need to deserialise elsewhere.

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,98 @@
...
+      (pop queue))
+    queue))
+
+;; FIXME: add permission checks for slack-channel-watchers
+;;
+(defn notify-watchers
+  "Notify `watchers` of an `exception` in the workload with `uuid`."
+  [watchers uuid exception]
+  (let [message        (format "Workload %s update threw %s" uuid (.getMessage exception))
+        [_emails channels] (split-with email-watcher? watchers)]
```
ehigham, line 85: I think this logic should be moved into `server.clj` and call `slack/notify-watchers` with "slack channel watchers" (in this case, `[tag channel]` pairs). I don't think this namespace should be concerned with separating out other types of watchers.

api/test/wfl/tools/workloads.clj
```diff
@@ -23,6 +23,10 @@
 (def email
   (delay (:email (gcs/userinfo {:headers (auth/get-auth-header)}))))
 
+(def watchers
+  [["EmailAddress" "hornet@broadinstitute.org"]
```
ehigham, line 27: I don't think you can change this without making a breaking change with what's in prod. I'd keep `watchers` to be an email address or a `["SlackChannel" "{channel-id}"]` pair.

```
Watcher -> EmailAddress | [ "SlackChannel" ChannelId ] 
```

api/src/wfl/server.clj
rexwangcc: that's a good idea, will do!

api/test/wfl/tools/workloads.clj
rexwangcc: My concern is that we will never have a chance to standardize the emails into tagged pairs, especially now we are not sending anything to any email watchers in prod. Do you think the "breaking change" can be done with a Postgres liquibase migration? e.g. use a cast or something similar to migrate all existing emails string records in prod to tagged pairs. It may challenge my psql skill though.

ehigham: Yes, that's the point - we won't and don't have to. It's in the dev and prod database as this. If you can do this in  db migration, fine. I think making a user type `["EmailAddress"` before every email address is a pain.

rexwangcc: I see, all make sense to me, will probably make it `Watcher -> EmailAddress | [ "SlackChannel" ChannelId ] `, as it leaves minimal work for myself to update. (I was just having OCD seeing them differ (string and a pair) 🤦 )

tbl3rd: Howzabout?
Watcher -> ["SlackChannel" ChannelId] | ["EmailAddress" EmailAddress] | EmailAddress

api/test/wfl/integration/jdbc_test.clj
```diff
@@ -7,17 +7,24 @@
 (use-fixtures :once fixtures/temporary-postgresql-database)
```
tbl3rd, line 7: Is this to support the new `:watchers` column?

api/test/wfl/integration/slack_test.clj
```diff
@@ -0,0 +1,34 @@
...
+  (:require [clojure.test :refer :all]
+            [wfl.service.slack :as slack]
+            [wfl.util :as util])
+  (:import [clojure.lang PersistentQueue]))
+
+(def ^:private testing-agent (agent (PersistentQueue/EMPTY)))
+(def ^:private testing-slack-channel "C026PTM4XPA")
+(defn ^:private testing-slack-notification []
+  {:channel testing-slack-channel :message (format "WFL Integration Test Message: %s" (util/utc-now))})
+(def ^:private notify-promise (promise))
```
tbl3rd, line 11: I want some blank lines. ;-)

api/test/wfl/integration/slack_test.clj
```diff
@@ -0,0 +1,34 @@
+(ns wfl.integration.slack-test
+  (:require [clojure.test :refer :all]
+            [wfl.service.slack :as slack]
+            [wfl.util :as util])
+  (:import [clojure.lang PersistentQueue]))
+
+(def ^:private testing-agent (agent (PersistentQueue/EMPTY)))
+(def ^:private testing-slack-channel "C026PTM4XPA")
+(defn ^:private testing-slack-notification []
+  {:channel testing-slack-channel :message (format "WFL Integration Test Message: %s" (util/utc-now))})
```
tbl3rd, line 10: A newline before `:message` would be nice.

api/test/wfl/integration/slack_test.clj
```diff
@@ -0,0 +1,34 @@
...
+                      (deliver notify-promise false))]
+      (slack/post-message channel message callback)
+      (pop queue))
+    queue))
+
+(add-watch testing-agent :watcher
+           (fn [_key _ref _old-state new-state]
+             (when (seq new-state)
+               (testing "notification is added to agent"
+                 (is (= (:channel (first (seq new-state))) testing-slack-channel))))))
```
tbl3rd, line 27: Why `(seq new-state)`?

api/test/wfl/integration/slack_test.clj
```diff
@@ -0,0 +1,34 @@
...
+
+(deftest test-send-notification-to-a-slack-channel
+  (with-redefs-fn {#'slack/send-notification mock-send-notification}
+    #(do (slack/add-notification testing-agent (testing-slack-notification))
+         (send-off testing-agent #'slack/send-notification)
+         (testing "notification can actually be sent to slack"
+           (is (true? @notify-promise) "Slack notification promise should be fulfilled")))))
```
tbl3rd: Add a newline to the end of file.
Use `with-redefs` and lose the `#`?
I believe the promise works by blocking the calling thread on the agent's thread.

api/test/wfl/unit/slack_test.clj
```diff
@@ -0,0 +1,16 @@
...
+(defn ^:private testing-slack-notification []
+  {:channel "C000BOGUS00" :message "WFL Test Message"})
+
+(deftest test-add-notification
+  (testing "notification can be added to the agent queue properly"
+    (let [num-msg 10]
+      (dotimes [_ num-msg]
+        (slack/add-notification testing-agent (testing-slack-notification)))
+      (await testing-agent)
+      (is (= num-msg (count (seq @testing-agent)))))))
```
tbl3rd, line 16: `=` -=> `==`
Again, why `seq`?

tbl3rd (approved):

api/src/wfl/module/batch.clj
```diff
@@ -124,7 +124,7 @@
         setter "UPDATE workload SET pipeline = ?::pipeline WHERE id = ?"
         [{:keys [id]}]
         (-> workload-request
-            (select-keys [:creator :executor :input :output :project :watchers :labels])
+            (select-keys [:creator :executor :input :output :project :watchers :labels :watchers])
```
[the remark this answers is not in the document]
rexwangcc: ah nice catch, I didn't realize it. Will fix.

api/src/wfl/module/all.clj
```diff
@@ -138,5 +140,8 @@
 (s/def ::name string?)
 (s/def ::methodConfiguration (s/and string? util/terra-namespaced-name?))
 (s/def ::methodConfigurationVersion integer?)
-(s/def ::watchers (s/* util/email-address?))
+(s/def ::watcher
+  (s/or :SlackChannel slack/email-watcher?
+        :EmailAddress slack/slack-channel-watcher?))
```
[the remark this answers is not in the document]
rexwangcc: nice catch , will fix!

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,101 @@
...
+(defn ^:private header
+  "Auth header for interacting with Slack API from `token`."
+  [token]
+  {:Authorization (str "Bearer " token)})
+
+(defn ^:private valid-channel-id?
+  [channel-id]
+  ; FIXME: suppress warning `javax.mail.internet.AddressException: Missing final '@domain'`
+  (and #(not (util/email-address? channel-id))
+       (str/starts-with? channel-id "C")))
```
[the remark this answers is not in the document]
rexwangcc: that is a good point, so far channels I have seen started with `\C`, but there's not a regex from slack defined it. will loosen the restrictions.

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,101 @@
...
+
+(defn slack-channel-watcher? [s]
+  (when-let [[tag value] (seq s)]
+    (and (= "SlackChannel" tag) (valid-channel-id? value))))
```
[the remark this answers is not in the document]
rexwangcc: interesting, turns out I don't need seq, thought it's needed for destructing, updated

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,101 @@
...
+
+(defn email-watcher? [s]
+  (when-let [[tag value] (seq s)]
+    (and (= "EmailAddress" tag) (util/email-address? value))))
```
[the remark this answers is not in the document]
rexwangcc: same as above, updated!

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,101 @@
...
+   (let [data {:channel channel
+               :text    message}]
+     (-> "https://slack.com/api/chat.postMessage"
+         (http/post {:headers      (header @token)
+                     :content-type :application/json
+                     :body         (json/write-str data)})
+         :body
+         slack-api-raise-for-status)))
+  ([channel message callback]
+   (callback (post-message channel message))))
```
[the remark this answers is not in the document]
rexwangcc: I need it for testing, all other methods I have tried failed ;(

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,101 @@
...
+(defn ^:private slack-api-raise-for-status
+  "Slack API has its own way of reporting
+   statuses, so we need to parse the `body`
+   to raise for status."
+  [body]
+  (let [response (json/read-str body)]
+    (when-not (get response "ok")
+      (throw (ex-info "failed to notify via Slack"
+                      {:error (get response "error")})))
+    response))
```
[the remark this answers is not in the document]
rexwangcc: it's more like personal preference, I will update it! And no it's not that big.

api/test/wfl/integration/jdbc_test.clj
rexwangcc: hmm, no, this was here for a while 🤔

api/test/wfl/integration/slack_test.clj
rexwangcc: You'll get plenty of them :'p

api/test/wfl/integration/slack_test.clj
rexwangcc: since the agent is created by `(agent (PersistentQueue/EMPTY))`, without `seq` it will not print something readable 🤔

api/test/wfl/unit/slack_test.clj
rexwangcc: sorry. but i'm not sure why we should use https://clojuredocs.org/clojure.core/== here?

api/test/wfl/integration/slack_test.clj
rexwangcc: It's super weird that once i switch to `with-redefs`, the test will timeout, even though the notification is sent.

api/test/wfl/integration/slack_test.clj
tbl3rd: OK.

api/test/wfl/integration/slack_test.clj
tbl3rd: OK. We'll figure it out later.

api/src/wfl/module/all.clj
```diff
@@ -138,5 +141,19 @@
 (s/def ::name string?)
 (s/def ::methodConfiguration (s/and string? util/terra-namespaced-name?))
 (s/def ::methodConfigurationVersion integer?)
-(s/def ::watchers (s/* util/email-address?))
+(s/def ::email util/email-address?)
+(s/def ::legacyWatcher ::email)
+(s/def ::taggedWatcher (s/or :email slack/email-watcher?
+                             :slack slack/slack-channel-watcher?))
```
tbl3rd, line 147: This is OK, but naming these keys with the actual tags (`:EmailAddress`, for example) might make them easier to find and interpret in spec errors.

api/src/wfl/module/all.clj
```diff
@@ -138,5 +141,19 @@
...
+        :tagged_watcher  ::taggedWatcher))
+(s/def ::watchers (s/* ::watcher))
 (s/def ::workspace (s/and string? util/terra-namespaced-name?))
+
+(defn parse-watcher
+  "Parse watchers based on specs."
+  [watcher]
+  (if (s/valid? ::legacyWatcher watcher)
+    watcher
+    (edn/read-string watcher)))
```
tbl3rd, line 159: Use `conform`?
```clojure
  (s/conform ::watcher ["EmailAddress" "the-mgt@omf.org"])
  ;; => [:tagged_watcher [:email ["EmailAddress" "the-mgt@omf.org"]]]
  (s/conform ::watcher ["SlackChannel" "Fnord"])
  ;; => :clojure.spec.alpha/invalid
  (s/conform ::watcher ["SlackChannel" "ChannelOfBob"])
  ;; => [:tagged_watcher [:slack ["SlackChannel" "ChannelOfBob"]]]
  (s/conform ::watcher "the-mgt@omf.org")
  ;; => [:legacy_watcher "the-mgt@omf.org"]
```
That will even tell you how to interpret each watcher spec.

api/src/wfl/module/all.clj
```diff
@@ -81,6 +83,7 @@
                                     :release  release
                                     :uuid     (UUID/randomUUID)
                                     :version  version
+                                    :watchers (mapv prn-str watchers)
```
tbl3rd, line 86: I think we want just `pr-str` here.

api/src/wfl/module/batch.clj
```diff
@@ -138,7 +138,10 @@
 (defn load-batch-workload-impl
   "Load workload metadata + trim any unused vars."
   [_ workload]
-  (into {:type :workload} (filter second workload)))
+  (update
+   (into {:type :workload} (filter second workload))
+   :watchers
+   #(mapv all/parse-watcher %)))
```
tbl3rd, line 144: That is also `(partial mapv all/parse-watcher)`.

api/src/wfl/module/covid.clj
```diff
@@ -75,6 +75,7 @@
                  vec))]
     (-> (update request :labels combine-labels)
         (select-keys [:creator :watchers :labels :project])
+        (update :watchers #(mapv prn-str %))
```
tbl3rd, line 78: Again: `pr-str`, I think.

api/src/wfl/module/covid.clj
```diff
@@ -111,6 +112,8 @@
                       :sink     (sink/load-sink! tx workload)}]
     (as-> workload $
       (select-keys $ workload-metadata-keys)
+      ;; need to deserialize the watchers array
+      (update $ :watchers #(mapv all/parse-watcher %))
```
tbl3rd, line 116: The comment can go; it just confuses.  Consider:
```
;; need to parse the watchers sequence
(update $ :watchers (partial mapv all/deserialize-watcher))
```
Call it either parse or deserialize, but not both.

api/src/wfl/server.clj
```diff
@@ -89,14 +86,16 @@
   (letfn [(do-update! [{:keys [id uuid]}]
             (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
               (let [{:keys [watchers] :as workload}
-                    (workloads/load-workload-for-id tx id)]
+                    (workloads/load-workload-for-id tx id)
+                    ;; we don't support notifying emails yet
+                    slack-watchers (filter slack/slack-channel-watcher? watchers)]
                 (log/info (format "Updating workload %s" uuid))
```
tbl3rd, line 92: You need to lift this log higher now, I think.

api/src/wfl/server.clj
```diff
@@ -89,14 +86,16 @@
   (letfn [(do-update! [{:keys [id uuid]}]
             (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
               (let [{:keys [watchers] :as workload}
-                    (workloads/load-workload-for-id tx id)]
+                    (workloads/load-workload-for-id tx id)
+                    ;; we don't support notifying emails yet
```
tbl3rd, line 90: This comment also confuses because `start-workload-manager` doesn't know about email.
All it knows is Slack notification.
Actually, we should consider dispatching the notifications generically, but not now.

api/src/wfl/service/slack.clj
```diff
@@ -0,0 +1,95 @@
...
+  (and (not (util/email-address? channel-id))
+       (str/starts-with? channel-id "C")))
+
+(defn slack-channel-watcher? [s]
+  (when-let [[tag value] s]
+    (and (= "SlackChannel" tag) (valid-channel-id? value))))
+
+(defn email-watcher? [s]
+  (when-let [[tag value] s]
+    (and (= "EmailAddress" tag) (util/email-address? value))))
```
tbl3rd, line 28: Don't need these 3 predicates if we call `conform` in the right place.

api/test/wfl/integration/jdbc_test.clj
tbl3rd: But why then? ;-)

api/test/wfl/integration/slack_test.clj
```diff
@@ -0,0 +1,41 @@
...
+
+;; This is test is flaky on Github Actions but works fine locally
+;;
+(deftest ^:kaocha/pending test-send-notification-to-a-slack-channel
+  (with-redefs-fn {#'slack/send-notification mock-send-notification}
+    #(do (slack/add-notification testing-agent (testing-slack-notification))
+         (send-off testing-agent #'slack/send-notification)
+         (testing "notification can actually be sent to slack"
+           ;; block for at most 10000ms
+           (is (true? (deref notify-promise 10000 :timeout)) "Slack notification promise should be fulfilled")))))
```
tbl3rd, line 41: TIL: `:kaocha/pending`.
Again, lose the "block" comment.
Change the assertion message to `"Waited 10s for notification."` or sumpin.

api/test/wfl/unit/slack_test.clj
tbl3rd: Because `count` always returns an integer.
`=` is OK, but `==` conveys a bit more.

api/test/wfl/unit/spec_test.clj
```diff
@@ -49,9 +49,11 @@
...
+            [[valid?   ["EmailAddress" "hornet@broadinstitute.org"]]
              ;; From tbl: https://www.netmeister.org/blog/email.html
-             [valid?   "'*+-/=?^_`{|}~#$@[IPv6:2001:470:30:84:e276:63ff:fe72:3900]"]
-             [invalid? "foo"]]]
-      (is (test? ::all/watchers [email]) (format "failed: %s" email)))))
+             [valid?   ["EmailAddress" "'*+-/=?^_`{|}~#$@[IPv6:2001:470:30:84:e276:63ff:fe72:3900]"]]
+             [valid?   ["SlackChannel" "C000XXX0XXX"]]
+             [invalid? "foo"]
+             [invalid? ["hornet@broadinstitute.org"]]]]
+      (is (test? ::all/watchers [watcher]) (format "failed: %s" watcher)))))
```
tbl3rd, line 59: Don't need `test-watchers-spec` if we use `conform`.
(-Otherwise, we're just testing Clojure spec for bugs!-)

tbl3rd (approved):
Sorry, and thanks.

api/src/wfl/module/all.clj
rexwangcc: Good idea, will change!

api/src/wfl/module/all.clj
rexwangcc: TIL

api/src/wfl/module/all.clj
rexwangcc: `watchers` is a `TEXT[]` in DB, so `pr-str` will always give us a TEXT[1] while `(mapv prn-str watchers)` gives TEXT[n], either works but I prefer latter as there are already stuff (one or a few emails per record) in prod DB, and latter doesn't require paying too much on migrations, what do you think?

api/src/wfl/module/covid.clj
rexwangcc: Same reason as above?

api/src/wfl/server.clj
rexwangcc: Agree!

api/src/wfl/server.clj
rexwangcc: Yeah, I thought about making `all/deserialize-watcher` a multimethod, but didn't bother as we only support slack…

api/src/wfl/module/all.clj
rexwangcc: For example, latter will store something like:
```clojure
["hornet@broadinstitute.org","[\"EmailAddress\" \"hornet@broadinstitute.org\"]","[\"SlackChannel\" \"C026PTM4XPA\"]"]
```
into DB: 
```sql
{"hornet@broadinstitute.org","[\"EmailAddress\" \"hornet@broadinstitute.org\"]","[\"SlackChannel\" \"C026PTM4XPA\"]"}
```

api/src/wfl/module/all.clj
rexwangcc: TIL, but I'm not sure what extra value `conform` brings here than `s/valid?`, the deserializer will need to understand both legacy and new watchers, and in reality it deals with stuff like:

```
["hornet@broadinstitute.org"
"[\"EmailAddress\" \"hornet@broadinstitute.org\"]"
"[\"SlackChannel\" \"C026PTM4XPA\"]"]
```
the `if` will binary classify them into legacy: `"hornet@broadinstitute.org"` and tagged (but edn-fied) watcher: `"[\"EmailAddress\" \"hornet@broadinstitute.org\"]"` and `"[\"SlackChannel\" \"C026PTM4XPA\"]"`, how does conform validate the latter without trying to parse them first?

api/src/wfl/module/all.clj
tbl3rd: Yeah, FTR: I did not spot this wrinkle.  Sorry.

api/src/wfl/module/covid.clj
tbl3rd: Yep. We don't want the `\newline`.

api/src/wfl/service/slack.clj
tbl3rd: FTR: I was kinda wrong about this.

rexwangcc:
FTR: based on our offline discussion, we moved forward with using:

```
[["email" "someone@acm.com"]
 ["slack" "CXXXXXXXX"]]
```

as the watchers of a workload, which requires us to do a DB cleanup + migration before deploying the next version of WFL. This gives us standardization on watchers from now on, at the expense of typing extra `email` for email watchers. But as we don't support notifying emails, this should be fine.

tbl3rd:
Cool.

rexwangcc:
> Cool.

Shall I merge? :p

PR #468, opened 2021-06-29
GH-1397: Sweep and snapshot dataset row IDs that were missed when polling a dataset

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1397

### Changes

- Document Swagger validation of API specs.
- Mock snapshot missing TDR rows.
- Expand query interval and filter out old rows.

### Review Instructions

- Start with wfl.integration.source-test/test-backfill-tdr-source.
- Evaluate how it mocks the problem.
- In wfl.source, evaluate how changes to find-new-rows and find-and-snapshot-new-rows address the problem.
- Perhaps back out wfl.source commits and watch the test-backfill-tdr-source test fail for more understanding.

### DIFF
--- a/api/src/wfl/source.clj
+++ b/api/src/wfl/source.clj
@@ -511,5 +511,4 @@
                  :end_time #inst "2021-07-21T18:12:58.077575000-00:00",
                  :id 3,
                  :consumed nil,
-                 :snapshot_creation_job_id "mock_job_id_2"}])
-  )
+                 :snapshot_creation_job_id "mock_job_id_2"}]))
--- a/api/src/wfl/source.clj
+++ b/api/src/wfl/source.clj
@@ -13,7 +13,7 @@
             [wfl.module.all        :as all])
   (:import [clojure.lang ExceptionInfo]
            [java.sql Timestamp]
-           [java.time Instant OffsetDateTime ZoneId]
+           [java.time OffsetDateTime ZoneId]
            [java.time.format DateTimeFormatter]
            [java.time.temporal ChronoUnit]
            [wfl.util UserException]))
@@ -158,9 +158,9 @@
       (assoc source :dataset dataset))))
 
 (defn combine-tdr-source-details
-  "Reduce to `result` by combining the Terra Data Repo source `detail`."
+  "Reduce to `result` by combining the Terra Data Repo source `_detail`."
   [result {:keys [datarepo_row_ids end_time snapshot_creation_job_status
-                  start_time] :as detail}]
+                  start_time] :as _detail}]
   (if (= "succeeded" snapshot_creation_job_status)
     (-> result
         (update :datarepo_row_ids into          datarepo_row_ids)
@@ -169,9 +169,9 @@
     result))
 
 (defn ^:private find-new-rows
-  "Query TDR for rows within `interval` that are new to `source`."
+  "Query TDR for rows within `_interval` that are new to `source`."
   [{:keys [dataset details table column] :as source}
-   [begin end                            :as interval]]
+   [begin end                            :as _interval]]
   (log/debug (format "%s Looking for rows in %s.%s between [%s, %s]..."
                      (log-prefix source) (:name dataset) table begin end))
   (let [wfl   (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
--- a/api/src/wfl/source.clj
+++ b/api/src/wfl/source.clj
@@ -158,7 +158,9 @@
       (assoc source :dataset dataset))))
 
 (defn combine-tdr-source-details
-  "Reduce to `result` by combining the Terra Data Repo source `_detail`."
+  "Reduce to `result` by combining the Terra Data Repo source `_detail`.
+  Collect `datarepo_row_ids` already seen while preserving the latest
+  `end_time` and the earliest `start_time`."
   [result {:keys [datarepo_row_ids end_time snapshot_creation_job_status
                   start_time] :as _detail}]
   (if (#{"running" "succeeded"} snapshot_creation_job_status)
@@ -474,41 +476,3 @@
 (defoverload stage/done? tdr-snapshot-list-type  tdr-snapshot-list-done?)
 
 (defoverload util/to-edn tdr-snapshot-list-type tdr-snapshot-list-to-edn)
-
-(comment
-  (def details [{:updated #inst "2021-07-21T18:12:58.578753000-00:00",
-                 :start_time #inst "2021-07-21T18:12:58.068140000-00:00",
-                 :snapshot_id "47cd816d-5c9f-4468-bc2a-cee00cfffe89",
-                 :datarepo_row_ids
-                 ["e60f8156-c35c-4f10-b8cf-21fb5905d68b"
-                  ,,,
-                  "8730e957-8e55-4f8a-9c27-eb6816098e5c"],
-                 :snapshot_creation_job_status "succeeded",
-                 :end_time #inst "2021-07-21T18:12:58.077575000-00:00",
-                 :id 1,
-                 :consumed nil,
-                 :snapshot_creation_job_id "mock_job_id_0"}
-                {:updated #inst "2021-07-21T18:12:58.709609000-00:00",
-                 :start_time #inst "2021-07-21T18:12:58.068140000-00:00",
-                 :snapshot_id "17c00f44-3404-4401-aa11-c6f1c84b7192",
-                 :datarepo_row_ids
-                 ["4eab02fb-4d63-4fc1-b487-22db54a90d2f"
-                  ,,,
-                  "5f5862ae-f579-4083-9b25-b039780e1b6d"],
-                 :snapshot_creation_job_status "succeeded",
-                 :end_time #inst "2021-07-21T18:12:58.077575000-00:00",
-                 :id 2,
-                 :consumed nil,
-                 :snapshot_creation_job_id "mock_job_id_1"}
-                {:updated #inst "2021-07-21T18:12:58.775861000-00:00",
-                 :start_time #inst "2021-07-21T18:12:58.068140000-00:00",
-                 :snapshot_id "dac97a4d-ef77-46da-8956-8d2354d68bcd",
-                 :datarepo_row_ids
-                 ["f25f0764-5fcc-4376-b9dc-6e78f5d8e6cf"
-                  ,,,
-                  "6ce89f55-6d27-4146-9ff7-bd47226818b9"],
-                 :snapshot_creation_job_status "succeeded",
-                 :end_time #inst "2021-07-21T18:12:58.077575000-00:00",
-                 :id 3,
-                 :consumed nil,
-                 :snapshot_creation_job_id "mock_job_id_2"}]))
--- a/api/src/wfl/source.clj
+++ b/api/src/wfl/source.clj
@@ -161,7 +161,7 @@
   "Reduce to `result` by combining the Terra Data Repo source `_detail`."
   [result {:keys [datarepo_row_ids end_time snapshot_creation_job_status
                   start_time] :as _detail}]
-  (if (= "succeeded" snapshot_creation_job_status)
+  (if (#{"running" "succeeded"} snapshot_creation_job_status)
     (-> result
         (update :datarepo_row_ids into          datarepo_row_ids)
         (update :end_time         util/latest   end_time)
--- a/api/src/wfl/source.clj
+++ b/api/src/wfl/source.clj
@@ -3,6 +3,7 @@
             [clojure.spec.alpha    :as s]
             [clojure.set           :as set]
             [clojure.tools.logging :as log]
+            [wfl.debug]
             [wfl.api.workloads     :refer [defoverload]]
             [wfl.jdbc              :as jdbc]
             [wfl.service.datarepo  :as datarepo]
@@ -12,7 +13,7 @@
             [wfl.module.all        :as all])
   (:import [clojure.lang ExceptionInfo]
            [java.sql Timestamp]
-           [java.time OffsetDateTime ZoneId]
+           [java.time Instant OffsetDateTime ZoneId]
            [java.time.format DateTimeFormatter]
            [java.time.temporal ChronoUnit]
            [wfl.util UserException]))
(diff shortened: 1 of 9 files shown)


api/test/wfl/integration/source_test.clj
```diff
@@ -37,18 +47,94 @@
...
+    (-> {:jobComplete true
+         :kind "bigquery#queryResponse"
+         :rows [@all-rows]
+         :totalRows (str (count @all-rows))}
+        (assoc-in [:schema :fields] (fields columns)))))
+
+(defn ^:private datarepo-query-table-between-miss
+  "Mock datarepo/query-table-between to miss some rows."
+  [dataset table between interval columns]
+  (let [miss (take-while (partial > mock-new-rows-size) fibonacci)
```
ehigham, line 88: Why fibonacci? Yes it's a nice implementation but not worth the conceptual burden. Can you take n instead?

api/src/wfl/source.clj
```diff
@@ -148,15 +149,18 @@
...
-      (datarepo/query-table-between table column interval [:datarepo_row_id])
-      :rows
-      flatten))
+  (let [tdr (-> dataset
+                (datarepo/query-table-between table column interval [:datarepo_row_id])
+                :rows flatten)]
+    (when (seq tdr)
+      (let [old (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+                  (mapcat :datarepo_row_ids (postgres/get-table tx details)))]
+        (remove (set old) tdr)))))
```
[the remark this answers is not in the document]
tbl3rd: > This makes me a little nervous doing this every time we update but I like the simplicity.
> What was your reasoning behind not doing this periodically?

We could query for all the rows in BQ periodically.
This implementation doubles the width of the query window to make the queries overlap,
and so runs on the same loop as the usual update.

api/src/wfl/source.clj
```diff
@@ -254,7 +258,10 @@
 (defn ^:private find-and-snapshot-new-rows
   "Create and enqueue snapshots from new rows in the `source` dataset."
   [{:keys [dataset table last_checked] :as source} utc-now]
-  (let [shards->jobs (->> [(timestamp-to-offsetdatetime last_checked) utc-now]
+  (let [checked      (timestamp-to-offsetdatetime last_checked)
+        hours-ago    (* 2 (max 1 (.between ChronoUnit/HOURS checked utc-now)))
```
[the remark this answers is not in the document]
tbl3rd: > Any particular reason for two hours? Also call it `two-hours-ago` maybe?

The intent is more like twice the requested interval at an hour's resolution:
so, up to an hour becomes 2 hours, 3 hours becomes 6, one day becomes 2, and so on.

api/test/wfl/integration/datarepo_test.clj
```diff
@@ -18,8 +18,9 @@
 
 (def ^:private testing-dataset {:id "4a5d30fe-1f99-42cd-998b-a979885dea00"
                                 :name "workflow_launcher_testing_dataset"})
-(def ^:private testing-snapshot {:id "0ef4bc30-b8a0-4782-b178-e6145b777404"
-                                 :name "workflow_launcher_testing_dataset7561609c9bb54ca6b34a12156dc947c1"})
+(def ^:private testing-snapshot
+  {:id "0ef4bc30-b8a0-4782-b178-e6145b777404"
+   :name "workflow_launcher_testing_dataset7561609c9bb54ca6b34a12156dc947c1"})
```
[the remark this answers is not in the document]
tbl3rd: I just added a newline so the def fit in my laptop window.
Nomadic development: I don't always have a wide screen.

api/test/wfl/integration/source_test.clj
```diff
@@ -37,18 +47,94 @@
...
+            (-> (Instant/now) str (subs 0 (count "2021-07-14T15:47:02"))))
+          (total-rows [query args]
+            (-> query (apply args) :totalRows Integer/parseInt))
+          (update-source [source mock-query-table-between]
+            (with-redefs [datarepo/query-table-between mock-query-table-between
+                          source/check-tdr-job         mock-check-tdr-job
+                          source/create-snapshots      mock-create-snapshots]
+              (source/update-source!
+               (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+                 (source/start-source! tx source)
```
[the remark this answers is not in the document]
tbl3rd: Thanks. Will fix.

api/test/wfl/integration/source_test.clj
tbl3rd: > Why fibonacci? Yes it's a nice implementation but not worth the conceptual burden. Can you take n instead?

Fibonacci felt like the sort of distribution we'd actually see:
several thick off the front with several stragglers following.
Didn't want to assume that missing rows were all at the front
of the interval.

api/src/wfl/source.clj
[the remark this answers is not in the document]
ehigham: Yes - that's exactly what makes me nervous.

api/src/wfl/source.clj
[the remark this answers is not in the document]
ehigham: sorry, misunderstood the code.

api/src/wfl/source.clj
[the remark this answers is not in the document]
ehigham: This implementation can check before the workload was :started which is a problem/

api/test/wfl/integration/source_test.clj
ehigham: Maybe the right thing to do is create a dataset specifically for this test. This smells like square pegs and round holes.

ehigham: Looking at the test, it's not immediately apparent what behaviour it's supposed to lock down.

docs/md/dev-process.md
```diff
@@ -225,6 +224,44 @@ silently fail or the Swagger page will fail to render. Check the
...
+With the UI running,
+open this URL in the browser.
+
+``` shell
+open http://localhost:8080/
+```
+
+Click the `LOGIN WITH GOOGLE` button if necessary,
+Then the `SWAGGER API` button at the upper right.
+You should see an interactive Swagger API page.
```
rexwangcc, line 262: Thanks for the docs, now I'm thinking that we should probably move the swagger back to api and remove the ui sometime…

rexwangcc (approved):
LGTM, thanks!

api/test/wfl/integration/source_test.clj
tbl3rd: Not sure what you mean.  Should the test just drop some rows from the start?
The test mocks the TDR and the dataset (mostly) specifically for this test.

docs/md/dev-process.md
tbl3rd: Sometime. ;-0

api/test/wfl/integration/source_test.clj
ehigham: I mean my main issue is that I don't understand what the fibonacci numbers have to do with missing a row when polling tdr.

I didn't remember if this test mocked TDR, thank you for clarifying that. 

It would be simpler if you returned one less that all the rows you're expecting the first time you poll, then allow all the rows to be found next time, with the assertion that the lower bound of the time interval is far back enough.

api/src/wfl/source.clj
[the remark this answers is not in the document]
tbl3rd: OK. Still not sure how to address what makes you nervous.
What hybrid do you suggest?
We could limit excluded rows from the database.
Also not sure what you mean by "not doing this periodically".

api/test/wfl/integration/source_test.clj
tbl3rd: The test is more comprehensive than that now.
It removes a set of rows scattered in the interval
from the results of the first query, then verifies
that only those rows show up in the cleanup round.
I can try to make that more obvious, but it sounds
like the test does not mock appropriate conditions,
or that the solution is wrong.  Maybe we should talk
after I address the bugs you found?

ehigham: > It removes a set of rows scattered in the interval

Why does it need to do that? what's wrong with removing all but the last?

tbl3rd: All but the last row?

api/src/wfl/source.clj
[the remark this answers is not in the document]
ehigham: What makes me nervous is doing this check every time we call update source.
What I meant by "periodically" was doing this once every so often at a longer period than the main update loop.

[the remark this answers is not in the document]
tbl3rd: > This implementation can check for new rows in TDR before the workload was :started which is a problem.

Don't we want to process rows in the source dataset added before the workload was started?

api/test/wfl/integration/source_test.clj
ehigham: Let's say you're mocking a BiqQuery query that's supposed to return a list 5 of rows that workflow-launcher will then snapshot. Instead of returning all 5 rows for the time interval specified, make the mock return all but the last row. Then in your test, assert that the next time wfl queries for the new rows (or when it next tries to find rows it had previously missed), the missed row is in the query interval.

tbl3rd: As you wish.

api/src/wfl/source.clj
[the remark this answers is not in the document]
ehigham: No.

api/test/wfl/integration/source_test.clj
ehigham: I'd prefer tests to be simple and not have additional logic that serves as a red herring.

api/test/wfl/integration/source_test.clj
```diff
@@ -37,18 +47,86 @@
...
+        (assoc-in [:schema :fields] (fields columns)))))
+
+(defn ^:private mock-query-table-between-miss
+  "Mock datarepo/query-table-between to miss some rows."
+  [dataset table between interval columns]
+  (-> (mock-query-table-between-all dataset table between interval columns)
+      (update-in [:rows 0] butlast)
+      (assoc :totalRows (str (dec mock-new-rows-size)))))
+
+;;234567890123456789012345678901234567890123456789012345678901234567890123456789
```
ehigham, line 88: What is this?

tbl3rd: Looks like two semi-colons followed by a string of digits. ;-0

api/src/wfl/source.clj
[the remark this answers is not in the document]
tbl3rd: I believe you mean re-starting a stopped workload, but OK.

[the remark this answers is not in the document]
ehigham: No, I mean creating a new one.

api/src/wfl/source.clj
```diff
@@ -448,3 +474,41 @@
...
+                 :snapshot_id "dac97a4d-ef77-46da-8956-8d2354d68bcd",
+                 :datarepo_row_ids
+                 ["f25f0764-5fcc-4376-b9dc-6e78f5d8e6cf"
+                  ,,,
+                  "6ce89f55-6d27-4146-9ff7-bd47226818b9"],
+                 :snapshot_creation_job_status "succeeded",
+                 :end_time #inst "2021-07-21T18:12:58.077575000-00:00",
+                 :id 3,
+                 :consumed nil,
+                 :snapshot_creation_job_id "mock_job_id_2"}]))
```
ehigham, line 514: please remove

api/src/wfl/source.clj
```diff
@@ -147,16 +157,37 @@
                                   column dataset)
       (assoc source :dataset dataset))))
 
+(defn combine-tdr-source-details
+  "Reduce to `result` by combining the Terra Data Repo source `_detail`."
```
ehigham, line 161: This docstring doesn't mean anything to me  - I'm none-the-wiser about what this function does

api/src/wfl/source.clj
```diff
@@ -147,16 +157,37 @@
...
+                (-> start_time
+                    (util/latest (instant/read-instant-timestamp begin))
+                    timestamp-to-offsetdatetime
+                    (.format bigquery-datetime-format))
+                begin)
+        tdr   (-> dataset
+                  (datarepo/query-table-between
+                   table column [start end] [:datarepo_row_id])
+                  :rows flatten)]
+    (when (seq tdr) (remove (set (:datarepo_row_ids old)) tdr))))
```
ehigham, line 190: I'm still not convinced this should be done on every call to `update-source`. I also think the way you're getting the `start_time` is overly complicated - The timestamp when the source was "started" could be written into the database as a new `started` column, rather than setting `last_checked`. The code to compute the start time could then be done once in the loader for any sources that have a `last_checked` but don't have a `started` (i.e. workloads created prior to this change).

ehigham (approved):

api/test/wfl/integration/source_test.clj
ehigham: I'm glad your having fun

api/src/wfl/source.clj
tbl3rd: OK, but WFL still has to read the details to get the row_ids already processed.
Time-fencing the processing interval is trivial aside from the differing formats.
The times are there now. Might as well use them? Why add another timestamp?

api/src/wfl/source.clj
tbl3rd: Sorry.

api/src/wfl/source.clj
tbl3rd: Howzabout?

"Reduce to `result` by combining the Terra Data Repo source `_detail`.
  Collect `datarepo_row_ids` already seen while preserving the latest
 `end_time` and the earliest `start_time`."

api/test/wfl/integration/source_test.clj
tbl3rd: Sad that you are not?

api/test/wfl/integration/source_test.clj
tbl3rd: Not a fan of herring myself.

api/src/wfl/source.clj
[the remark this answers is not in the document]
tbl3rd: Can we point WFL at a dataset to process rows already there,
or can it process only rows added after a workload is started?

tbl3rd:
Thank you for the 👍 .

PR #469, opened 2021-07-01
Remove trim-slashes and fix endpoints in wfl.tools namespace

rfricke-asymmetrik:
### Purpose
Updates the wfl-url function in the endpoints namespace to not use trim-slashes and removes the trim-slashes function from util.

### Changes
- Remove trim-slashes function and unit test
- Remove use of trim-slashes from wfl-url
- Update endpoints in wfl.tools.endpoints by removing slashes from beginning and end of the url parts

### Review Instructions

- No instructions.

### DIFF
--- a/api/src/wfl/util.clj
+++ b/api/src/wfl/util.clj
@@ -384,15 +384,6 @@
     (recur (subs url 0 (dec (count url))))
     url))
 
-(defn trim-slashes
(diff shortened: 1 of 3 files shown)


tbl3rd (approved):
Thanks!

ehigham (approved):

PR #470, opened 2021-07-06
[GH-1139] Added new log namespace

rfricke-asymmetrik:
### Purpose
We want to replace the current logging with json output so that it can be picked up by stackdriver and provide easier querying.

https://broadinstitute.atlassian.net/browse/GH-1139

### Changes
- Added new log namespace with macros for info, warn, debug and error.

### Review Instructions

- Build WFL and start the server locally
- Watch the stream of logs in the terminal and make sure they are all in json format with a severity, message and timestamp at the least
- Hit some of the endpoints via swagger and make sure that there are log entries coming through for those as well
- Try to produce an error to make sure that exceptions are logged as ERROR severity log entries


api/src/wfl/log.clj
```diff
@@ -0,0 +1,72 @@
...
+  (:require [clojure.core]
+            [clojure.data.json :refer [write-str]])
+  (:import [java.sql Timestamp]
+           [java.time OffsetDateTime]))
+
+(defn googleize-field
+  "Prepends the google logging domain to the fields that require it for Google Logging."
+  [field]
+  (let [field-name (name field)]
+    (if (some #(= field-name %) '("labels" "insertId" "operation" "sourceLocation" "spanId" "trace")) (str "logging.googleapis.com/" (name field-name))
```
ehigham, line 12: You can test if a string is member of that set: 
```clojure
(if (#{"labels" "insertId" "operation" "sourceLocation" "spanId" "trace"} field-name)
   ...)
```

api/src/wfl/log.clj
```diff
@@ -0,0 +1,72 @@
...
+        field-name)))
+
+(defn write-to-stdout
```
ehigham, line 15: Any of these functions you don't expect clients to use, can you annotate them with `^:private`?

api/src/wfl/log.clj
```diff
@@ -0,0 +1,72 @@
...
+      (throw (ex-info "Invalid severity provided to logging function." {:severity severity-upper}))
+      (->> (add-common-fields {:severity severity-upper} message)
+           (merge additional-fields)
+           (write-to-stdout)))))
+
+(defmacro info
+  "Log as Information"
+  [message & more]
+  `(->> (print-str ~message ~@more)
+        (log "INFO")))
```
ehigham, line 48: Not sure the threading macro is necessary here, simply 
```clojure
(log :info (print-str ~message ~@more))
```

api/src/wfl/log.clj
```diff
@@ -0,0 +1,72 @@
...
+  (merge json
+         {:message message
+          :timestamp (get-timestamp)}))
+
+(defn log
+  "Write a log entry to standard output."
+  [severity message & {:as additional-fields}]
+  (let [severity-upper (.toUpperCase severity)]
+    (if (not (some #(= severity-upper %) '("DEFAULT" "DEBUG" "INFO" "NOTICE" "WARNING" "ERROR" "CRITICAL" "ALERT" "EMERGENCY")))
+      (throw (ex-info "Invalid severity provided to logging function." {:severity severity-upper}))
```
ehigham, line 39: Not sure this is really necessary - I'd let clients pass in the log level and just convert to the required format - you can use `to-upper` instead of the java method to make it seem more idiomatic, for example:
```clojure 
(-> severity name to-upper)
```

api/src/wfl/log.clj
```diff
@@ -0,0 +1,72 @@
...
+  [message & more]
+  `(log :error (print-str ~message ~@more)))
+
+(defmacro errorf
+  "Log as Error with formatting"
+  [message & more]
+  `(log :error (format ~message ~@more)))
+
+(defmacro trace
+  "Log as Trace. Because there isn't a TRACE Severity in GCP Logging, this will just add an entry as DEBUG"
```
ehigham, line 70: Then do we need it?

api/src/wfl/log.clj
```diff
@@ -0,0 +1,67 @@
...
+            [clojure.string :refer [upper-case]])
+  (:import [java.sql Timestamp]
+           [java.time OffsetDateTime]))
+
+(defn ^:private googleize-field
+  "Prepends the google logging domain to the fields that require it for Google Logging."
+  [field]
+  (let [field-name (name field)]
+    (if (#{"labels" "insertId" "operation" "sourceLocation" "spanId" "trace"} field-name)
+      (str "logging.googleapis.com/" (name field-name))
```
ehigham, line 14: Repeat call to `name` ?

api/src/wfl/log.clj
```diff
@@ -0,0 +1,67 @@
...
+      field-name)))
+
+(defn log
+  "Write a log entry to standard output."
+  [severity message & {:as additional-fields}]
+  (let [severity-upper (-> severity name upper-case)]
```
ehigham, line 20: You can inline this - don't think it'll hurt readability

api/src/wfl/api/routes.clj
```diff
@@ -138,7 +137,7 @@
   [status message exception {:keys [uri] :as request}]
   (let [response (exception-handler status message exception request)]
     (log/errorf "Server %s error at occurred at %s :" (:status response) uri)
-    (logr/error exception (:body response))
+    (log/error exception (:body response))
```
okotsopoulos, line 140: I imagine there was a reason for using the readable logger initially:

https://github.com/clojure/tools.logging/blob/master/src/main/clojure/clojure/tools/logging/readable.clj

"Logging macros that support printing message (some) arguments as
  if wrapped in pr-str, the goal being to preserve their data representation
  distinct from the explanatory text."
  
  Does the new logger take care of this or should the argument be wrapped in `pr-str`?

api/src/wfl/log.clj
```diff
@@ -0,0 +1,84 @@
+(ns wfl.log
+  "Logging for WFL"
```
okotsopoulos, line 2: Would be nice to specify that the logging is backed by Google Stackdriver (unless we're supporting different kinds of loggers?) along with a link to external docs.

rfricke-asymmetrik: You are right. I do need to update with more documentation for this.

api/src/wfl/api/routes.clj
rfricke-asymmetrik: I suppose in this case it does not matter since there is a call to log/errorf right above it that will create the json log, but pr-str would print to output as a string instead of json which likely would cause stackdriver to not pick it up.

okotsopoulos:
From a local development perspective, these logs are harder to read through in the terminal.  A few related questions:

- Log4j allowed us to set a root log level, e.g. setting this to `DEBUG` would only yield `DEBUG, INFO, WARN, …` logs. Can we do something similar here?
- Is there a way to easily query logs locally, via Stackdriver or some other way?

Ultimately I'm excited for what this will mean for logging in production though!

api/src/wfl/log.clj
```diff
@@ -0,0 +1,84 @@
...
+(defn ^:private googleize-field
+  "Prepends the google logging domain to the fields that require it for Google Logging."
+  [field]
+  (let [field-name (name field)]
+    (if (#{"labels" "insertId" "operation" "sourceLocation" "spanId" "trace"} field-name)
+      (str "logging.googleapis.com/" field-name)
+      field-name)))
+
+(defprotocol Logger
+  ""
```
okotsopoulos, line 18: Oversight?

api/src/wfl/api/routes.clj
okotsopoulos: I guess it would be helpful to see concrete examples of these error logs and hear from whoever added them and/or has consumed them in the past.

api/src/wfl/log.clj
```diff
@@ -0,0 +1,84 @@
...
+(defn log
+  "Write a log entry to standard output."
+  [severity message & {:as additional-fields}]
+  (write *logger* (merge {:severity (-> severity name upper-case)
+                          :message message
+                          :timestamp (Timestamp/from (.toInstant (OffsetDateTime/now)))}
+                         additional-fields)))
+
+(defmacro info
+  "Log as Information"
```
okotsopoulos, line 47: I'd like the docstrings for these macros to include the enum description for the associated log level from [Google's documentation](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity), if that's appropriate.  The enum descriptions did a good job of communicating to me when to use which log level.

api/src/wfl/log.clj
```diff
@@ -0,0 +1,84 @@
...
+(defmacro warn
+  "Log as a Warning"
+  [message & more]
+  `(log :warning (print-str ~message ~@more)))
+
+(defmacro warnf
+  "Log as a Warning with formatting"
+  [message & more]
+  `(log :warning (format ~message ~@more)))
+
```
okotsopoulos, line 65: I'd personally use the `NOTICE` log level as well :)

rfricke-asymmetrik: I am going to add macros for all of the google supported log levels. :)

api/src/wfl/log.clj
rfricke-asymmetrik: Thats a good idea. I will do that.

api/src/wfl/sink.clj
```diff
@@ -148,17 +148,17 @@
 
 (defn ^:private update-terra-workspace-sink
   [executor {:keys [fromOutputs workspace entityType identifier details] :as _sink}]
-  (when-let [[_ {:keys [uuid outputs] :as workflow}] (stage/peek-queue executor)]
```
ehigham, line 151: restore this.

api/src/wfl/source.clj
```diff
@@ -429,7 +428,11 @@
   (zero? (tdr-snapshot-list-queue-length source)))
 
 (defn ^:private tdr-snapshot-list-to-edn [source]
-  (let [read-snapshot-id (comp :id edn/read-string :item)]
+  (letfn [(read-snapshot-id [ss]
+            (-> ss
+                :item
+                edn/read-string
+                :id))]
```
ehigham, line 435: revert?

api/src/wfl/jdbc.clj
```diff
@@ -1,18 +1,21 @@
...
-                   identifier (str (:user db) "@" (:connection-uri db) "#" random)]
-               (log/infof "JDBC new DB config %s observed (the #%s isn't important, just unique):"
-                          identifier random)
+  (memoize (fn [{:keys [connection-uri user] :as db}]
+             (let [random     (rand-int 10000)
+                   identifier (str user "@" connection-uri "#" random)]
+               (log/info
+                (format (str/join \space
+                                  ["JDBC new DB config %s observed"
+                                   "(the #%s isn't important, just unique):"])
```
ehigham, line 17: delete this line - it's really not interesting!

api/src/wfl/log.clj
```diff
@@ -0,0 +1,67 @@
...
+          (json/write-str :key-fn key-fn :escape-slash false)
+          println))))
+
+(def ^:dynamic *logger*
+  "The logger now."
+  stdout-logger)
+
+(defmacro log
+  "Log `expression` with `severity`."
+  [severity expression]
```
ehigham, line 38: If your log function took an `expression & tail`, you can consolidate all the string concatenation here by
```clojure
(str/join " " (cons ~expression ~tail))
```

api/src/wfl/sink.clj
```diff
@@ -148,17 +148,17 @@
...
       (when (entity-exists? workspace entity)
-        (log/debug "entity" name "already exists - deleting previous entity.")
+        (log/debug (str "entity" name "exists - deleting previous entity."))
         (firecloud/delete-entities workspace [entity]))
-      (log/debug "upserting workflow" uuid "outputs as" name)
+      (log/debug (str "upserting workflow" uuid "outputs as" name))
       (rawls/batch-upsert workspace [(conj entity attributes)])
       (stage/pop-queue! executor)
-      (log/info "sunk workflow" uuid "to" workspace "as" name)
+      (log/info (str "sunk workflow" uuid "to" workspace "as" name))
```
ehigham, line 161: previously these were separated with a space. now they're not.

api/test/wfl/unit/spec_test.clj
```diff
@@ -47,11 +47,8 @@
...
-            [[valid?   "hornet-eng@broadinstitute.org"]
-             ;; From tbl: https://www.netmeister.org/blog/email.html
-             [valid?   "'*+-/=?^_`{|}~#$@[IPv6:2001:470:30:84:e276:63ff:fe72:3900]"]
-             [invalid? "foo"]]]
-      (is (test? ::all/watchers [email]) (format "failed: %s" email)))))
+  (let [valid?   (partial s/valid? ::all/watchers)
+        invalid? (complement valid?)]
+    (is (valid?   ["hornet-eng@broadinstitute.org"]))
+    (is (valid?   ["'*+-/=?^_`{|}~#$@[IPv6:2001:470:30:84:e276:63ff:fe72:3900]"]))
+    (is (invalid? ["foo"]))))
```
ehigham, line 54: This previously printed the email that failed in the log - that was helpful for immediately recognising what was wrong in PRs.

docs/md/dev-logging.md
```diff
@@ -1,67 +1,58 @@
...
     ...
-    [clojure.tools.logging :as log]
+    [wfl.log :as log]
     ...))
 
 (log/info "Hello!")
 ```
 
-Full documentation is available [here](http://clojure.github.io/tools.logging/#clojure.tools.logging).
+There are currently 4 macros for creating simple json logs for `INFO, WARN, DEBUG, ERROR`. There is also
```
okotsopoulos, line 29: I believe you also have `NOTICE` too?

rfricke-asymmetrik: Good catch, fixing now.

okotsopoulos (commented):
Closing the loop from our Slack convo publicly here:

While I know that we want to avoid scope creep on this PR, I'm withholding my approval because the impact on local logging could set back development on other initiatives.  I recognize that opinions may differ here, if this gets approved by someone else then that's fine.

My concerns:
- No ability to throttle the fire hose of all logs emitted
- Human readability suffers without a companion UI to search and browse

@rfricke-asymmetrik and I chatted about these possible improvements:
- Adding an Atom in the namespace that we could set via an endpoint that only hornet team members have access to
- Have a way to send the logs to a file (this option actually opens up some secondary tools like a simple cli or ui can be used to say "I want only logs from the file with this label" or certain operations)

I'd be happy to approve with only the ability to throttle implemented.  That could even be a piece of well-documented "grep magic" to serve as a stopgap.

rfricke-asymmetrik:
One note on the atom/endpoint approach to changing the logging level: 

- The app is deployed via kubernetes which would have autoscaling. This means that a call to the endpoint would only really be setting the logging level for one pod. For this approach to work for the entire api, the logging configuration would need to be persisted to the db with some caching/memoization for the change to propagate to all of the running pods.

ehigham (approved):
A great improvement. Thank you very much.

rexwangcc (commented):
LGTM, sorry for the delayed review! (not sure if it was intentional to check in `error.txt` ?)

rfricke-asymmetrik:
Did not even realize that was there. I made a new PR to remove it.

PR #471, opened 2021-07-07
[GH-1411] Describe Workflow Outputs

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1411
Inform the `Sink` downstream of the executor of the workflow type.

When sinking outputs to the TDR, you need to ingest the file outputs followed by the metadata with the files replaced by their fileref URIs. Thus, we need a way of determining if a particular output is a `File` that needs ingesting or just a metadata `String` that happens to be a gs url. I've had some luck in the past using womtool to describe the WDL and then walk the description of the outputs, dispatching to the relevant handlers (see `workflows.clj` in the test tools). For lack of a better alternative, I'm going to pursue using womtool to handle outputs in a generic way.

Two design approaches were immediately obvious:
1. When configuring the workload, pass the whole workload configuration to each of the source, sink and executor and make the sink do extra work if it needs to.
2. Modify peek-queue to peek each object as a [type value] pair out of the queue.

I don't like 1 because it means now every source/sink/executor impl has to know about each other. 2 seems like a better design but impacts a number of things.

Implemented by changing the `peek-queue` multimethod to return a `'(description value)` pair  where `description` is the workflow description as reported by womtool and the `value` is the workflow.
I'm not caching the description for the TerraExecutor - while it *should* be constant for the whole workload, a user might specify a branch rather than a specific commit... I'm really not sure what to do here.

Since I've changed the `Queue` interface, I've propagated this change to all affected areas including the TDR source implementations - I think this makes the intent of `from-source` in the executor far more explicit.


api/src/wfl/executor.clj
```diff
@@ -404,6 +423,7 @@
 (defn ^:private terra-executor-to-edn [executor]
   (-> executor
       (util/select-non-nil-keys (keys terra-executor-serialized-fields))
+      (dissoc :description)
```
okotsopoulos, line 426: Can you explain why we pull this out here?

api/src/wfl/service/dockstore.clj
```diff
@@ -0,0 +1,15 @@
+(ns wfl.service.dockstore
+  "Analyze data in Terra using the Firecloud/Terra API."
```
okotsopoulos, line 2: Copypasta?

api/src/wfl/service/firecloud.clj
```diff
@@ -238,11 +238,19 @@
   [workspace]
   (get-workspace-json workspace "methodconfigs?allRepos=true"))
 
-(defn get-method-configuration
+(defn method-configuration
```
okotsopoulos, line 241: Nice.

api/src/wfl/source.clj
```diff
@@ -136,24 +136,13 @@
...
-                       e)))))
-
 (defn verify-data-repo-source!
   "Verify that the `dataset` exists and that WFL has the necessary permissions
    to read it."
   [{:keys [dataset table column skipValidation] :as source}]
   (if skipValidation
     source
-    (let [dataset (get-dataset-or-throw dataset)]
+    (let [dataset (datarepo/dataset dataset)]
```
okotsopoulos, line 145: Thanks.

api/src/wfl/executor.clj
ehigham: The user doesn't need to see it

api/src/wfl/service/dockstore.clj
ehigham: Ah yep

api/src/wfl/executor.clj
```diff
@@ -137,7 +144,7 @@
            methodConfiguration
            methodConfigurationVersion] :as _executor}
    {:keys [name]                       :as _reference}]
-  (let [current (method-config-or-throw workspace methodConfiguration)
+  (let [current (firecloud/method-configuration workspace methodConfiguration)
```
tbl3rd, line 147: Suggestion:
```clojure
  "Update `methodConfiguration` in `workspace` with snapshot reference `name`
  Update executor table record ID with incremented `methodConfigurationVersion`."
  [{:keys [id workspace methodConfiguration methodConfigurationVersion] :as _executor}
   {:keys [name]                                                        :as _reference}]
  (let [current (firecloud/method-configuration workspace methodConfiguration)]
    (throw-on-method-config-version-mismatch current methodConfigurationVersion)
    (let [inc'd (inc methodConfigurationVersion)]
      (-> current
          (assoc :dataReferenceName name :methodConfigVersion inc'd)
          (->> (firecloud/update-method-configuration workspace methodConfiguration)))
        (jdbc/update! tx terra-executor-table
                      {:method_configuration_version inc'd}
                      ["id = ?" id])))))
```

api/src/wfl/executor.clj
```diff
@@ -285,15 +292,37 @@
...
+  "Return the URL of the WDL contents associated with this `method`."
+  [{:keys [methodPath methodVersion] :as method}]
+  (throw-unless-dockstore-method method)
+  (let [id  (url-encode (str "#workflow/" methodPath))
+        url (:url (dockstore/ga4gh-tool-descriptor id methodVersion "WDL"))]
+    (if-let [groups (re-find #"^https://raw.githubusercontent.com/(.*)$" url)]
+      ;; use jsdelivr to avoid GitHub rate-limits
+      (let [[author repo version path] (str/split (second groups) #"/" 4)]
+        (str/join "/" ["https://cdn.jsdelivr.net/gh" author (str repo "@" version) path]))
+      url)))
```
tbl3rd, line 305: Move `;;` comment on top of `defn`?

api/src/wfl/service/datarepo.clj
```diff
@@ -28,7 +30,13 @@
...
   [dataset-id]
   {:pre [(some? dataset-id)]}
-  (get-repository-json "datasets" dataset-id))
+  (try
+    (get-repository-json "datasets" dataset-id)
+    (catch ExceptionInfo e
+      (throw
+       (UserException. "Cannot access dataset"
+                       {:dataset dataset-id :status (-> e ex-data :status)}
+                       e)))))
```
tbl3rd, line 39: Moved from source.clj?

api/src/wfl/service/dockstore.clj
```diff
@@ -0,0 +1,15 @@
+(ns wfl.service.dockstore
+  "Wrappers for Dockstore HTTP APIs."
+  (:require [clj-http.client   :as http]
+            [clojure.string    :as str]
+            [wfl.environment   :as env]
+            [wfl.util          :as util]))
+
+(defn ^:private dockstore-url [& parts]
+  (let [url (util/de-slashify (env/getenv "WFL_DOCKSTORE_URL"))]
+    (str/join "/" (cons url parts))))
```
tbl3rd, line 10: I don't get the `de-slashify`, but OK.

api/src/wfl/service/dockstore.clj
```diff
@@ -0,0 +1,15 @@
...
+
+(defn ga4gh-tool-descriptor [id version type]
+  (-> (dockstore-url "api/api/ga4gh/v2/tools" id "versions" version type "descriptor")
+      http/get
+      (util/response-body-json)))
```
tbl3rd, line 15: Extra parens but I don't care.

api/src/wfl/service/firecloud.clj
```diff
@@ -238,11 +238,19 @@
...
-  (get-workspace-json workspace "method_configs" methodconfig))
+  (try
+    (get-workspace-json workspace "method_configs" methodconfig)
+    (catch ExceptionInfo cause
+      (throw (UserException.
+              "Cannot access method configuration in workspace"
+              {:workspace           workspace
+               :methodConfiguration methodconfig
+               :status              (-> cause ex-data :status)}
+              cause)))))
```
tbl3rd, line 253: Thank you.

api/src/wfl/sink.clj
```diff
@@ -148,7 +148,7 @@
 
 (defn ^:private update-terra-workspace-sink
   [executor {:keys [fromOutputs workspace entityType identifier details] :as _sink}]
-  (when-let [{:keys [uuid outputs] :as workflow} (stage/peek-queue executor)]
+  (when-let [[_ {:keys [uuid outputs] :as workflow}] (stage/peek-queue executor)]
```
tbl3rd, line 151: Suggest `_type` (or is it `_name`) instead of just `_`?

api/src/wfl/source.clj
```diff
@@ -136,24 +136,13 @@
...
-(defn ^:private get-dataset-or-throw
-  "Return the dataset with `dataset-id` or throw"
-  [dataset-id]
-  (try
-    (datarepo/dataset dataset-id)
-    (catch ExceptionInfo e
-      (throw
-       (UserException. "Cannot access dataset"
-                       {:dataset dataset-id :status (-> e ex-data :status)}
-                       e)))))
```
tbl3rd, line 148: Moved (and renamed) to source.clj?

api/src/wfl/source.clj
tbl3rd, line 145: Ah.  OK.

api/src/wfl/source.clj
```diff
@@ -311,7 +300,7 @@
   "Get first unconsumed snapshot from `source` queue."
   [source]
   (when-let [{:keys [snapshot_id] :as _record} (peek-tdr-source-details source)]
-    (datarepo/snapshot snapshot_id)))
+    ["snapshot" (datarepo/snapshot snapshot_id)]))
```
tbl3rd, line 303: `"snapshot"` is a `:type`, right?

api/src/wfl/source.clj
```diff
@@ -417,32 +410,32 @@
...
 (defoverload start-source!  tdr-snapshot-list-type  start-tdr-snapshot-list)
 (defoverload stop-source!   tdr-snapshot-list-type  stop-tdr-snapshot-list)
 (defoverload update-source! tdr-snapshot-list-type  update-tdr-snapshot-list)
+
+(defoverload create-source! tdr-snapshot-list-name  create-tdr-snapshot-list)
 (defoverload load-source!   tdr-snapshot-list-type  load-tdr-snapshot-list)
 
-(defoverload stage/peek-queue
-  tdr-snapshot-list-type (comp edn/read-string :item peek-tdr-snapshot-list))
+(defoverload stage/peek-queue tdr-snapshot-list-type peek-tdr-snapshot-list)
```
tbl3rd, line 434: Thank you.

api/test/wfl/integration/executor_test.clj
```diff
@@ -225,20 +225,21 @@
...
        #'firecloud/get-submission              mock-firecloud-get-submission
        #'firecloud/get-workflow                mock-workflow-keep-status}
       #(executor/update-executor! source executor))
     (with-redefs-fn
-      {#'firecloud/get-workflow         (constantly succeeded-workflow-mock)
+      {#'executor/describe-method       (constantly nil)
+       #'firecloud/get-workflow         (constantly succeeded-workflow-mock)
        #'firecloud/get-workflow-outputs mock-firecloud-get-workflow-outputs}
-      #(let [workflow (stage/peek-queue executor)]
+      #(let [[_ workflow] (stage/peek-queue executor)]
```
tbl3rd, line 242: Don't want to validate the type?

api/test/wfl/integration/sink_test.clj
```diff
@@ -109,7 +109,7 @@
                      :outputs (-> "sarscov2_illumina_full/outputs.edn"
                                   resources/read-resource
                                   (assoc :flowcell_id testing-entity-name))}
-        executor    (make-queue-from-list [workflow])
+        executor    (make-queue-from-list [[nil workflow]])
```
tbl3rd, line 112: Why `nil`s?

tbl3rd (approved):
OK, but I have questions …

api/src/wfl/environment.clj
```diff
@@ -30,31 +30,33 @@
...
    ;; -- variables used in test code below this line --
    "WFL_CROMWELL_URL"
-   #(-> "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org")
+   (fn [] "https://cromwell-gotc-auth.gotc-dev.broadinstitute.org")
    "WFL_TDR_DEFAULT_PROFILE"
-   #(-> "6370f5a1-d777-4991-8200-ceab83521d43")
+   (fn [] "6370f5a1-d777-4991-8200-ceab83521d43")
    "WFL_WFL_URL"
-   #(-> "https://dev-wfl.gotc-dev.broadinstitute.org")})
+   (fn [] "https://dev-wfl.gotc-dev.broadinstitute.org")})
```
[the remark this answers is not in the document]
ehigham: the linter complained. Yes their all `constantly`, but clojure makes that more characters to type than `fn [] `.

api/src/wfl/executor.clj
ehigham: Not relevant to this change.

api/src/wfl/executor.clj
ehigham: why? It describes why I'm doing this, it's not relevant to the workings of the whole function

api/src/wfl/service/datarepo.clj
ehigham: yes

api/src/wfl/service/dockstore.clj
ehigham: If you specify the URL ending with a `/` in the environment variable.

api/src/wfl/service/dockstore.clj
ehigham: clearly you do lol

api/src/wfl/sink.clj
ehigham: again, I don't agree that this adds clarity.

api/src/wfl/source.clj
ehigham: from source.clj - yes.

api/src/wfl/source.clj
ehigham: maybe? Not sure what you mean by a `:type`.

api/test/wfl/integration/executor_test.clj
ehigham: it's `(constantly nil)` here.

api/src/wfl/executor.clj
```diff
@@ -114,19 +113,27 @@
...
-  (cond (= "importSnapshot" fromSource)
-        (rawls/create-or-get-snapshot-reference workspace id name)
-        :else
-        (throw (ex-info "Unknown fromSource" {:executor executor}))))
+  "Coerce `object` to form understood by `executor``."
+  [executor [description item :as object]]
+  (cond (= "snapshot" description) (entity-from-snapshot executor item)
+        :else                      (throw (ex-info "No method to coerce object into workspace entity"
+                                                   {:executor executor
+                                                    :object   object}))))
```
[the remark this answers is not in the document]
okotsopoulos: I'll take responsibility here, I added this method initially.  I liked the `cond` because I was pretty sure we'd be supporting other methods of coercion in the future.

api/test/wfl/integration/sink_test.clj
ehigham: because I'm not testing it here.

api/src/wfl/sink.clj
okotsopoulos: I'm with Tom here, I find meaningful names to add clarity even if they're not currently being used.

api/src/wfl/executor.clj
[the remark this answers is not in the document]
ehigham: I'm trying to program intent. So far we only support snapshots. If we did https://docs.google.com/presentation/d/159ZnFwWh2tCTv4e9n3IxnGkC5VNfMorO4FqKJpjXuyY/edit#slide=id.ge1e3db2ca4_0_0, we'd have a terra workspace source that would supply `entity`s and so we wouldn't require any "coercion".

api/src/wfl/sink.clj
ehigham: but it doesn't add any clarity because its not used - the whole point of `_` is you don't care what it is.

api/src/wfl/environment.clj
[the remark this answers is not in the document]
tbl3rd: I think `constantly` conveys intent better than golf, but OK.

api/src/wfl/source.clj
tbl3rd: I mean a method dispatch value, but not sure which is why I ask.

api/src/wfl/service/dockstore.clj
tbl3rd: Obviously not a big deal, but just don't do that?
Who else is specifying our environment variables?

api/src/wfl/environment.clj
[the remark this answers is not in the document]
ehigham: `fn []` is golf?

api/src/wfl/source.clj
ehigham: Ah no this isn't for type dispatching. I've changed the type of the objects stored in the source and executor queues to be the pair `[description value]`. I suppose someone could implement a consumer of this to dispatch off the description if they so wished.

This was really what I was trying to get reviewed - how should information be passed down the processing stages? I've decided to use tuples but I could have `assoc`'ed a description into the object being peeked.

api/src/wfl/executor.clj
ehigham: moved into docstring

api/src/wfl/service/dockstore.clj
ehigham: It's an easy mistake to make I guess in testing/overriding for system tests. At least this way it doesn't matter if someone specifies a url with a trailing `/`.

api/src/wfl/source.clj
tbl3rd: A pair is probably OK, but it is not clear to me what it means.
(I noticed in other places that `"snapshot"` was a `description`.)
I've used namespaced keywords for stuff like this.
Stages in the data pipeline associate a private key to maps they process.
Other stages act on that key as necessary.
Namespacing prevents clashes among users of the data.
(This is part of what I was trying to explain when discussing protocol/record trade-offs.)

tbl3rd: Consider `::Headers` and `::Properties` in PTC's JMS code, for example.

api/src/wfl/environment.clj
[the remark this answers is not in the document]
tbl3rd: The "more characters to type than" argument is golf.
Brevity is nice, but to me `constantly` means I have
a constant, but this context requires a callable thing.
(As always, YMMV.)

api/src/wfl/source.clj
ehigham: So I haven't completely made my mind up about what the thing should be. I'm not sure how to convey this without a type system so here goes:

For source->executor, my thinking was that it was kind of like a data constructor for sum types, ie
```haskell
data ThingThatCanBeImportedIntoAWorkspace = Snapshot s
                                                                             |  Entity e
```
This works well because we can define up front all the things that we can import into a workspace. In clojure, I can implement this as the pair `[name value]` then do a case analysis on the name (you're right - it should be a case analysis in `executor.cjl`).

For the executor->sink interface this doesn't work because the types of the outputs are dynamic. I need to traverse the outputs though in order to ingest files into TDR so i need a representation of this somehow. Right now, I'm just using the raw womtool WDL description so the executor becomes a queue of `[Description Workflow]` pairs. I didn't think making `Description` a member of `Workflow` because it feels more like metadata.

api/src/wfl/executor.clj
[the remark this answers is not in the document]
ehigham: Used a case analysis.

api/src/wfl/source.clj
ehigham: Also used qualified keywords as you suggested

api/src/wfl/executor.clj
```diff
@@ -285,15 +292,37 @@
...
+(defn ^:private method-content-url
+  "Return the URL to the WDL content used by this `methodRepoMethod`. If the WDL
+   is hosted on GitHub, returns a jsdelivr URL to avoid GitHub rate limits."
+  [{:keys [methodPath methodVersion] :as methodRepoMethod}]
+  (throw-unless-dockstore-method methodRepoMethod)
+  (let [id  (url-encode (str "#workflow/" methodPath))
+        url (:url (dockstore/ga4gh-tool-descriptor id methodVersion "WDL"))]
+    (if-let [groups (re-find #"^https://raw.githubusercontent.com/(.*)$" url)]
+      (let [[author repo version path] (str/split (second groups) #"/" 4)]
+        (str/join "/" ["https://cdn.jsdelivr.net/gh" author (str repo "@" version) path]))
```
tbl3rd, line 304: This provokes a `Disallowed domain in URI` response from Firecloud now.

PR #472, opened 2021-07-08
Add Retry functionality to the Covid Executor

rhiananthony:
### Purpose
The purpose of this ticket is to add retry functionality for workflows. The plan is to:
     1. Get the workflows of status X and null retry field
     2. Get the distinct list of snapshots for these workflows
     3. Submit the snapshots. Hold onto the new workflow ids
     4. Update the retry field of the executor details table inserting the new workflow id for the old workflow.

### Ticket
This is related to ticket GH-1394

### Changes
Additional method added plus tests. When complete, this will be generalized and moved to the generic executor.

### Review Instructions
Please let me know if there are better ways to do this. Also, if any part of what I am doing is unclear


api/src/wfl/module/covid.clj
```diff
@@ -156,20 +158,131 @@
...
+
+  ; 4. Update the retry field of the executor details table inserting the new workflow id for the old workflow.
+
+)
+
+(comment
+  (let [tx (jdbc/with-db-transaction [tx (postgres/wfl-db-config)])
+        w (workloads/load-workload-for-uuid tx "b8b5bbe6-faad-4d03-ad45-a68f6c860c7b")]
+    (pprint w)
+    (pprint (util/to-edn w)))
```
ehigham, line 188: You cannot return `tx` from this scope - it's lifetime is bound to the body in which it's created
```clojure
  (let [w (workloads/load-workload-for-uuid tx "b8b5bbe6-faad-4d03-ad45-a68f6c860c7b")]
    (pprint w)
    (pprint (util/to-edn w))))
```

rhiananthony: I tried this:
`(jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
    (let [w (workloads/load-workload-for-uuid tx "b8b5bbe6-faad-4d03-ad45-a68f6c860c7b")]
      (pprint w)
      (pprint (util/to-edn w))))`

Same result - 
`04:14:54 INFO  wfl.module.covid - JDBC transaction 2495 started to "ranthony@jdbc:postgresql:wfl#9806"
Execution error (NoSuchMethodError) at clojure.java.jdbc/db-transaction* (jdbc.clj:838).
clojure.java.jdbc$db_transaction_STAR_$fn__8398.<init>(ZLjava/lang/Object;)V`

api/src/wfl/module/covid.clj
```diff
@@ -94,82 +96,122 @@
...
+      ; resubmit the reference values
+      ; write back to the table, putting the new workflow id into the retry field of the failed one
+      (let [workflow_ids (util/to-quoted-comma-separated-list (map :uuid workflows))
+            executor (get-in workload [:executor])
+            executor_details (get-in workload [:executor :details])
+            query (format "SELECT * FROM %s WHERE workflow IN (%s)" executor_details workflow_ids)
+            results (jdbc/query (postgres/wfl-db-config) [query])
+            distinct_references (set (map :reference results))]
+           (for [ref distinct_references]
+                (let [submission (executor/create-submission! executor ref)]
```
okotsopoulos, line 185: Take a look at the source code for [`executor/create-submission!`](https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/executor.clj#L159-L164):
```
(defn ^:private create-submission!
  "Update `methodConfiguration` to use `reference`.
  Create and return submission in `workspace` for `methodConfiguration` via Firecloud."
  [{:keys [workspace methodConfiguration] :as executor} reference]
  (update-method-configuration! executor reference)
  (firecloud/submit-method workspace methodConfiguration))
```

Before submitting the method, we programmatically alter the method configuration to use the snapshot reference's *name* as its root entity type.  That takes place in [`executor/update-method-configuration!`](https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/executor.clj#L138-L157).

Note that the destructuring of the `_reference` argument gives us a hint that we expect the reference passed in to be a map like `{:name "some-snapshot-reference-name" …}`. 
```
  "Update `methodConfiguration` in `workspace` with snapshot reference `name`
  Update executor table record ID with incremented `methodConfigurationVersion`."
  [{:keys [id
           workspace
           methodConfiguration
           methodConfigurationVersion] :as _executor}
…
```

We see how the name is used, talking to Firecloud to update the method configuration:
```
…
    (firecloud/update-method-configuration
     workspace methodConfiguration
     (assoc current :dataReferenceName name :methodConfigVersion inc'd))
…
```

Where this method was originally called, we passed it a [*snapshot reference object fetched from Rawls*](https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/executor.clj#L127).  Ex.
```
{:cloningInstructions "COPY_NOTHING",
  :description "",
  :name "GP20210430",
  :reference {:instanceName "terra", :snapshot "0e308334-80f0-4cb2-9988-e57a6f67f0c5"},
  :referenceId "6374c360-b980-442c-abe4-37f144b65109",
  :referenceType "DATA_REPO_SNAPSHOT",
  :workspaceId "42f490fc-6bc6-4080-af85-dc3175e84815"}
```

What you've passed to `executor/create-submission!` and by extension to `executor/update-method-configuration!` is a *snapshot reference ID* pulled from the executor details table.  Ex.
```
"6374c360-b980-442c-abe4-37f144b65109"
```

I will put in a specific recommendation above, but my general advice is that you need to pass a snapshot reference object from Rawls and not a standalone ID.

api/src/wfl/module/covid.clj
```diff
@@ -94,82 +96,122 @@
...
+      ; Get the rows from the terra details table for the workflows
+      ; iterate through the results and get the distinct reference values
+      ; resubmit the reference values
+      ; write back to the table, putting the new workflow id into the retry field of the failed one
+      (let [workflow_ids (util/to-quoted-comma-separated-list (map :uuid workflows))
+            executor (get-in workload [:executor])
+            executor_details (get-in workload [:executor :details])
+            query (format "SELECT * FROM %s WHERE workflow IN (%s)" executor_details workflow_ids)
+            results (jdbc/query (postgres/wfl-db-config) [query])
+            distinct_references (set (map :reference results))]
```
okotsopoulos, line 183: This gives you the distinct reference IDs but you want the snapshot reference objects from Rawls to pass to `executor/create-submission!` (more context below). 

Suggest:
```
distinct_references (->> (map :reference results)
                                 set
                                 (map (partial rawls/get-snapshot-reference workspace)))
```

api/src/wfl/module/covid.clj
```diff
@@ -94,82 +96,122 @@
...
+      ;     (wfl.debug/trace)
+      ;     )
+
+      ; Get the rows from the terra details table for the workflows
+      ; iterate through the results and get the distinct reference values
+      ; resubmit the reference values
+      ; write back to the table, putting the new workflow id into the retry field of the failed one
+      (let [workflow_ids (util/to-quoted-comma-separated-list (map :uuid workflows))
+            executor (get-in workload [:executor])
+            executor_details (get-in workload [:executor :details])
```
okotsopoulos, line 180: Suggest destructuring instead:

```
{keys [workspace details] :as executor} (:executor workload)
```

api/src/wfl/module/covid.clj
okotsopoulos: Something else that we may do is add a precondition to `executor/update-method-configuration!` to validate that `_reference` has a non-nil field `name`, or add some similar check.  The fact that the method configuration update happened successfully and didn't complain about the missing snapshot reference name made debugging more challenging.

api/src/wfl/module/covid.clj
```diff
@@ -94,82 +96,122 @@
...
-  (executor/executor-workflows-by-status tx executor status))
-(defoverload workloads/to-edn             pipeline workload-to-edn)
+      (-> workload
+          (util/select-non-nil-keys workload-metadata-keys)
+          (dissoc :pipeline)
+          (update :source util/to-edn)
+          (update :executor util/to-edn)
+          (update :sink util/to-edn)))
+
+(defn retry-workflows
```
okotsopoulos, line 165: (A general note so that I don't forget, not specific to your debugging today)

Ultimately the code that you're writing here looks like it should live in the `executor` namespace.  None of it seems COVID-specific, or code that bridges different stages.  A tip-off is how you've publicized several `executor` methods to make this possible.

rhiananthony: Agreed. It was always my intention to make this generalized. I just have been doing my work in the covid file, but should probably move it to the executor file.

okotsopoulos: Actually, if you want to hold off on that move, I should have a PR for executor logging tomorrow!  Then we won't step on each other's toes.

okotsopoulos:
The linked PR above takes the place of this one, closing.

PR #474, opened 2021-07-09
[GH-1402] Write Workflow Outputs to the Terra Data Repository

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1402
Add a MVP `TerraDataRepoSink` to write workflow outputs back to the Terra DataRepo. 
Note that each commit in this PR (except merging develop) has been reviewed separately.


api/src/wfl/api/handlers.clj
```diff
@@ -1,11 +1,11 @@
 (ns wfl.api.handlers
   "Define handlers for API endpoints. Require wfl.module namespaces here."
   (:require [clojure.set                    :refer [rename-keys]]
-            [wfl.log                        :as log]
             [wfl.configuration              :as config]
```
okotsopoulos, line 4: Nit: as long as we're alphabetizing, could move this line too!

api/test/wfl/integration/sinks/workspace_sink_test.clj
```diff
@@ -1,4 +1,4 @@
-(ns wfl.integration.sink-test
+(ns wfl.integration.sinks.workspace-sink-test
```
okotsopoulos, line 1: I appreciate the renaming and breaking out into separate test files.

okotsopoulos, line 1: I appreciate the renaming and breaking out into separate test files.

api/test/wfl/integration/sinks/datarepo_sink_test.clj
```diff
@@ -0,0 +1,131 @@
+(ns wfl.integration.sinks.datarepo-sink-test
+  "Test validation and operations on Sink stage implementations."
```
okotsopoulos, line 2: This comment could be updated since we're only testing one implementation here.

api/test/wfl/integration/sinks/datarepo_sink_test.clj
```diff
@@ -0,0 +1,131 @@
...
+   :outfloat  (* 4 (Math/atan 1))
+   :outint    27
+   :outstring "Hello, World!"})
+
+;; Must be a function to capture value of *dataset* after setting up test
+;; fixtures.
+(defn ^:private datarepo-sink-request
+  "Return a valid Tera DataRepo sink request."
+  []
+  {:name        @#'sink/datarepo-sink-name
```
okotsopoulos, line 39: Is the `@#'` prefix how we reference a `^:const`?

api/src/wfl/sink.clj
```diff
@@ -171,12 +196,198 @@
...
 
-(defoverload stage/done? terra-workspace-sink-type terra-workspace-sink-done?)
+(defoverload load-sink!   terra-workspace-sink-type load-terra-workspace-sink)
+(defoverload update-sink! terra-workspace-sink-type update-terra-workspace-sink)
+(defoverload stage/done?  terra-workspace-sink-type terra-workspace-sink-done?)
 
 (defoverload util/to-edn terra-workspace-sink-type terra-workspace-sink-to-edn)
+
+;; TerraDataRepo Sink
+(def ^:private ^:const datarepo-sink-name  "Terra DataRepo")
```
okotsopoulos, line 211: Sorry, can you remind me why we're tagging these as `^:const` but not the other names, types, etc. for other stage implementations?

okotsopoulos (approved):
You were right, lots to review here!  But looks good to me, aside from a few qs / nits.

api/test/wfl/integration/sinks/datarepo_sink_test.clj
ehigham: This works around the fact that `datarepo-sink-name` is private - `#` creates a `Var` from the `Symbol` `sink/datarepo-sink-name` and `@` dereferences it.

api/src/wfl/sink.clj
ehigham: They're constants so marking them as `^:const` allows the clojure compiler to inline them: https://stackoverflow.com/questions/9162558/how-does-clojure-const-work
I didn't change others because I wanted to limit scope drift.

ehigham: I've propagated `^:const`.

api/test/wfl/integration/sinks/datarepo_sink_test.clj
okotsopoulos: Ahhh cool, this is helpful to me… I ran into an issue trying to access a private var in test recently.  Worked around by making it public, but now I know.

ehigham: I'm not sure which is the best way to do this. I'm kind of against accessing private stuff but it's also easy...

PR #477, opened 2021-07-14
GH-1414 Executor falls back to fetching existing snapshot reference

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1414

Fallback to importing an existing snapshot reference when creation fails with a 409 (duplicate resource exception) stopped working.

### Changes

- Calling `rawls/create-or-get-snapshot-reference` rather than `rawls/create-snapshot-reference` to fall back to fetching an existing snapshot reference on a 409 error.
- Mocking in testing updated to mock `rawls/create-or-get…` rather than `rawls/create…`.  This previous mocking allowed tests to pass as the creation method is called within the create-or-get, but we weren't mocking what was actually being called.  If this method call is incorrectly changed back in the future, tests will fail.

### Review Instructions

I tested manually:
- Create and start a workload that tries to import a snapshot as a reference with the same name as one already in the workspace.  Observe that it succeeds.

And added new automated test `wfl.test.integration.executor-test/test-terra-executor-entity-from-snapshot`.

### DIFF
--- a/api/test/wfl/integration/executor_test.clj
+++ b/api/test/wfl/integration/executor_test.clj
@@ -104,6 +104,7 @@
 (defn ^:private mock-rawls-snapshot-reference [& _]
   {:referenceId snapshot-reference-id
    :name        snapshot-reference-name})
+()
 
 (def ^:private method-config-version-mock 1)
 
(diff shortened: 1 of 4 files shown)


ehigham (approved):
Looks good. I think it would be good to lock down that behaviour in an automated test though.

okotsopoulos:
> Looks good. I think it would be good to lock down that behaviour in an automated test though.

Yes, fair!  Better to be explicit about this expectation in testing, just added.  LMK if you had something else in mind.

api/test/wfl/integration/executor_test.clj
```diff
@@ -104,6 +104,7 @@
 (defn ^:private mock-rawls-snapshot-reference [& _]
   {:referenceId snapshot-reference-id
    :name        snapshot-reference-name})
+()
```
okotsopoulos, line 107: Whoops, will clean this up in my next PR for executor logging.

PR #478, opened 2021-07-14
[GH-1415] Implement `update-sink` for TerraDataRepoSink

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1415

This change contains a rough implementation of the TerraDataRepo sink update functionality. After discovering certain hurdles in the original design, I've made some tweaks and incorporated TDR's work https://broadworkbench.atlassian.net/browse/DR-1960 so that we only need one ingest request instead of separately loading the output files and output metadata. At the time of writing, DR-1960 is work in progress so our ingests won't work just yet.

Some things to call out in this change that I'm deferring to a follow up change into the feature branch:
- I need to upload the json file for TDR to ingest. I'm currently using our test outputs bucket as a temporary folder for this which is obviously not a production-ready solution. Possible fixes include creating a scratch bucket for workflow-launcher, creating a bucket per workload, using the executor's execution bucket... etc. Suggestions welcome.
- Since the ingests always fail, I'm leaving some of the work post-ingest as a TODO.

### DIFF
--- a/api/src/wfl/sink.clj
+++ b/api/src/wfl/sink.clj
@@ -251,11 +251,11 @@
       (postgres/throw-unless-table-exists tx details)
       (first (jdbc/query tx (format query details))))))
 
-(defn ^:private pop-job-queue! [{:keys [details] :as sink}]
-  (if-let [{:keys [id]} (peek-job-queue sink)]
-    (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-      (jdbc/update! tx details {:consumed (utc-now)} ["id = ?" id]))
-    (throw (ex-info "TerraDataRepoSink job queue is empty" {:sink sink}))))
+(defn ^:private pop-job-queue!
+  [{:keys [details] :as _sink} {:keys [id] :as _job}]
+  {:pre [(some? id) (integer? id)]}
+  (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+    (jdbc/update! tx details {:consumed (utc-now)} ["id = ?" id])))
 
 (defn ^:private update-datarepo-job-statuses
   [{:keys [details] :as _sink}]
@@ -308,13 +308,13 @@
     (start-ingesting-outputs sink workflow)
     (stage/pop-queue! executor))
   (update-datarepo-job-statuses sink)
-  (when-let [{:keys [workflow job]} (peek-job-queue sink)]
+  (when-let [{:keys [workflow job] :as record} (peek-job-queue sink)]
     (try
       (let [res (datarepo/get-job-result job)]
         (log/info "Sunk workflow outputs to dataset"
                   {:dataset (get-in sink [:dataset :id]) :workflow workflow}))
       (finally
-        (pop-job-queue! sink)))))
+        (pop-job-queue! sink record)))))
 
 (defn ^:private datarepo-sink-done? [{:keys [details] :as _sink}]
   (let [query "SELECT COUNT(*) FROM %s
--- a/api/src/wfl/sink.clj
+++ b/api/src/wfl/sink.clj
@@ -278,9 +278,10 @@
                             {:status status :updated (utc-now)}
                             ["id = ?" job-id])))]
     (doseq [[id job-id workflow] (read-active-jobs)]
-      (let [{:keys [job_status] :as job} (datarepo/job-metadata job-id)]
-        (write-job-status id job_status)
-        (when (= "failed" job_status)
+      (let [job    (datarepo/job-metadata job-id)
+            status (-> job :job_status str/lower-case)]
+        (write-job-status id status)
+        (when (= "failed" status)
           (throw (UserException. "A DataRepo ingest job failed"
                                  {:job      job
                                   :workflow workflow})))))))
--- a/api/src/wfl/sink.clj
+++ b/api/src/wfl/sink.clj
@@ -245,7 +245,7 @@
 
 (defn ^:private peek-datarepo-sink-job-queue [{:keys [details] :as _sink}]
   (let [query "SELECT * FROM %s
-               WHERE status <> 'failed' AND consumed IS NOT NULL
+               WHERE status = 'succeeded' AND consumed IS NOT NULL
                ORDER BY id ASC
                LIMIT 1"]
     (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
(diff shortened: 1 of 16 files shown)


api/src/wfl/sink.clj
```diff
@@ -242,21 +243,104 @@
                                 :fromOutputs fromOutputs}))))
     (assoc request :dataset dataset')))
 
+(defn ^:private push-job [{:keys [details] :as _sink} job]
+  (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+    (jdbc/insert! tx details job)))
+
+(defn ^:private peek-job-queue [{:keys [details] :as _sink}]
+  (let [query "SELECT * FROM %s
+               WHERE status = 'succeeded' AND consumed IS NULL
```
okotsopoulos, line 252: Double checking that succeeded is meant to be in lowercase.

ehigham: yeah - tdr job statuses are all lower case. Probably should to a `to-lower` to make sure.

api/src/wfl/sink.clj
```diff
@@ -242,21 +243,104 @@
...
+(defn ^:private update-datarepo-job-statuses
+  "Fetch and record the status of all 'running' jobs from tdr created by the
+   `sink`. Throws `UserException` when any new job status is `failed`."
+  [{:keys [details] :as _sink}]
+  (letfn [(read-active-jobs []
+            (let [query "SELECT id, job, workflow FROM %s
+                         WHERE status = 'running'
+                         ORDER BY id ASC"]
+              (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+                (map (juxt :id :job :workflow)
```
okotsopoulos, line 273: Neat, haven't seen `juxt` before.

api/src/wfl/sink.clj
```diff
@@ -242,21 +243,104 @@
...
+                     (jdbc/query tx (format query details))))))
+          (write-job-status [job-id status]
```
okotsopoulos, line 275: The `job-id` name here is confusing.  Is this meant to be the row ID or the job ID, from where this is called / how this is used I think row ID?

api/src/wfl/sink.clj
```diff
@@ -242,21 +243,105 @@
...
+            (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+              (jdbc/update! tx details
+                            {:status status :updated (utc-now)}
+                            ["id = ?" job-id])))]
+    (doseq [[id job-id workflow] (read-active-jobs)]
+      (let [job    (datarepo/job-metadata job-id)
+            status (-> job :job_status str/lower-case)]
+        (write-job-status id status)
+        (when (= "failed" status)
+          (throw (UserException. "A DataRepo ingest job failed"
```
okotsopoulos, line 285: This reminds me, maybe we should throw when TDR source snapshot creation jobs fail too (outside of the scope of this ticket).  More generally, a failure condition that would result in a stage's queue element never being eligible for consumption would benefit from an exception propagated.

api/src/wfl/sink.clj
```diff
@@ -242,21 +243,105 @@
...
+  "Start ingesting the `workflow` outputs as a new row in the target dataset
+   and push the tdr ingest job into the `sink`'s job queue."
+  [{:keys [dataset table fromOutputs] :as sink}
+   {:keys [uuid] :as workflow}]
+  (let [file (storage/gs-url "broad-gotc-dev-wfl-ptc-test-outputs" (str uuid ".json"))]
+    (-> (to-dataset-row fromOutputs workflow)
+        (json/write-str :escape-slash false)
+        (storage/upload-content file))
+    (->> (datarepo/ingest-table (:id dataset) file table)
+         (assoc {:status "running" :workflow uuid} :job)
```
okotsopoulos, line 313: We're positive that if we've gotten past `datarepo/ingest-table` successfully, then "running" is the correct status for the job?  Can we get this from an external service instead of hard-coding?  I noticed that `datarepo/ingest-table` pulls out the job id from the response and returns that, I wonder if the status is available in the response.

api/src/wfl/sink.clj
```diff
@@ -242,21 +243,105 @@
...
+  (when-let [[_ workflow] (stage/peek-queue executor)]
+    (start-ingesting-outputs sink workflow)
+    (stage/pop-queue! executor))
+  (update-datarepo-job-statuses sink)
+  (when-let [{:keys [workflow job] :as record} (peek-job-queue sink)]
+    (try
+      (let [res (datarepo/get-job-result job)]
+        (log/info "Sunk workflow outputs to dataset"
+                  {:dataset (get-in sink [:dataset :id]) :workflow workflow}))
+      (finally
```
okotsopoulos, line 329: Can you explain why this is in a `finally`, if we failed to get the job result then I'd think that the sink wasn't successful?

api/src/wfl/sink.clj
```diff
@@ -242,21 +243,105 @@
...
+        (pop-job-queue! sink record)))))
 
-(defn ^:private datarepo-sink-done? [{:keys [details] :as _sink}]
+(defn ^:private datarepo-sink-done?
+  "True when all tdr ingest jobs created by the `_sink` have terminated."
+  [{:keys [details] :as _sink}]
   (let [query "SELECT COUNT(*) FROM %s
-               WHERE status   <> 'failed'
-               AND   consumed IS NOT NULL"]
+               WHERE status <> 'failed' AND consumed IS NULL"]
```
okotsopoulos, line 336: (Fun with Postgres) if an unconsumed record has null status it won't be included in the count, not sure if that's relevant for this sink impl.

api/src/wfl/sink.clj
ehigham: It's meant to be the job ID. Its called on line 283 within this function.

api/test/wfl/integration/executor_test.clj
```diff
@@ -24,7 +24,7 @@
   {:type testing-queue-type :queue (ArrayDeque. items)})
 
 (defn ^:private testing-queue-peek [this]
```
okotsopoulos, line 26: We repeat these testing queue def'ns in a few different tests, could be candidates to pull out to`wfl.integration.modules.shared`.

api/src/wfl/sink.clj
ehigham: Yeah we probably should. Think right now we just log it.

api/src/wfl/sink.clj
ehigham: Ah so right now `get-job-result` throws if the job has failed as TDR sends back a non-200 response. Finally is to make sure we shift past the bad job (we've already thrown in `update-jobs` so there's no point in erroring again)

api/src/wfl/sink.clj
ehigham: Thanks for spotting that! I don't think it can have a null status. It's certainly an error if it does.

api/test/wfl/integration/executor_test.clj
ehigham: Yeah seems like a reasonable thing to do. I keep finding bugs and have to fix them all over!

api/src/wfl/sink.clj
ehigham: I would prefer to get them form another service alas no service enumerates them. These values are lifted from the prod swagger page so I'm sure they're correct for now...

ehigham: It is available in the response yes but we know it's running when we create the job. I suppose we should use the value tdr gives us but that's a bigger change.

okotsopoulos (approved):
Especially considering this is getting merged to a feature branch, looks good!

api/src/wfl/sink.clj
okotsopoulos: Got it, I had been thinking that we popped from the queue only when the job was successfully consumed.

api/src/wfl/sink.clj
okotsopoulos: Agreed, that's a bigger scope.

api/src/wfl/sink.clj
ehigham: Oh wait - no i'm wrong.

PR #480, opened 2021-07-23
[GH-1139] Logging level and configuration

rfricke-asymmetrik:
### Purpose
Adds configuration for setting the severity level of logs that are printed to stdout.

### Changes
Added a configuration table for key value pairs. Currently used to hold a logging level config.
Added a logging-level atom to the log namespace to limit messages to only enabled severity levels.
Added endpoints for getting the current logging level and updating/inserting configurations.
Added a timed check for current level to set the current logging level from the configuration table.

### Review Instructions
Verify that all tests pass.
Verify that the logging_level GET and POST request return the current level and set the current level respectively.
Verify that no messages below the current level print to stdout (give time for the delay after changing the level).

### DIFF
--- a/api/src/wfl/configuration.clj
+++ b/api/src/wfl/configuration.clj
@@ -1,0 +1,27 @@
+(ns wfl.configuration
+  "Configuration for WFL. A configuration table exists as a simple key
+   value store for simple configuration."
+  (:require [clojure.string                 :as str]
+            [wfl.jdbc                       :as jdbc]))
+
+(def ^:private configuration-table "configuration")
+
+(defn load-logging-level
+  "Use transaction `tx` to load the current logging level of the application."
+  [tx]
+  (let [levels (jdbc/query tx ["SELECT value FROM configuration WHERE key = 'LOGGING_LEVEL' LIMIT 1"])]
+    (if (empty? levels)
+      (-> :info name str/upper-case)
+      (str/upper-case (get (first levels) :value)))))
+
+(defn upsert-config
+  "Use transaction `tx` to insert or update a configuration row."
+  [tx key value]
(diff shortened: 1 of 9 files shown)


ehigham (approved):

api/src/wfl/configuration.clj
```diff
@@ -0,0 +1,27 @@
...
+    (if (empty? levels)
+      (-> :info name str/upper-case)
+      (str/upper-case (get (first levels) :value)))))
+
+(defn upsert-config
+  "Use transaction `tx` to insert or update a configuration row."
+  [tx key value]
+  (letfn [(get-row [] (jdbc/query tx [(str "SELECT * FROM configuration WHERE key = '" key "' LIMIT 1")]))]
+    (if (empty? (get-row))
+      (if (empty? (jdbc/insert! tx configuration-table {:key key :value value}))
```
ehigham, line 22: The postgres driver returns the newly inserted row so I don't think the additional call to `get-row` is needed!

api/src/wfl/configuration.clj
```diff
@@ -6,22 +6,20 @@
...
+  (let [configs (jdbc/query tx ["SELECT value FROM configuration WHERE key = ?" key])]
+    (if (empty? configs)
       (-> :info name str/upper-case)
-      (str/upper-case (get (first levels) :value)))))
+      (str/upper-case (get (first configs) :value)))))
 
 (defn upsert-config
   "Use transaction `tx` to insert or update a configuration row."
   [tx key value]
   (letfn [(get-row [] (jdbc/query tx [(str "SELECT * FROM configuration WHERE key = '" key "' LIMIT 1")]))]
```
ehigham, line 20: Can you also make this query use a prepared statement?

PR #482, opened 2021-07-28
[GH-1394] TerraExecutor retry implementation and Sarscov2IlluminaFull integration

okotsopoulos:
### Purpose

- Ticket [GH-1394](https://broadinstitute.atlassian.net/browse/GH-1394)
- @tbl3rd's [spike doc](https://broadinstitute.atlassian.net/wiki/spaces/GHConfluence/pages/2674393095/Retries+in+Terra+WFL) -- thanks!
- Rhian's [original PR](https://github.com/broadinstitute/wfl/pull/472) which this replaces -- thanks!

Retry functionality is now supported in staged workflows within their executors, with COVID workloads altered to take advantage.

`TerraExecutor` retry resubmits the snapshot reference associated with the specified workflows, and updates each original workflow record to point to its retry record.  The existing update loop will handle updating workflow uuids and statuses for us.

### Changes

In `wfl.handlers/post-retry`:
- Raise HTTP 400 if passing in an unsupported status, or if no workflows available for a supported status

In `wfl.executor`:
- Added multimethod `executor-retry-workflows!` dispatched on `:type` with `TerraExecutor` implementation `terra-executor-retry-workflows`
- Added logging throughout

In `wfl.module.covid`, previously retries were unsupported.  Now we dispatch to `retry-covid-workload` which...
- Confirms that the workload has started
- Retries via the executor
- Reanimates the workload if needed so that it will be visible to the update loop
- Returns the updated workload

In `wfl.integration.executor-test`:
- Added new automated test `test-terra-executor-retry-workflows` and adjusted existing mock utilities to accommodate 

In `wfl.system.v1_endpoint_test/test-retry-workload`:
- Checking for expected errors if status unsupported or no workflows available for a supported status

### Manual Testing

I created a workload associated with workspace [okotsopo/CDC_Viral_Sequencing_okotsopo_20210707](https://app.terra.bio/#workspaces/wfl-dev/CDC_Viral_Sequencing_okotsopo_20210707/) and started it.  Then I aborted the resulting running workflow and observed that the workload was marked as finished.

I was then able to trigger a retry of the workflow with the following:
```
  (let [workload-uuid "e66c86b2-120d-4f7f-9c3a-b9eaadeb1919"
         workload           (workloads/load-workload-for-uuid tx workload-uuid)
         status                "Aborted"
         workflows         (workloads/workflows-by-status tx workload status)]
    (workloads/retry workload workflows)))
```

And observed the new workflow running in the Terra UI, and its record linked to its original via the `retry` column.
Screen Shot 2021-07-28 at 2 33 28 PM

I also observed that the workload update loop took care of updating retry workflow uuids and statuses, and appropriately remarked the workload as failed once they had finished.

Same behavior confirmed when hitting the endpoint:

```
wm111-e35:wfl okotsopo$ curl -X POST "http://localhost:3000/api/v1/workload/e66c86b2-120d-4f7f-9c3a-b9eaadeb1919/retry" \
> -H "Content-Type: application/json" \
> -d '{ "status": "Failed" }'
{
  "started" : "2021-07-14T15:36:47Z",
  "watchers" : [ "okotsopo@broadinstitute.org" ],
  "labels" : [ "hornet:test", "project:okotsopo testing enhanced source, executor, sink logging", "project:wfl-dev/CDC_Viral_Sequencing_okotsopo_20210707" ],
  "creator" : "okotsopo@broadinstitute.org",
  "updated" : "2021-07-28T22:49:30Z",
  "created" : "2021-07-14T15:36:07Z",
  "source" : {
    …
  },
  "commit" : "9719eda7424bf5b0804f5493875681fa014cdb29",
  "uuid" : "e66c86b2-120d-4f7f-9c3a-b9eaadeb1919",
  "executor" : {
    "workspace" : "wfl-dev/CDC_Viral_Sequencing_okotsopo_20210707",
    "methodConfiguration" : "wfl-dev/sarscov2_illumina_full",
    "methodConfigurationVersion" : 38,
    "fromSource" : "importSnapshot",
    "name" : "Terra"
  },
  "version" : "0.8.0",
  "sink" : {
    …
    },
    "identifier" : "run_id",
    "name" : "Terra Workspace"
  }
}
```

### Review Instructions

- Please confirm that implementation matches spike doc's specifications
- I was looking at this existing system test **[test-automate-sarscov2-illumina-full](https://github.com/broadinstitute/wfl/blob/develop/api/test/wfl/system/automation_test.clj)** to see if it could be altered to retry as well.  But I wasn't convinced that in its current form it's useful / doing what we want.  Notably, it never stops the workload, so will it ever finish / possibly clog up the DB?

### TODO

In separate PRs:
- Update [WFL GitHub page](https://github.com/broadinstitute/wfl/blob/develop/docs/md/modules-covid.md) for Sarscov2IlluminaFull to include info on new `/retry` endpoint.
- Tweak logging, see if new log functions accept maps.  More discussion [here](https://github.com/broadinstitute/wfl/pull/482#discussion_r683643315).

### Spin-Off Tickets

[GH-1421](https://broadinstitute.atlassian.net/browse/GH-1421) - `stage/validate-or-throw` redesigned to accommodate workload retry as well as creation
[GH-1422](https://broadinstitute.atlassian.net/browse/GH-1422) - Terra Executor should deal in generic entities and not assume snapshot references
[GH-1423](https://broadinstitute.atlassian.net/browse/GH-1423) - Terra Executor retry should be thread-safe

### DIFF
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -98,7 +98,7 @@
            (mapv util/to-edn)
            succeed))))
 
-;; Visible for wfl.system.v1-endpoint-test/test-retry-workload
+;; Visible for testing
 (def retry-unsupported-status-error-message
   "Retry unsupported for requested status.")
 (def retry-no-workflows-error-message
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -102,6 +102,8 @@
   (log/info (select-keys request [:request-method :uri :parameters]))
   (let [uuid   (get-in request [:path-params :uuid])
         status (get-in request [:body-params :status])]
+    ;; TODO: if status='Succeeded', no op / error?
+    ;; TODO: if no workflows found, no op / error?
     (->> (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
            (let [w (workloads/load-workload-for-uuid tx uuid)]
              [w (workloads/workflows-by-status tx w status)]))
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -103,14 +103,16 @@
   (log/info (select-keys request [:request-method :uri :parameters]))
   (let [uuid   (get-in request [:path-params :uuid])
         status (get-in request [:body-params :status])]
-    ;; TODO: If status not in whitelist, throw.
+    ;; TODO: Refactor to workflow-status whitelist
     ;; https://broadinstitute.atlassian.net/browse/GH-1424
+    (when (= "Succeeded" status)
+      (throw (UserException. "Retry unsupported for workflow status."
+                             {:uuid            uuid
+                              :workflow-status status
+                              :status          400})))
     (->> (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
            (let [workload  (workloads/load-workload-for-uuid tx uuid)
                  workflows (workloads/workflows-by-status tx workload status)]
-             ;; TODO: If no workflows, throw.
-             ;; System tests leveraging v1-endpoint-test/test-retry-workload
-             ;; will need to be updated.
              [workload workflows]))
          (apply workloads/retry)
          util/to-edn
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -12,6 +12,7 @@
             [wfl.module.sg]
             [wfl.module.wgs]
             [wfl.module.xx]
+            [wfl.service.cromwell           :as cromwell]
             [wfl.service.google.storage     :as gcs]
             [wfl.service.postgres           :as postgres]
             [wfl.util                       :as util])
@@ -97,22 +98,33 @@
            (mapv util/to-edn)
            succeed))))
 
+;; Visible for wfl.system.v1-endpoint-test/test-retry-workload
+(def retry-unsupported-status-error-message
+  "Retry unsupported for requested status.")
+(def retry-no-workflows-error-message
+  "No workflows to retry for requested status.")
+
 (defn post-retry
   "Retry the workflows identified in `request`."
   [request]
   (log/info (select-keys request [:request-method :uri :parameters]))
-  (let [uuid   (get-in request [:path-params :uuid])
-        status (get-in request [:body-params :status])]
-    ;; TODO: Refactor to workflow-status whitelist
-    ;; https://broadinstitute.atlassian.net/browse/GH-1424
-    (when (= "Succeeded" status)
-      (throw (UserException. "Retry unsupported for workflow status."
-                             {:uuid            uuid
-                              :workflow-status status
-                              :status          400})))
+  (let [uuid       (get-in request [:path-params :uuid])
+        status     (get-in request [:body-params :status])
+        supported? (cromwell/retry-status? status)]
+    (when-not supported?
+      (throw (UserException. retry-unsupported-status-error-message
+                             {:uuid               uuid
+                              :supported-statuses cromwell/retry-status?
+                              :requested-status   status
+                              :status             400})))
     (->> (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
            (let [workload  (workloads/load-workload-for-uuid tx uuid)
                  workflows (workloads/workflows-by-status tx workload status)]
+             (when (empty? workflows)
+               (throw (UserException. retry-no-workflows-error-message
+                                      {:uuid             uuid
+                                       :requested-status status
+                                       :status           400})))
              [workload workflows]))
          (apply workloads/retry)
          util/to-edn
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -14,7 +14,8 @@
             [wfl.module.xx]
             [wfl.service.google.storage     :as gcs]
             [wfl.service.postgres           :as postgres]
-            [wfl.util                       :as util]))
+            [wfl.util                       :as util])
+  (:import  [wfl.util UserException]))
 
 (defn succeed
   "A successful response with BODY."
@@ -102,11 +103,15 @@
   (log/info (select-keys request [:request-method :uri :parameters]))
   (let [uuid   (get-in request [:path-params :uuid])
         status (get-in request [:body-params :status])]
-    ;; TODO: if status='Succeeded', no op / error?
-    ;; TODO: if no workflows found, no op / error?
+    ;; TODO: If status not in whitelist, throw.
+    ;; https://broadinstitute.atlassian.net/browse/GH-1424
     (->> (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-           (let [w (workloads/load-workload-for-uuid tx uuid)]
-             [w (workloads/workflows-by-status tx w status)]))
+           (let [workload  (workloads/load-workload-for-uuid tx uuid)
+                 workflows (workloads/workflows-by-status tx workload status)]
+             ;; TODO: If no workflows, throw.
+             ;; System tests leveraging v1-endpoint-test/test-retry-workload
+             ;; will need to be updated.
+             [workload workflows]))
          (apply workloads/retry)
          util/to-edn
          succeed)))
--- a/api/src/wfl/executor.clj
+++ b/api/src/wfl/executor.clj
@@ -448,10 +448,10 @@
      executor
      (jdbc/query tx [(format query details) status]))))
 
-(defn ^:private reference-id-to-workflow-records
-  "Find all submission IDs associated with `workflows`.
-  Return a mapping of the submissions' snapshot reference IDs
-  to their associated workflow records."
+(defn ^:private workflow-and-sibling-records
+  "Return the workflow records for all workflows in submissions
+  associated with the specified `workflows` --
+  i.e. records for `workflows` and their siblings."
   [tx {:keys [details] :as executor} workflows]
   (postgres/throw-unless-table-exists tx details)
   (let [workflow-ids (util/to-quoted-comma-separated-list (map :uuid workflows))
@@ -472,9 +472,7 @@
                WHERE submission IN
                  (SELECT DISTINCT submission FROM %s
                  WHERE workflow IN %s)"]
-    (->> (format query details details workflow-ids)
-         (jdbc/query tx)
-         (group-by :reference))))
+    (jdbc/query tx (format query details details workflow-ids))))
 
 (defn update-retried-workflow-records
   "Match `original-records` with `retry-records` on entity uuid,
@@ -517,7 +515,7 @@
   "Resubmit the snapshot references associated with `workflows` in `workspace`
   and update each original workflow record with the row ID of its retry."
   [{:keys [workspace] :as executor} workflows]
-  (letfn [(retried-workflow-records [reference-id]
+  (letfn [(submit-reference [reference-id]
             ;; Further work required to deal in generic entities
             ;; rather than assumed snapshot references:
             ;; https://broadinstitute.atlassian.net/browse/GH-1422
@@ -526,8 +524,9 @@
                    (create-submission! executor)
                    (allocate-submission executor reference))))]
     (->> (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-           (reference-id-to-workflow-records tx executor workflows))
-         (map-keys retried-workflow-records)
+           (workflow-and-sibling-records tx executor workflows))
+         (group-by :reference)
+         (map-keys submit-reference)
          (run! (partial update-retried-workflow-records executor)))))
 
 (defn ^:private terra-executor-done? [executor]
--- a/api/src/wfl/executor.clj
+++ b/api/src/wfl/executor.clj
@@ -10,8 +10,7 @@
             [wfl.service.firecloud :as firecloud]
             [wfl.service.postgres  :as postgres]
             [wfl.service.rawls     :as rawls]
-            [wfl.stage             :as stage :refer [log-prefix
-                                                     throw-if-no-details-table]]
+            [wfl.stage             :as stage :refer [log-prefix]]
             [wfl.util              :as util :refer [map-keys utc-now]])
   (:import [wfl.util UserException]))
 
@@ -423,7 +422,7 @@
 (defn ^:private terra-executor-workflows
   "Return all the non-retried workflows executed by the `executor`."
   [tx {:keys [details] :as executor}]
-  (throw-if-no-details-table tx executor)
+  (postgres/throw-unless-table-exists tx details)
   (let [query "SELECT * FROM %s
                WHERE workflow IS NOT NULL
                AND   retry    IS NULL
@@ -436,7 +435,7 @@
   "Return all the non-retried workflows matching `status` executed by the
   `executor`."
   [tx {:keys [details] :as executor} status]
-  (throw-if-no-details-table tx executor)
+  (postgres/throw-unless-table-exists tx details)
   (let [query "SELECT * FROM %s
                WHERE workflow IS NOT NULL
                AND retry      IS NULL
@@ -451,7 +450,7 @@
   Return a mapping of the submission reference (ex. snapshot reference id)
   to their associated workflow records."
   [tx {:keys [details] :as executor} workflows]
-  (throw-if-no-details-table tx executor)
+  (postgres/throw-unless-table-exists tx details)
   (let [workflow-ids (util/to-quoted-comma-separated-list (map :uuid workflows))
         _            (-> "%s Fetching submission IDs linked to workflows %s"
                          (format (log-prefix executor) workflow-ids)
--- a/api/src/wfl/executor.clj
+++ b/api/src/wfl/executor.clj
@@ -34,7 +34,7 @@
    matching `status`"
   (fn [_transaction executor _status] (:type executor)))
 
-(defmulti retry-executor!
+(defmulti executor-retry-workflows!
   "Retry/resubmit the `workflows` managed by the `executor`."
   (fn [executor _workflows] (:type executor)))
 
@@ -240,20 +240,21 @@
              (map zip-record (sort-by :entity records))
              (write-workflow-statuses (utc-now)))))))
 
-(def ^:private final-statuses (util/to-quoted-comma-separated-list rawls/final-statuses))
+(def ^:private active-workflows-query-template
+  "Query template for fetching active workflows
+  from an executor details table to be specified at runtime."
+  (let [final-statuses (util/to-quoted-comma-separated-list rawls/final-statuses)]
+    (format "SELECT * FROM %%s
+             WHERE submission IS NOT NULL
+             AND   workflow   IS NOT NULL
+             AND   status     NOT IN %s
+             ORDER BY id ASC"
+            final-statuses)))
 
 (defn ^:private update-terra-workflow-statuses!
   "Update statuses in `details` table for active `workspace` workflows."
   [{:keys [workspace details] :as executor}]
-  (letfn [(read-active-workflows []
-            (let [query "SELECT * FROM %s
-                         WHERE submission IS NOT NULL
-                         AND   workflow   IS NOT NULL
-                         AND   status     NOT IN %s
-                         ORDER BY id ASC"]
-              (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-                (jdbc/query tx (format query details final-statuses)))))
-          (update-status-from-firecloud
+  (letfn [(update-status-from-firecloud
             [{:keys [submission workflow] :as record}]
             (->> (firecloud/get-workflow workspace submission workflow)
                  :status
@@ -267,7 +268,8 @@
                (sort-by :entity records))))]
     (log/debug (format "%s Updating statuses for active workflows..."
                        (log-prefix executor)))
-    (->> (read-active-workflows)
+    (->> (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+           (jdbc/query tx (format active-workflows-query-template details)))
          (mapv update-status-from-firecloud)
          (write-workflow-statuses (utc-now)))))
 
@@ -511,7 +513,7 @@
          (map link-retry-to-original (sort-by :entity original-records))
          (write-retries (utc-now)))))
 
-(defn ^:private retry-terra-executor
+(defn ^:private terra-executor-retry-workflows
   "Resubmit the snapshot references associated with `workflows` in `workspace`
   and update each original workflow record with the row ID of its retry."
   [{:keys [workspace] :as executor} workflows]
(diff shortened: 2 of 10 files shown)


api/src/wfl/executor.clj
```diff
@@ -247,6 +264,9 @@
                  (jdbc/update! tx details {:status status :updated now}
                                ["id = ?" id]))
                (sort-by :entity records))))]
+    (log/debug (format "%s Updating statuses for eligible workflows..."
+                       (log-prefix executor)))
+    ;; TODO: do we still need to update failed statuses in light of retry design?
```
okotsopoulos, line 269: Flagging this for discussion.  A failed workflow itself wouldn't need to be reconsidered in the update loop, as we'll link it to its retry via the `retry` column.

api/src/wfl/executor.clj
```diff
@@ -428,6 +446,80 @@
...
+  "Find all submission IDs associated with `workflows`.
+  Return a mapping of the submissions' snapshot reference IDs
+  to their associated workflow records."
+  [tx {:keys [details] :as executor} workflows]
+  (throw-if-no-details-table tx executor)
+  (let [workflow-ids (util/to-quoted-comma-separated-list (map :uuid workflows))
+        _            (-> "%s Fetching submission IDs linked to workflows %s"
+                         (format (log-prefix executor) workflow-ids)
+                         log/debug)
+        ;; Presently Rawls doesn't support submitting a subset of a snapshot:
```
okotsopoulos, line 459: Flagging this, if Rawls is updated to support submitting snapshots of the subset then this will need to change.

api/src/wfl/module/covid.clj
```diff
@@ -152,6 +152,22 @@
...
+  (when-not finished
+    (throw (UserException. "Cannot retry workload before it's finished."
+                           {:workload workload})))
+  ;; TODO: validate executor and sink.
+  ;; Existing stage/validate-or-throw dispatches on :name.
+  ;; The workload's executor and sink have :type at this stage, not :name.
+  (executor/retry-executor! executor workflows)
+  (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+    (when-not (stage/done? executor)
+      (patch-workload tx workload {:finished nil :updated (utc-now)}))
```
okotsopoulos, line 168: I got burned by this!

Initially, I revived the workload *before* the retry.  Then I noticed a race condition where the update loop might mark the workload as finished before the retry workflow records could be written.

api/src/wfl/executor.clj
```diff
@@ -428,6 +446,80 @@
...
+  ;; may not always be dealing with references, should handle generic entity.
+  (letfn [(retried-workflow-records [reference-id]
+            (let [reference (rawls/get-snapshot-reference workspace reference-id)]
+              (->> reference
+                   (create-submission! executor)
+                   (allocate-submission executor reference))))]
+    (->> (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+           (reference-id-to-workflow-records tx executor workflows))
+         (map-keys retried-workflow-records)
+         (run! (partial update-retried-workflows executor)))))
```
okotsopoulos, line 521: I got burned by this too!  Initially used a `map` which yields a lazy sequence.  When I cleaned up debug printing, this was no longer called.  We're running the function for its side effects rather than its returned value.

api/src/wfl/module/covid.clj
```diff
@@ -152,6 +152,22 @@
                              {:workload workload})))
     (if-not (or stopped finished) (stop! workload (utc-now)) workload)))
 
+(defn ^:private retry-covid-workload
+  "Retry/resubmit the `workflows` managed by the `workload` and return the
+   workload that manages the new workflows."
+  [{:keys [finished id executor] :as workload} workflows]
+  (when-not finished
```
okotsopoulos, line 159: I did note this specification from the spike doc:

"The /retry request reanimates a workload if it was finished **or stopped**…"

But I elected only to check if the workload had finished, since I don't think we want to allow a retry for a workload that's still running in any capacity.  LMK if this is incorrect.

api/src/wfl/module/covid.clj
```diff
@@ -152,6 +152,22 @@
...
+    (throw (UserException. "Cannot retry workload before it's finished."
+                           {:workload workload})))
+  ;; TODO: validate executor and sink.
```
okotsopoulos, line 162: Seeking feedback:

Existing validation functionality is primed to validate a source/executor/sink request, not the resulting source/executor/sink itself.

I wasn't sure of the best Clojure-y way to take advantage of the existing validation functions, ex. for [TerraExecutor](https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/executor.clj#L446).  My idea is a multimethod to tag a stage with its `:name`, but I also looked at dispatching on multiple values (seems more elegant but I got confused).

api/src/wfl/module/covid.clj
ehigham: I think it's is right - stopped basically means stop looking for new inputs. It doesn't mean stop processing.

api/src/wfl/module/covid.clj
ehigham: No I don't think I agree. I think one should be able to retry failed workflows while the workload is running.

api/src/wfl/api/handlers.clj
```diff
@@ -102,6 +102,8 @@
   (log/info (select-keys request [:request-method :uri :parameters]))
   (let [uuid   (get-in request [:path-params :uuid])
         status (get-in request [:body-params :status])]
+    ;; TODO: if status='Succeeded', no op / error?
```
okotsopoulos, line 105: The ticket had this recommendation:

"Consider no-oping for retries where the status is “Succeeded” to avoid workflow-launcher blowing up."

I wanted to see if we wanted to throw an error in this case?  Same question if no workflows are found for the status.

api/src/wfl/executor.clj
```diff
@@ -428,6 +446,91 @@
...
+        _            (-> "%s Fetching submission IDs linked to workflows %s"
+                         (format (log-prefix executor) workflow-ids)
+                         log/debug)
+        ;; Presently Rawls doesn't support submitting a subset of a snapshot:
+        ;; If we wish to retry 1+ workflows stemming from a snapshot,
+        ;; we must resubmit the snapshot in its entirety.
+        query "SELECT * FROM %s
+               WHERE submission IN
+                 (SELECT DISTINCT submission FROM %s
+                 WHERE workflow IN %s)"]
```
ehigham, line 465: Don't you want distinct references rather than submissions? Seems like the logic is a bit twisted?

api/src/wfl/executor.clj
```diff
@@ -428,6 +446,91 @@
...
+    (->> (format query details details workflow-ids)
+         (jdbc/query tx)
+         (group-by :reference))))
+
+(defn ^:private entity-from-id
+  "Use `fromSource` to direct fetch of existing `workspace` entity
+  identified by `entity-id`."
```
ehigham, line 472: This doesn't make sense to me, at least `fromSource` is being used in a way that I didn't intend. `fromSource` takes whatever the `Source` emitted and converts it into something that the executor can understand (i.e. an Entity`). Ideally, one would pattern match on the entity to tell what kind of thing your dealing with. This being clojure, nearest thing we can do is use tag-value pairs:

```clojure
(when-let [obj (peek-source s)]
  (let [[tag value :as entity] (coerce fromSource obj)]
    (case tag
      "Reference" ;; use reference
      "Row"       ;; use row
``` 

Thereafter, the executor knows what that `entity` is and `fromSource` is meaningless - one just needs to a case analysis on the entity to figure out what to do with it.

Sadly this isn't how it was written because we were in a rush. I maintain that using `fromSource` for any other purpose is the wrong thing to do.

api/src/wfl/executor.clj
```diff
@@ -403,8 +423,7 @@
 (defn ^:private terra-executor-workflows
   "Return all the non-retried workflows executed by the `executor`."
   [tx {:keys [details] :as executor}]
-  (when-not (postgres/table-exists? tx details)
-    (throw (ex-info "Missing executor details table" {:table details})))
+  (throw-if-no-details-table tx executor)
```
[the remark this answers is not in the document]
okotsopoulos: Good call, I will use that.

api/src/wfl/module/covid.clj
okotsopoulos: This decision is how I interpreted the [Spike doc](https://broadinstitute.atlassian.net/wiki/spaces/GHConfluence/pages/2674393095/Retries+in+Terra+WFL#Proposal):

"The /retry request … reanimates a workload if it was finished or stopped, …"

But the right interpretation may instead be to reanimate if needed, but not require the workload to be finished to retry.

The reason I put in this restriction is I was worried about overloading resources given the Rawls limitation on submitting snapshot subsets.  So resubmitting a snapshot with 500 resulting workflows, 1 failure and the rest running could result in concurrently running workflows for the same entity.  When we can submit a subset of a snapshot then I'd advise taking this restriction out.

api/src/wfl/executor.clj
okotsopoulos: I flip-flopped between distinct references and submissions, agreed that it is twisty.

My thoughts were that…
- When a reference is submitted successfully it yields one submission
- One submission maps to one set of workflows
- Given the capacity for retries, one reference can map to many sets of workflows

That Rawls limitation on submitting snapshot subsets again makes things hairy here.  I want to get all of the workflow records for the same submissions as the ones we need to retry, since they'll all be retried.

A possible alternative, though I couldn't convince myself 100% of its infallibility it may make more sense to everyone else:
```
"SELECT * FROM <details>
 WHERE retry IS NULL
 AND reference IN 
     (SELECT DISTINCT reference FROM %s
     WHERE workflow IN %s)

 ```

File: api/src/wfl/module/covid.clj (hunk 6 above)
Comment (ehigham) [reply api/src/wfl/module/covid.clj:162]: I think we should communicate this limitation rather than restrict people from doing what they need to do.

File: api/src/wfl/executor.clj (hunk 8 above)
Comment (okotsopoulos) [reply api/src/wfl/executor.clj:464]: I will add a comment to this effect and file a ticket to make this threadsafe.

File: api/src/wfl/executor.clj (hunk 9 above)
Comment (okotsopoulos) [reply api/src/wfl/executor.clj:471]: File a ticket to switch `reference` in executor details table to `entity` that's like `[tag value]` pair, serialized/deserialized as needed.

File: api/src/wfl/module/covid.clj (hunk 6 above)
Comment (okotsopoulos) [reply api/src/wfl/module/covid.clj:162]: I'll change this back (thinking of COVID processing, which is really one workload that never finishes unless we stop it for a WDL change or similar!)

Comment (okotsopoulos) [reply api/src/wfl/module/covid.clj:162]: And re: validation (OG comment in thread), I'll file a ticket as validation in general could use further attention.

Specifically we do more than validate a request in this method, we also acquire resources as part of generating the stage object.  Maybe we will split out the validation and resource acquisition to be able to use validation as part of workload creation and retry.

Comment (okotsopoulos) [reply api/src/wfl/module/covid.clj]: Validation ticket: https://broadinstitute.atlassian.net/browse/GH-1421

File: api/src/wfl/api/handlers.clj (hunk 7 above)
Comment (okotsopoulos) [reply api/src/wfl/api/handlers.clj]: Ticketed separately: https://broadinstitute.atlassian.net/browse/GH-1424

File: api/src/wfl/executor.clj (hunk 8 above)
Comment (okotsopoulos) [reply api/src/wfl/executor.clj:472]: Comment added, and ticket: https://broadinstitute.atlassian.net/browse/GH-1423

File: api/src/wfl/executor.clj (hunk 9 above)
Comment (okotsopoulos) [reply api/src/wfl/executor.clj]: For now we accept that we only support references.  Ticket to support generic entity submission: https://broadinstitute.atlassian.net/browse/GH-1422

File: api/src/wfl/executor.clj [hunk 11]
```diff
@@ -217,24 +232,27 @@
...
                           {:submission submission :executor executor})))
+        (log/debug (format "%s Writing %s workflow uuids and statuses for submission %s..."
+                           (log-prefix executor) (count workflows) submission))
         (->> workflows
              (map #(update % :workflowEntity :entityName))
              (sort-by :workflowEntity)
              (map zip-record (sort-by :entity records))
              (write-workflow-statuses (utc-now)))))))
 
+(def ^:private final-statuses (util/to-quoted-comma-separated-list rawls/final-statuses))
```
okotsopoulos, line 243: Calling this out, I used the Rawls final statuses instead of the Cromwell statuses since we're talking to Rawls to get our workflows.  They do match, but the active statuses differ between Rawls and Cromwell.

api/src/wfl/api/handlers.clj
```diff
@@ -102,9 +103,17 @@
...
+      (throw (UserException. "Retry unsupported for workflow status."
+                             {:uuid            uuid
+                              :workflow-status status
+                              :status          400})))
     (->> (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
-           (let [w (workloads/load-workload-for-uuid tx uuid)]
-             [w (workloads/workflows-by-status tx w status)]))
+           (let [workload  (workloads/load-workload-for-uuid tx uuid)
+                 workflows (workloads/workflows-by-status tx workload status)]
+             [workload workflows]))
```
okotsopoulos, line 116: I wanted to throw here if we couldn't find any workflows for the status, but I had some trouble adjusting the system tests accordingly for pipelines without retry implemented.  So I settled for a check on this in the COVID implementation.

api/test/wfl/integration/executor_test.clj
```diff
@@ -104,80 +104,114 @@
   {:referenceId snapshot-reference-id
    :name        snapshot-reference-name})
 
-(def ^:private method-config-version-mock 1)
+;; Method configuration
```
okotsopoulos, line 107: And thus begins mock salad…

It's no fun to work with all of these helpers and I've only added to the complexity.  We're trying to mock out external service calls, of which there are many.  I do welcome feedback on how to streamline to make this space easier to work in in the future.

api/src/wfl/executor.clj
```diff
@@ -428,6 +446,88 @@
...
+         (map #(rename-keys % {:entity :retryEntity}))
+         (sort-by :retryEntity)
+         (map link-retry-to-original (sort-by :entity original-records))
+         (write-retries (utc-now)))))
+
+(defn ^:private retry-terra-executor
+  "Resubmit the snapshot references associated with `workflows` in `workspace`
+  and update each original workflow record with the row ID of its retry."
+  [{:keys [workspace] :as executor} workflows]
+  (letfn [(retried-workflow-records [reference-id]
```
ehigham, line 518: `submit-reference`?

api/src/wfl/module/covid.clj
```diff
@@ -152,6 +152,24 @@
...
+(defn ^:private retry-covid-workload
+  "Retry/resubmit the `workflows` managed by the `workload` and return the
+   workload that manages the new workflows."
+  [{:keys [started id executor] :as workload} workflows]
+  (when-not started
+    (throw (UserException. "Cannot retry workload before it's been started."
+                           {:workload workload})))
+  (when (empty? workflows)
+    (throw (UserException. "No workflows specified to retry in workload."
+                           {:workload workload})))
```
ehigham, line 164: If you're going to throw, throw sooner upstream - that way you can make the error message more meaningful, i.e.
```clojure
(throw (UserException. "Retry failed: no workflows found matching status" {:status status}))
```
(this will cause another issue though in that the `UserException` handler will try to convert this `:status` into an HTTP status code, so need to change that to test if it has a numeric `:status`, or change it to test for something else like `:http-status` or `:code` or `:status-code` or something)

api/test/wfl/integration/executor_test.clj
```diff
@@ -267,12 +310,83 @@
          (is (== 1 (stage/queue-length executor)))
          (is (not (stage/done? executor)))))))
 
+(deftest test-retry-terra-executor
```
ehigham, line 313: my goodness... I'll need some time for this!

ehigham (approved):
Pretty much all much all my feedback is subjective. This looks good - thank you.

api/src/wfl/module/covid.clj
ehigham: Also, the fact that a workload implementation doesn't support retries should be secondary - you've already implied this by checking for `Succeeded`

api/test/wfl/integration/executor_test.clj
okotsopoulos: Some time and a :beer: :)

api/src/wfl/api/handlers.clj
```diff
@@ -102,9 +103,17 @@
   (log/info (select-keys request [:request-method :uri :parameters]))
   (let [uuid   (get-in request [:path-params :uuid])
         status (get-in request [:body-params :status])]
+    ;; TODO: Refactor to workflow-status whitelist
+    ;; https://broadinstitute.atlassian.net/browse/GH-1424
+    (when (= "Succeeded" status)
```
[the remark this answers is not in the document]
okotsopoulos: Excellent, I was actually playing with this locally as I waited for further review!

api/src/wfl/executor.clj
```diff
@@ -142,12 +149,15 @@
...
-           methodConfigurationVersion] :as _executor}
+           methodConfigurationVersion] :as executor}
    {:keys [name]                       :as _reference}]
   (let [current (firecloud/method-configuration workspace methodConfiguration)
         _       (throw-on-method-config-version-mismatch
                  current methodConfigurationVersion)
         inc'd   (inc methodConfigurationVersion)]
+    (-> "%s Updating %s in %s with {:dataReferenceName %s :methodConfigVersion %s}"
+        (format (log-prefix executor) methodConfiguration workspace name inc'd)
+        log/debug)
```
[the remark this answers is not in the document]
okotsopoulos: I'll be happy to move away from log-prefix.  That was a "hack" predating the new logging.  I will look a bit more closely at passing the map, and ping Rich tomorrow if I get stuck.

api/src/wfl/executor.clj
```diff
@@ -428,6 +446,88 @@
      executor
      (jdbc/query tx [(format query details) status]))))
 
+(defn ^:private reference-id-to-workflow-records
```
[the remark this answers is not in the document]
okotsopoulos: I do think this is worth it because the name was fuzzy for even me to grasp, and I'm the author.  You're right that I named based on the return type.

I'll look into this, may be a scenario where breaking out into an external method reduces clarity rather than improving it.

api/src/wfl/executor.clj
```diff
@@ -428,6 +446,88 @@
...
+                       :executor         executor})))
+    (log/debug (format "%s Linking %s retry records to their originals..."
+                       (log-prefix executor) (count retry-records)))
+    (->> retry-records
+         (map #(rename-keys % {:entity :retryEntity}))
+         (sort-by :retryEntity)
+         (map link-retry-to-original (sort-by :entity original-records))
+         (write-retries (utc-now)))))
+
+(defn ^:private retry-terra-executor
```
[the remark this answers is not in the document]
okotsopoulos: I was following `update-terra-executor` naming but I think your suggestion is better.

api/src/wfl/executor.clj
okotsopoulos: This one's interesting, another case where I named based on return type.  In my development I didn't realize initially that `allocate-submission` returned the newly allocated records, but I've since updated its docstring.

I think your suggestion works in context of the threading below, since it is passed to `update-retried-workflow-records`.

api/src/wfl/executor.clj
```diff
@@ -217,24 +232,27 @@
...
-  (letfn [(read-active-or-failed-workflows []
+  "Update statuses in `details` table for active `workspace` workflows."
+  [{:keys [workspace details] :as executor}]
+  (letfn [(read-active-workflows []
             (let [query "SELECT * FROM %s
                          WHERE submission IS NOT NULL
                          AND   workflow   IS NOT NULL
-                         AND   status     NOT IN ('Succeeded', 'Aborted')
+                         AND   status     NOT IN %s
                          ORDER BY id ASC"]
```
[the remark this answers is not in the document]
okotsopoulos: Ahh, I understand.  Thanks for the example.

api/src/wfl/module/covid.clj
okotsopoulos: (To make sure I understand) "sooner upstream" should be within `handlers/post-retry`.

That's a good point that by checking if the status is `Succeeded` I've already shown that checking whether retries are implemented isn't a check that needs to happen first.  I agree that a more meaningful error message outweighs the changes to system tests.

(I did get burned by the `:status` naming collision earlier, thanks for flagging it.)

api/test/wfl/integration/executor_test.clj
okotsopoulos: (That being said, if it would help to walk through verbally what I'm testing LMK.)

api/test/wfl/integration/executor_test.clj
rexwangcc: I faced the same problem with testing the Slack notification, and really mocking out those function adds a huge amount of complexity and sometimes doesn't really add too much value in terms of testing. I had to refactor a lot of functions so mock is minimally required. But in this case of talking to Firecloud/Rawls, I cannot think of a better way yet.

api/test/wfl/integration/executor_test.clj
rexwangcc: WOW, I need a ☕  before going through this …

rexwangcc (approved):
So far LGTM! Still looking through the last bit of testing part.

api/src/wfl/module/covid.clj
ehigham: Yeah exactly - in `post-retry` :)

api/src/wfl/executor.clj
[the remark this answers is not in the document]
okotsopoulos: I didn't get to this, but would like to address in a separate PR since all other feedback has been incorporated.

api/src/wfl/executor.clj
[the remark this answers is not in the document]
okotsopoulos: I refactored as suggested but renamed thusly:

```
(defn ^:private workflow-and-sibling-records
  "Return the workflow records for all workflows in submissions
  associated with the specified `workflows` --
  i.e. records for `workflows` and their siblings."
  …
```

Because `workflows-to-workflow-records` implies that you're getting the workflow records for the specified workflows, and this is a funky case.

api/src/wfl/api/handlers.clj
```diff
@@ -96,15 +98,34 @@
            (mapv util/to-edn)
            succeed))))
 
+;; Visible for wfl.system.v1-endpoint-test/test-retry-workload
```
ehigham, line 101: Don't bother mentioning the test as other tests might use this too!

ehigham (approved):

PR #483, opened 2021-07-30
GH-1395: Document the new WFL retry capability for users.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1395

### Changes

- Document that /retry in not implemented.

### Review Instructions

- No instructions.

### DIFF
--- a/docs/md/usage-retry.md
+++ b/docs/md/usage-retry.md
@@ -20,6 +20,8 @@ curl -X POST -H "$AUTH" $WFL/$UUID/retry --data '"Failed"' | jq
 
 The `/retry` endpoint is not yet implemented.
 and returns a `501` HTTP failure status.
+Until it is supported,
+see the method below.
 
 # Retrying Failures via WFL
 
--- a/docs/md/usage-retry.md
+++ b/docs/md/usage-retry.md
@@ -15,7 +15,7 @@ export WFL=https://gotc-prod-wfl.gotc-prod.broadinstitute.org/api/v1/workload
 export AUTH="Authorization: Bearer $(gcloud auth print-access-token)"
 export UUID=0d307eb3-2b8e-419c-b687-8c08c84e2a0c # workload UUID
 
-curl -X POST -H "$AUTH" $WFL/$UUID/retry --data '"Failed"' | jq
+curl -X POST -H "$AUTH" $WFL/$UUID/retry --data '{"status":"Failed"}' | jq
 ```
 
 The `/retry` endpoint is not yet implemented.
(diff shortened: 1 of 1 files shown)


rexwangcc (approved):
Thanks

docs/md/usage-retry.md
```diff
@@ -1,3 +1,28 @@
...
+
+The following `curl` shell command
+finds the workflows with the status `"Failed"`
+in the workload with `$UUID`
+and resubmits them for processing.
+
+```bash
+export WFL=https://gotc-prod-wfl.gotc-prod.broadinstitute.org/api/v1/workload
+export AUTH="Authorization: Bearer $(gcloud auth print-access-token)"
+export UUID=0d307eb3-2b8e-419c-b687-8c08c84e2a0c # workload UUID
```
ehigham, line 16: Why `export`?

docs/md/usage-retry.md
```diff
@@ -1,3 +1,28 @@
...
+
+curl -X POST -H "$AUTH" $WFL/$UUID/retry --data '{"status":"Failed"}' | jq
+```
+
+The `/retry` endpoint is not yet implemented.
+and returns a `501` HTTP failure status.
+Until it is supported,
+see the method below.
```
ehigham, line 24: When the code is released, it will. Could you write this as if it is correct at the time of release to the best of our knowledge?

ehigham (commented):
What about limitations? Are all statuses supported? What are these statuses or where can a user find them? What's the expected response from retry when it succeeds? When it fails?

docs/md/usage-retry.md
tbl3rd: OK. I'll give it a shot.

docs/md/usage-retry.md
```diff
@@ -1,3 +1,52 @@
...
+[Cromwell statuses](https://github.com/broadinstitute/wfl/blob/c30e77450926d25b085696b08a306775cd244ea1/api/src/wfl/service/cromwell.clj#L14)
+to drive the `/retry` API are:
+- `"Aborted"`
+- `"Failed"`
+
+`"Aborted"` and `"Failed"` are terminal,
+and likely failure,
+statuses for workflows.
+
+Specifying other Cromwell statuses
```
okotsopoulos, line 32: Howdy, my [implementation PR](https://github.com/broadinstitute/wfl/pull/482) is ready for review.  What I want to note is that when we retry a workflow, we have to resubmit all workflows from the same submission since we can't submit subsets of snapshots yet. So if you're retrying a failed workflow in a still-running submission, note that all workflows in the submission including those still running will be retried.

PR #484, opened 2021-08-02
[GH-1425] Expose `TerraDataRepoSink` Via the HTTP API

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1425
Add specs and register in `wfl.api/spec`.
Add end-to-end system test for reading/writing workflow inputs/outputs to TDR (used the `illumina_genotyping_array` pipeline as it's smaller than `sarscov2_illumina_full`).
- Note this doesn't quite work yet until Jade finish their hybrid ingest work


api/src/wfl/service/datarepo.clj
```diff
@@ -95,15 +94,14 @@
 (defn poll-job
   "Poll the job with `job-id` every `seconds` [default: 5] and return its
    result."
```
rexwangcc, line 96:
```suggestion
  "Poll the job with `job-id` every `seconds` [default: 5] and return its
   result with `max-attempts` [default: 20]."
```

api/test/wfl/tools/endpoints.clj
```diff
@@ -107,3 +112,28 @@
...
+          {:data {:coercion   reitit.coercion.spec/coercion
+                  :middleware [coercion/coerce-exceptions-middleware
+                               coercion/coerce-request-middleware
+                               coercion/coerce-response-middleware]}}))]
+    (fn [request]
+      (let [{:keys [status body] :as _response}
+            (app {:request-method :post
+                  :uri            "/test"
+                  :body-params    request})]
+        (is (== 200 status) (pr-str body))))))
```
rexwangcc, line 139: Nice!

rexwangcc (commented):
Looks great to me, thanks!

api/src/wfl/sink.clj
```diff
@@ -352,3 +371,10 @@
 (defoverload stage/done?  datarepo-sink-type datarepo-sink-done?)
 
 (defoverload util/to-edn datarepo-sink-type datarepo-sink-to-edn)
+
+;; reitit http coercion specs for a sink
+;; Recall s/or doesn't work (https://github.com/metosin/reitit/issues/494)
+(s/def ::sink
+  #(condp = (:name %)
+     terra-workspace-sink-name (s/valid? ::terra-workspace-sink %)
+     datarepo-sink-name        (s/valid? ::terra-datarepo-sink %)))
```
ehigham, line 380: Sadly when this fails we don't get very useful error messages as it won't tell you what's wrong with your request. All it'll say is your request doesn't match a workspace sink or a datarepo sink request. That said, the errors we return currently are truly dreadful.
I created https://broadinstitute.atlassian.net/browse/GH-1355 a while ago to address the shortcomings of these coercion errors.

api/src/wfl/service/datarepo.clj
```diff
@@ -94,16 +93,15 @@
...
-     (while (running?) (.sleep TimeUnit/SECONDS seconds))
-     (result)))
-  ([job-id]
-   (poll-job job-id 5)))
+   result with `max-attempts` [default: 20]."
+  [job-id & [seconds max-attempts]]
+  (let [result   #(get-repository-json "jobs" job-id "result")
+        done? #(-> (get-repository-json "jobs" job-id)
+                   :job_status
+                   #{"running"}
```
okotsopoulos, line 101: Are there intermediate states other than "running"?

ehigham: No - It's running, succeeded or failed. I'm checking that it's not running, just `poll` returns when the value is non-nil.

ehigham: I'm unsure what's better, checking if it's running then inverting or checking if it's either succeeded or failed.

okotsopoulos: My instinct is that it's more likely that intermediate active states are added vs. more final states added.  It might be good to pull these known states out like how we do for [Cromwell](https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/service/cromwell.clj#L12-L18) and [Rawls](https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/service/rawls.clj#L12-L25), then check against the final state list.  Will make future debugging easier if the states change upstream.

ehigham: Good idea. I'll do that.

api/src/wfl/sink.clj
```diff
@@ -280,17 +279,19 @@
   (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
     (jdbc/update! tx details {:consumed (utc-now)} ["id = ?" id])))
 
+(def ^:private active-job-query
+  (format
+   "SELECT id, job, workflow FROM %%s WHERE status IN %s ORDER BY id ASC"
```
okotsopoulos, line 284: Just double-checking on the "%%s", is that intended?

ehigham: Yes - that's to escape the table name in the first format call. I can do `str/replace` if you prefer?

okotsopoulos: No need, all good!

okotsopoulos (approved):

PR #487, opened 2021-08-09
[GH-1418] sourceLocation null fix

rfricke-asymmetrik:
### Purpose
sourceLocation prints out as null for every log entry. This is the result of clojure dropping the &form fields in nested macros.

https://broadinstitute.atlassian.net/browse/GH-1418

### Changes

Update severity specific macros to pass sourceLocation as an additional field to the log macro so that line and file can be included.

### Review Instructions
- All of the logs that print to stdout should include a line number and file.


api/src/wfl/log.clj
```diff
@@ -89,29 +89,34 @@
...
+  (let [{:keys [line]} (meta &form)]
+    `(log :debug ~expression :logging.googleapis.com/sourceLocation {:file ~*file* :line ~line})))
 
 (defmacro info
   "Log `expression` as information.
    Used for routine information, such as ongoing status or performance."
   [expression]
-  `(log :info ~expression))
+  (let [{:keys [line]} (meta &form)]
+    `(log :info ~expression :logging.googleapis.com/sourceLocation {:file ~*file* :line ~line})))
```
ehigham, line 100:
```suggestion
  (log :info expression))
```

tbl3rd (approved):
Thank you for figuring this out.

PR #490, opened 2021-08-10
[GH-1430] Remove `stage/validate-or-throw` and inline their implementations into `create-X`.

ehigham:
RR: https://broadinstitute.atlassian.net/browse/GH-1430
Remove `stage/validate-or-throw` and inline their implementations in their respective `create-X` functions.

This is done because
- we were losing context (ie. was this a source/sink/executor) and had a collision in the tdr sink and source impls
- these multimethods didn't really make sense and require some redesign work

### DIFF
--- a/api/src/wfl/executor.clj
+++ b/api/src/wfl/executor.clj
@@ -54,6 +54,15 @@
    association in the `workload`."
   (fn [_transaction workload] (:executor_type workload)))
 
+(defmethod create-executor! :default
+  [_ _ {:keys [name] :as request}]
+  (throw (UserException.
+          "Invalid executor name"
+          {:name      name
+           :request   request
+           :status    400
+           :executors (-> create-executor! methods (dissoc :default) keys)})))
+
 ;; Terra Executor
 (def ^:private terra-executor-name  "Terra")
 (def ^:private terra-executor-type  "TerraExecutor")
--- a/api/src/wfl/executor.clj
+++ b/api/src/wfl/executor.clj
@@ -14,16 +14,6 @@
   (:import [wfl.util UserException]))
 
 ;; executor operations
-(defmulti validate-or-throw
-  "Return a validated executor request (i.e. resources exist and wfl can access
-   them) or throw."
-  :name)
-
-(defmethod validate-or-throw :default
-  [{:keys [name] :as request}]
-  (throw (UserException. "No such executor"
-                         (util/make-map name request))))
-
 (defmulti update-executor!
   "Consume items from the `upstream-queue` and enqueue to the `executor` queue
    for consumption by a later processing stage, performing any external effects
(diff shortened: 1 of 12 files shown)


api/src/wfl/executor.clj
```diff
@@ -100,7 +100,7 @@
     (throw (UserException. "Only Dockstore methods are supported."
                            {:status 400 :methodRepoMethod methodRepoMethod}))))
 
-(defn verify-terra-executor
+(defn validate-terra-executor-request
```
okotsopoulos, line 103: I like this naming better.

api/src/wfl/sink.clj
```diff
@@ -20,7 +20,6 @@
   (:import [clojure.lang ExceptionInfo]
            [wfl.util UserException]))
 
-;; Interface
```
okotsopoulos, line 23: Just curious why this moved?

api/src/wfl/sink.clj
```diff
@@ -102,7 +102,7 @@
   (str/join " " ["Found additional attributes in fromOutputs that are not"
                  "present in the entityType."]))
 
-(defn ^:private terra-workspace-validate-request-or-throw
+(defn terra-workspace-validate-request-or-throw
```
okotsopoulos, line 105: Huh, I'm surprised that this is the only validation taking place on a sink request.

api/src/wfl/executor.clj
okotsopoulos: Actually, I wouldn't mind consistent naming of all the validate request methods, here's what they look like right now:

- validate-terra-executor-request
- terra-workspace-validate-request-or-throw
- datarepo-sink-validate-request-or-throw
- validate-datarepo-source!
- validate-tdr-snapshot-list

If I were to choose one naming convention, I'd pick `<type>-validate-request-or-throw` or similar.

api/src/wfl/sink.clj
```diff
@@ -187,17 +187,19 @@
...
-(defoverload update-sink!  terra-workspace-sink-type update-terra-workspace-sink)
-(defoverload stage/done?   terra-workspace-sink-type terra-workspace-sink-done?)
+(defoverload load-sink!   terra-workspace-sink-type load-terra-workspace-sink)
+(defoverload update-sink! terra-workspace-sink-type update-terra-workspace-sink)
+(defoverload stage/done?  terra-workspace-sink-type terra-workspace-sink-done?)
 
 (defoverload util/to-edn terra-workspace-sink-type terra-workspace-sink-to-edn)
 
 ;; TerraDataRepo Sink
-(def ^:private ^:const datarepo-sink-name  "Terra DataRepo Sink")
```
okotsopoulos, line 200: Yay!

api/src/wfl/executor.clj
ehigham: sure - sounds like a plan

rexwangcc: +1

rexwangcc (approved):
LGTM!

PR #495, opened 2021-08-19
Github actions should build docs on develop.

rexwangcc:
### Purpose

### Changes

- Github actions should build docs on develop.

### Review Instructions

- I looked into [how `mkdocs` uses `mike` as a versioning manager](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?h=version), but it introduces too much complication into our build workflow, so maybe we should just deploy docsite on pushing to `develop`.


tbl3rd (approved):
Awesome!

PR #497, opened 2021-08-24
[GH-1353] Fix spec error when skipValidation sent

rfricke-asymmetrik:
### Purpose
Fix the spec error that occurs when skipValidation is sent in the source of a create workload request.

https://broadinstitute.atlassian.net/browse/GH-1353

### Changes

- Changed update to assoc and moved to before select-keys
- Added code to write dataset into id when skipValidation is provided

### Review Instructions
- No instructions.


tbl3rd (approved):
This fixes the spec error.

PR #502, opened 2021-09-03
GH-1446: fix system tests again

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1446

### Changes

- Extended and renamed the `wfl.service.datarepo/bigquery-name` hack to cover more cases.
- Renamed `wfl.service.datarepo/dataset` to `wfl.service.datarepo/datasets` for consistency.
- Added `wfl.service.google.bigquery/bigquery-v2-rest-api` for convenience.
- `wfl.util/parse-json` should no longer fail when it gets a string that doesn't parse as JSON.
- `wfl.system.v1-endpoint-test/wfl.system.v1-endpoint-test` should be less flaky now.
- `wfl.system.v1-endpoint-test/test-workload-sink-outputs-to-tdr` should work now.
- Bumped the `methodConfigurationVersion` in the COVID workload to conform to the workspace.
- Fixed the `wfl.tools.fixtures/with-temporary-dataset` fixture so it works with snapshots.
- Fixed `wfl.tools.workloads/when-all-workflows-finish` to extract status correctly.

### Review Instructions

- 2 of the integration tests will fail until `general-dev-billing-account` works again.
- The docs have only whitespace changes from an aborted `to-edn` -=> `to-user-edn` renaming.
- Similarly, `log.clj` has only whitespace cleaned up by the editor.
- The rest is clarification and fallout from other changes.

### DIFF
--- a/api/src/wfl/api/routes.clj
+++ b/api/src/wfl/api/routes.clj
@@ -21,8 +21,7 @@
             [wfl.util                           :as util]
             [wfl.wfl                            :as wfl])
   (:import [java.sql SQLException]
-           [wfl.util UserException]
-           [org.apache.commons.lang3.exception ExceptionUtils]))
+           [wfl.util UserException]))
 
 (def endpoints
   "Endpoints exported by the server."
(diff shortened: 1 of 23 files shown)


okotsopoulos:
(Sorry if you are aware already but) looks like this needs another lint.

tbl3rd:
Was not aware. Thanks.

rfricke-asymmetrik (approved):

PR #505, opened 2021-09-10
GH-1450 Add retry attribute to TerraDataRepoSource type and instances

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1450

In preparation for automatic retries of failed snapshot creation jobs within `TerraDataRepoSource`'s update loop, added a retry column to the `TerraDataRepoSource` type and its instances.

I looked at all existing queries of this source details table (details table peek and queue length).  I found nowhere that needed an additional filter to exclude retried snapshot creation jobs, since we will only be automating retries for failed (and as such, never-to-be-consumed) jobs.

*However*, it might be worth codifying this filter if we think we might ever support API-driven retry of snapshot creation jobs, specifically successful ones.  LMK if you'd like me to make this change.

### Review Instructions

Prior to this change, I executed a COVID workload with a `TerraDataRepoSource`.
After making the change and running the Liquibase update, I confirmed in my local Postgres that the existing source details table had been updated with a retry column.

```
api $ clojure -M:liquibase
```

Screen Shot 2021-09-10 at 3 48 43 PM

### DIFF
--- a/database/changelog.xml
+++ b/database/changelog.xml
@@ -43,4 +43,5 @@
     <include file="changesets/20210709_TerraDataRepoSink.xml"                         relativeToChangelogFile="true"/>
     <include file="changesets/20210720_AddConfigurationTable.xml"                     relativeToChangelogFile="true"/>
     <include file="changesets/20210823_ConvertWatchersArrayToText.xml"                relativeToChangelogFile="true"/>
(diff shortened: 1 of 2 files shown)


okotsopoulos:
Test failures addressed in separate PR:

https://github.com/broadinstitute/wfl/pull/504

rfricke-asymmetrik (approved):

rhiananthony (approved):
👍🏼

PR #514, opened 2021-09-22
GH-1452: Bump AoU Arrays.wdl version to 2.4.1 to drop default genotype_concordance_threshold to 0.95.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1452

### Changes

- Just did what it says.

### Review Instructions

- No instructions.

### DIFF
--- a/api/test/wfl/tools/workloads.clj
+++ b/api/test/wfl/tools/workloads.clj
@@ -93,7 +93,7 @@
    :sample_lsid                     "broadinstitute.org:bsp.dev.sample:NOTREAL.03C17319",
    :analysis_version_number         1,
    :call_rate_threshold             0.98,
-   :genotype_concordance_threshold  0.98,
+   :genotype_concordance_threshold  0.95,
    :reported_gender                 "Male",
    :chip_well_barcode               "200598830050_R07C01",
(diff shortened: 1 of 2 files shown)


api/test/wfl/tools/workloads.clj
```diff
@@ -93,7 +93,7 @@
    :sample_lsid                     "broadinstitute.org:bsp.dev.sample:NOTREAL.03C17319",
    :analysis_version_number         1,
    :call_rate_threshold             0.98,
-   :genotype_concordance_threshold  0.98,
+   :genotype_concordance_threshold  0.95,
```
okotsopoulos, line 96: (For my own understanding) this `arrays-sample` is just got give an example of what a real sample would look like for use in tests, and the threshold change is encompassed in the bump to WDL version.

okotsopoulos (approved):

api/test/wfl/tools/workloads.clj
tbl3rd: Yes. As stated in the ticket, this isn't strictly necessary, but without it someone will have to decide whether the discrepancy matters. I kinda did this to preclude questions like yours. ;-)

PR #520, opened 2021-10-04
[GH-1474-1494] TDR Sink Changes

rfricke-asymmetrik:
### Purpose
The TerraDataRepoSink was waiting for changes to TDR during implementation. Because of this, it was not tested fully.

- https://broadinstitute.atlassian.net/browse/GH-1474
- https://broadinstitute.atlassian.net/browse/GH-1494
- https://broadinstitute.atlassian.net/browse/GH-1475

### Changes
- System test for TDR Sink needed to check for a seq on rows instead of whole json object
- In order for TDR to be able to access files generated in Terra, the service account for TDR needs to have shared access to the Terra workspace.
- The single step process for ingesting into TDR requires a BulkFileLoadModel be sent instead of a string.
- Integration test added for new Terra -> TDR translation for sink.
- The Sink can return a result of `succeeded` but still have failed rows. A UserException is now thrown if there is at least one `bad_row_count`.
- Added an environment variable for the tdr outputs bucket and modified the code in `start-ingesting-outputs` to use said variable. There will be a separate bucket for each environment.

### Review Instructions
- System test `test-workload-sink-outputs-to-tdr` passes
- To do a manual test, create a terra workspace with the TDR service account shared as owner. Execute a workload with TerraDataRepo as the sink with the desired outputs. Create your source, whether via Terra or TDR (this would require an inputs dataset and an ingest into that dataset like in the system test). Once workload completes. Check that the output dataset table in TDR has a new row.
- Verify that a file was uploaded to `broad-gotc-dev-wfl-tdr-outputs` bucket.


api/src/wfl/sink.clj
```diff
@@ -317,27 +325,63 @@
...
+                                      :sourcePath val
+                                      :targetPath (get-target val)}])
+                   (literal? v) [k (subs v 1 (count v))]
+                   (boolean? v) [k (values (keyword v))]
+                   (string?  v) [k (values (keyword v))]
+                   (map?     v) [k (rename-gather-bulk workflow-id dataset table values v)]
+                   (coll?    v) [k (keep (go! k v))]
+                   :else        (throw (ex-info "Unknown operation"
+                                                {:operation v}))))]
+     (into {} (for [[k v] mapping] (go! k v))))))
```
tbl3rd, line 364: Why doesn't `rename-gather` have `boolean` and `fileref` clauses?
Do we still need `rename-gather`?  Looks like it.
(-I'm so confused!-)

api/src/wfl/service/firecloud.clj
```diff
@@ -5,7 +5,8 @@
             [clojure.string    :as str]
             [wfl.auth          :as auth]
             [wfl.environment   :as env]
-            [wfl.util          :as util])
+            [wfl.util          :as util]
+            [clojure.java.io    :as io])
```
tbl3rd, line 9: This may be cruft.

tbl3rd (commented):
Checkpoint because I don't remember where I left off, sorry.

api/src/wfl/sink.clj
rfricke-asymmetrik: The `rename-gather` function is used for Terra Workspace Sinks, which it only sends the fileref fields as strings. In this case, for TerraDataRepoSink's single step ingest, we need to send a json blob with the sourcePath, targetPath, mimeType and a description. Because of this, there is special handling to turn the string based fileref outputs to this BulkFileLoadModel that TDR expects.

api/src/wfl/sink.clj
```diff
@@ -317,27 +325,63 @@
...
   (when-not (map? fromOutputs)
     (throw (IllegalStateException. "fromOutputs is malformed")))
   (try
-    (rename-gather outputs fromOutputs)
+    (rename-gather-bulk uuid dataset table outputs fromOutputs)
     (catch Exception cause
       (throw (ex-info "Failed to coerce workflow outputs to dataset columns"
-                      {:fromOutputs fromOutputs :workflow workflow}
+                      {:outputs outputs :table table :fromOutputs fromOutputs :workflow workflow :dataset dataset}
                       cause)))))
```
tbl3rd, line 376: Perhaps just dump the sink itself here?
What are we leaving out of the `ex-data`?
Is that intentional?

tbl3rd (commented):
Apologies for the piecemeal commenting.

api/src/wfl/sink.clj
tbl3rd: But shouldn't `rename-gather` be a relational subset of `rename-gather-bulk`?
If not, maybe we should rename them for `workspace` vs `tdr` sinks?

tbl3rd: Must `target-bucket` name a bucket?
It looks like a generic GCS prefix or "folder".

tbl3rd: And now I'm having trouble seeing how that can be used in practice.
What sets target-bucket? I see it used in a test.
Is it intended for production?

rfricke-asymmetrik: It was admittedly added so that I could save the files to a specific bucket in the test so that the files could be cleaned up by the fixture afterwards. But could serve a purpose in production in the future if there is a preference on where the target files are saved during ingest.

tbl3rd (approved):
I appreciate your answers.

PR #528, opened 2021-10-28
[GH-1517] Add automated test for updating method configurations with mismatched versions

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1517

I added an integration test to codify the manual verification I performed for https://github.com/broadinstitute/wfl/pull/527

(Resharing for context) When attempting to submit a snapshot, if the method config's version doesn't match what we expect, we log an error rather than throw an exception. This allows submissions to move forward without Hornet developer intervention. In the edge case where a method configuration change caused workflow(s) to fail, we will have the record of the error log and Stackdriver alert to aide in debugging.

### DIFF
--- a/api/src/wfl/executor.clj
+++ b/api/src/wfl/executor.clj
@@ -98,14 +98,19 @@
         (update :fromSource edn/read-string))
     (throw (ex-info "Invalid executor_items" {:workload workload}))))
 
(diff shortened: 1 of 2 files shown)


rfricke-asymmetrik (approved):
👍

PR #538, opened 2021-11-15
Update test inputs buckets

rfricke-asymmetrik:
### Purpose

In the TDR Sink PR, there were a couple of changes to which buckets some of the test inputs come from. These changes were not correct, as the bucket that was being used has a 14 day lifecycle. This PR reverts those bucket changes.

Related PR: https://github.com/broadinstitute/gotc-deploy/pull/309
Related Ticket: https://broadinstitute.atlassian.net/browse/GH-1475

In making these changes, it was found that there was an integration test that was giving a false negative result. The test was checking that a UserException was thrown when there would not be because the ingest succeeds. This PR contains the rework of the test to correctly check for UserException on an actual failing ingest.

### Changes

- Restore the buckets used to ingest inputs into TDR Sink test datasets
- Rework `update-datarepo-sink` integration test

### Review Instructions

- No instructions.

### DIFF
--- a/api/test/resources/workflows/illumina_genotyping_array/inputs.json
+++ b/api/test/resources/workflows/illumina_genotyping_array/inputs.json
@@ -1,8 +1,8 @@
 {
   "analysis_version_number": 1,
   "chip_well_barcode": "7991775143_R01C01",
-  "green_idat_cloud_path": "gs://broad-gotc-dev-wfl-datarepo-outputs/test-inputs/illumina-genotyping-array_inputs_7991775143_R01C01_Grn.idat",
-  "red_idat_cloud_path": "gs://broad-gotc-dev-wfl-datarepo-outputs/test-inputs/illumina-genotyping-array_inputs_7991775143_R01C01_Red.idat",
+  "green_idat_cloud_path": "gs://broad-gotc-dev-wfl-ptc-test-inputs/illumina-genotyping-array/inputs/7991775143_R01C01_Grn.idat",
+  "red_idat_cloud_path": "gs://broad-gotc-dev-wfl-ptc-test-inputs/illumina-genotyping-array/inputs/7991775143_R01C01_Red.idat",
   "sample_alias": "NA12878",
   "reported_gender": "Female"
 }
--- a/api/test/wfl/integration/sinks/datarepo_sink_test.clj
+++ b/api/test/wfl/integration/sinks/datarepo_sink_test.clj
@@ -133,15 +133,15 @@
   "Update sink until there is a truthy result or the max number of attempts is reached."
   [executor sink seconds max-attempts]
   (letfn [(update-and-check []
-                            (sink/update-sink! executor sink)
-                            (stage/done? sink))]
+            (sink/update-sink! executor sink)
+            (stage/done? sink))]
     (loop [attempt 1]
-     (if-let [result (update-and-check)]
-       result
-       (if (<= max-attempts attempt)
-         false
-         (do (.sleep TimeUnit/SECONDS seconds)
-             (recur (inc attempt))))))))
+      (if-let [result (update-and-check)]
+        result
+        (if (<= max-attempts attempt)
+          false
+          (do (.sleep TimeUnit/SECONDS seconds)
+              (recur (inc attempt))))))))
 
 (deftest test-update-datarepo-sink
   (let [description      (resources/read-resource "primitive.edn")
--- a/api/test/wfl/integration/sinks/datarepo_sink_test.clj
+++ b/api/test/wfl/integration/sinks/datarepo_sink_test.clj
@@ -11,10 +11,10 @@
             [wfl.tools.queues     :refer [make-queue-from-list]]
             [wfl.tools.resources  :as resources]
             [wfl.tools.workloads  :refer [evalT]]
-            [wfl.util             :as util]
-            [wfl.log              :as log])
+            [wfl.util             :as util])
   (:import [java.util UUID]
-           [wfl.util UserException]))
+           [wfl.util UserException]
+           [java.util.concurrent TimeUnit]))
 
 (def ^:private ^:dynamic *dataset*)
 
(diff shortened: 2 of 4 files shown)


api/test/resources/workflows/illumina_genotyping_array/inputs.json
```diff
@@ -1,8 +1,8 @@
 {
   "analysis_version_number": 1,
   "chip_well_barcode": "7991775143_R01C01",
-  "green_idat_cloud_path": "gs://broad-gotc-dev-wfl-datarepo-outputs/test-inputs/illumina-genotyping-array_inputs_7991775143_R01C01_Grn.idat",
-  "red_idat_cloud_path": "gs://broad-gotc-dev-wfl-datarepo-outputs/test-inputs/illumina-genotyping-array_inputs_7991775143_R01C01_Red.idat",
+  "green_idat_cloud_path": "gs://broad-gotc-dev-wfl-ptc-test-inputs/illumina-genotyping-array/inputs/7991775143_R01C01_Grn.idat",
```
okotsopoulos, line 4: Uninformed question, so the path generation differs based on the bucket?

okotsopoulos (approved):
In the description, would you please link to the TDR sink PR, the corresponding ticket(s), and the corresponding Terraform PR?  Thank you!

https://github.com/broadinstitute/gotc-deploy/pull/309/files

Otherwise, I think this looks good.

api/test/resources/workflows/illumina_genotyping_array/inputs.json
rfricke-asymmetrik: Correct. The path to the object in storage is `gs://` then the bucket name then the path inside the bucket to the object

api/test/wfl/integration/sinks/datarepo_sink_test.clj
```diff
@@ -119,13 +119,19 @@
...
+                                                            :outfile   "gs://not-a-real-bucket/external-reprocessing/exome/develop/not-a-real.unmapped.bam"
+                                                            :outfloat  (* 4 (Math/atan 1))
+                                                            :outint    27
+                                                            :outstring "Hello, World!"}}
+        failing-upstream (make-queue-from-list [[description failing-workflow]])
+        sink             (create-and-load-datarepo-sink)]
     (sink/update-sink! upstream sink)
     (is (stage/done? upstream))
-    (eventually (throws? UserException) #(sink/update-sink! upstream sink)
+    (is (or (sink/update-sink! upstream sink) true) "subsequent updates do nothing")
```
okotsopoulos, line 134: Do we want something that tests that eventually the good job succeeds?  That would be a nice addition here. (But maybe not a release blocker.)

okotsopoulos: A few more thoughts.

I think that "subsequent updates do nothing" does not adequately describe the behavior here, or what a test failure would mean.

Wouldn't `(or (sink/update-sink! upstream sink) true)` always return true?

okotsopoulos: (And, now I see that you didn't add this test initially.  Sorry to pull you into it!)

okotsopoulos: Apologies for the "thinking out loud" in this thread.

I think initially, the purpose of this check following the `(eventually <updating this sink yields a UserException>)` check is to confirm that once a job is identified as having failed, we won't continue to throw exceptions for it.  So in its current location it's not giving us anything useful.

rfricke-asymmetrik: I think you are correct. I will move it back down to under the error check and remove the or and see what happens.

okotsopoulos: The `or` might be there if `sink/update-sink!` is executed for side effects and returns `nil` in normal operations.  Again with the thinking out loud here :)

rfricke-asymmetrik: Sorry, yea. I realized that after my comment. I am leaving it. The real test for that line is whether it throws another exception and not the return of the function. I am also adding a check for the status of the job. Should have it up in a few minutes.

api/test/wfl/integration/sinks/datarepo_sink_test.clj
```diff
@@ -118,14 +120,33 @@
...
+
+(defn ^:private check-for-succeeded-jobs
+  "True when all tdr ingest jobs created by the `_sink` have terminated."
+  [{:keys [details] :as _sink}]
+  (let [query "SELECT COUNT(*) FROM %s
+               WHERE status = 'succeeded' AND consumed IS NOT NULL"]
+    (log/info "Checking for succeeded jobs")
+    (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+      (postgres/throw-unless-table-exists tx details)
+      (-> (jdbc/query tx (format query details)) first :count))))
```
tbl3rd, line 133: Name it `count-succeeded-injest-jobs`?

api/test/wfl/integration/sinks/datarepo_sink_test.clj
```diff
@@ -10,7 +11,8 @@
             [wfl.tools.queues     :refer [make-queue-from-list]]
             [wfl.tools.resources  :as resources]
             [wfl.tools.workloads  :refer [evalT]]
-            [wfl.util             :as util])
+            [wfl.util             :as util]
+            [wfl.log              :as log])
```
tbl3rd, line 15: Not important, but we usually sort these.

tbl3rd (approved):

api/test/wfl/integration/sinks/datarepo_sink_test.clj
```diff
@@ -118,14 +120,33 @@
...
+                                                            :outfile   "gs://not-a-real-bucket/external-reprocessing/exome/develop/not-a-real.unmapped.bam"
+                                                            :outfloat  (* 4 (Math/atan 1))
+                                                            :outint    27
+                                                            :outstring "Hello, World!"}}
+        failing-upstream (make-queue-from-list [[description failing-workflow]])
+        sink             (create-and-load-datarepo-sink)]
     (sink/update-sink! upstream sink)
     (is (stage/done? upstream))
-    (eventually (throws? UserException) #(sink/update-sink! upstream sink)
+    (is (= 1 (check-for-succeeded-jobs sink)) "job succeeded")
```
tbl3rd, line 148: Not important, but you could use `==` here.

PR #547, opened 2021-12-08
GH-1526: Bump the Arrays WDL release to 2.6.0

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1526

### Changes

- `:release "Arrays_v2.6.0"`

### Review Instructions

- No instructions.

### DIFF
--- a/api/src/wfl/module/aou.clj
+++ b/api/src/wfl/module/aou.clj
@@ -33,7 +33,7 @@
 
 (def workflow-wdl
   "The top-level WDL file and its version."
-  {:release "Arrays_v2.4.1"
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):
Looks good.  Are there system tests that run this pipeline?

Also will share point 2 in the [statement of work](https://docs.google.com/document/d/10WKOyLmT2BRIPWlu-sclMl91Qd6zH2okxm9LNErrcOg/edit) in case it highlights additional work for us:

>Then, Hornet needs to redeploy WFL. **They may also need to install the new extended manifest and cluster file in their metadata location.**

PR #553, opened 2022-01-14
GH-1542 Notify watchers on workflow completion

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1542
- Corresponding `gotc-deploy` PR to add new env vars: https://github.com/broadinstitute/gotc-deploy/pull/319

WFL staged workload power users (Christine, Sushma, and… me 😄) wanted Slack notifications when workflows completed, no matter the status.  Previously, Slack notifications were only emitted when a `UserException` was thrown.  But workflow completion isn't an exceptional condition which should block processing.

In this PR I call the Slack notifier directly when we've registered that an active workflow has finished.  The messages emitted reflect feedback from Christine and Sushma re: what info would help them, and include:
- Link to Terra Job History page for the workflow's submission
- Link to Job Manager UI for the workflow -- if a workflow failed, this link directs to the cause of failure
- Icons lining up with those in Terra UI

This leaves the door open to add additional state change notifications in the future (ex. submission started), but we may want to let users configure watchers separately for state change notifications (can be high volume) vs. exceptional conditions.

### Changes

- Call the Slack notifier when we've registered that an active workflow has finished
- Update Slacker to accept full workload object
- Added automated tests
- Update public-facing docs

### Manual Verification

Here's some scratch code I ran in `wfl.executor` to generate Slack notifications for real COVID workflows:
```
(let [workspace         "pathogen-genomic-surveillance/CDC_Viral_Sequencing_dev"
        executor          {:workspace workspace}
        uuid              "3c45595f-8794-4718-911c-f03fc66dfd52" ;; WFL prod COVID workload
        labels            ["project:GH-1543"]
        watchers          [["slack" "C026PTM4XPA"]]  ;; #hornet-slack-app-testing
        workload          {:uuid     uuid
                           :labels   labels
                           :watchers watchers
                           :executor executor}
        submission        "6a8d7467-6d40-4ce7-a3d3-0a62f196b959"
        workflowFailed    "45e2ffd6-d30e-477d-8168-ced7ad95e5de"
        recordFailed      {:submission submission
                           :workflow   workflowFailed
                           :status     "Failed"}
        workflowSucceeded "cbac4c3d-da0d-4f65-897a-74c3938aa47c"
        recordSucceeded   {:submission submission
                           :workflow   workflowSucceeded
                           :status     "Succeeded"}
        records           [recordFailed recordSucceeded]]
    (slack/start-notification-loop slack/notifier)
    (notify-on-workflow-completion workload records))
```
And the result:

Screen Shot 2022-01-14 at 5 54 14 PM

https://broadinstitute.slack.com/archives/C026PTM4XPA/p1642200996000500
https://broadinstitute.slack.com/archives/C026PTM4XPA/p1642200997000600

The system test `wfl.system.v1-endpoint-test/test-workload-sink-outputs-to-tdr` also generates a Slack notification when its one workflow completes:

Screen Shot 2022-01-21 at 1 48 39 PM

https://broadinstitute.slack.com/archives/C026PTM4XPA/p1642538862002600

But, the links on such a message will break once the test's temporary workspace is deleted.

### Review Instructions

A good place to start is to look at example Slack notifications and click around their links (see Manual Verification above).

Then look at the new documentation: [spin up a local docsite](https://github.com/broadinstitute/wfl/blob/develop/docs/readme.md) then navigate to http://localhost:8000/staged-workload/#slack-notifications-for-watchers

When getting into code, start at `wfl.executor/notify-on-workflow-completion` (note that we call it after updating workflow statuses).


api/src/wfl/service/firecloud.clj
```diff
@@ -11,6 +11,18 @@
            [java.util UUID]
            [wfl.util UserException]))
 
+(defn terra-ui-url
```
okotsopoulos, line 14: We use these URL helpers to generate links in Slack notifications, but do not interact with the corresponding services programmatically.

To move quickly I pulled them into `firecloud` since that's the service powering Terra, but I'd move these somewhere else if preferred.

api/src/wfl/service/slack.clj
```diff
@@ -77,12 +84,17 @@
...
 ;;
-(defn notify-channels
-  "Notify `watchers` of an `exception` in the workload with `uuid`."
-  [channels uuid exception]
-  (let [message        (format "Workload %s update threw %s" uuid (.getMessage exception))]
+(defn notify-watchers
+  "Send `message` associated with workload `uuid` to Slack `watchers`."
+  [{:keys [watchers uuid project labels] :as _workload} message]
+  (let [channels (filter slack-channel-watcher? watchers)
+        project  (or project (util/label-value labels "project"))
```
okotsopoulos, line 91: Annoyingly, loading a staged workload doesn't expose the project directly, we need to pull it from the labels list.

docs/md/staged-workload.md
```diff
@@ -25,21 +25,11 @@ The specific values below are from the
...
-    Your channel needs to live in the `broadinstitute.slack.com` Slack organization
-    and you also have to `/invite @WorkFlow Launcher Notifier`
-    to invite the WFL notifier to your channel
-    before the channel can receive user facing messaged from WFL.
-
 ```
 {
     "watchers": [
         ["slack", "C000XXX0XXX"],
-        ["email", "tester@broadinstitute.org"]
```
okotsopoulos, line 42: I removed all traces of emails as watchers from these docs.  They're misleading, since we don't do anything with them.

api/src/wfl/service/slack.clj
tbl3rd: Better name though.

api/src/wfl/util.clj
```diff
@@ -562,6 +562,16 @@
...
 
+(defn label-value
+  "Return first value associated with `label-name` in `labels`
+  or nil if no match."
+  [labels label-name]
+  (when-let [matched (->> labels
+                          (filter #(str/starts-with? % (str label-name ":")))
+                          first)]
+    (let [[_name value & _rest] (str/split matched #":" 3)]
+      value)))
```
tbl3rd, line 573: Might also deconstruct labels into a map but OK.
Or maybe repeated keys are allowed or something.

api/test/wfl/integration/executor_test.clj
```diff
@@ -67,15 +67,16 @@
...
 ;; Snapshot and snapshot reference mocks
-(def ^:private snapshot
-  {:name   "test-snapshot-name"
+(def ^:private snapshot-name "test-snapshot-name")
+(def ^:private snapshot-table-name "snapshot-table")
+(defn ^:private mock-datarepo-snapshot [& _]
+  {:name   snapshot-name
    :id     (str (UUID/randomUUID))
-   :tables [{:name "table"}]})
+   :tables [{:name snapshot-table-name}]})
```
tbl3rd, line 75: This is fine, but as a reminder to discuss this …
Consider working with an expected data structure instead of breaking this down to strings.

api/test/wfl/integration/executor_test.clj
```diff
@@ -482,11 +485,11 @@
...
-              (is (= rootEntityTypePost (:rootEntityType mc))
+              (is (= snapshot-table-name (:rootEntityType mc))
                   "Method configuration root entity type should be snapshot's table name"))]
       (with-redefs-fn
         {#'firecloud/method-configuration        firecloud-method-configuration
-         #'datarepo/snapshot                     datarepo-snapshot
+         #'datarepo/snapshot                     mock-datarepo-snapshot
          #'firecloud/update-method-configuration verify-firecloud-update-params}
         #(#'executor/update-method-configuration! executor reference))
       (let [reloaded (reload-terra-executor executor)]
```
tbl3rd, line 495: When I see mocks stacked 8 deep I wonder what is left to actually test.
(Unfortunately I have no better idea of what to do now.)

api/test/wfl/system/v1_endpoint_test.clj
```diff
@@ -453,6 +453,10 @@
               "a workflow should have been created")
           (finally
             (endpoints/stop-workload workload)))
+        ;; Note: when the workload's workflows have finished,
+        ;; we expect a notification for each workflow
+        ;; to be emitted to the Slack channels in
+        ;; wfl.tools.workloads/watchers.
```
tbl3rd, line 459: Does this comment substitute for verifying the notifications arrived?

api/test/wfl/tools/workloads.clj
```diff
@@ -25,7 +25,8 @@
 
 (def watchers
   [["email" "hornet@broadinstitute.org"]
-   ["slack" "C026PTM4XPA"]])
+   ["slack" "C026PTM4XPA"]  ;; Private channel #hornet-slack-app-testing
+   ])
```
tbl3rd, line 29: Comments like this are evidence of poor design.
(Looks like the channel name should be in the data.)

api/test/wfl/unit/executor_test.clj
```diff
@@ -133,3 +134,18 @@
...
+        workload {:executor {:workspace "workspaceNs/workspaceName"}}]
+    (letfn [(mock-workflow-finished-slack-msg
+              [_executor {:keys [status] :as _record}]
+              (is (cromwell/final? status)
+                  "Should not notify for non-final workflows"))]
+      (with-redefs-fn
+        {#'executor/workflow-finished-slack-msg mock-workflow-finished-slack-msg
+         #'slack/notify-watchers                (constantly nil)}
+        #(is (= records (#'executor/notify-on-workflow-completion workload records))
+             "Should return all passed-in records")))))
```
tbl3rd, line 151: Ah. Good! This is an example of what I wanted to talk about above.
"Consider working with an expected data structure instead of breaking this down to strings."

tbl3rd (approved):

api/src/wfl/executor.clj
```diff
@@ -265,12 +266,47 @@
...
+  (let [emoji                 (cromwell/emoji status)
+        workflow-slack-link   (slack/link (workflow-url workflow) workflow)
+        submission-slack-link (slack/link (submission-url executor submission)
+                                          submission)]
+    (str/join \newline
+              [(format "*%s Workflow %s %s*" emoji workflow-slack-link status)
+               (format "Submission *%s*" submission-slack-link)])))
+
+;; Should only be called while updating active workflows.
+;;
```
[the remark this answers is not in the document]
okotsopoulos: I can update this comment.

We don't want repeated notifications for the same workflow completing, so want to only notify as part of recognizing a state change.  (We don't currently fetch completed workflows in `executor` save for peeking or popping the workflow queue.)

api/src/wfl/executor.clj
```diff
@@ -265,12 +266,47 @@
...
+       (run! #(slack/notify-watchers workload %)))
+  records)
+
 ;; Workflows in newly created firecloud submissions may not have been scheduled
 ;; yet and thus do not have a workflow uuid. Thus, we first "allocate" the
 ;; workflow in the database and then later assign workflow uuids.
 (defn ^:private update-unassigned-workflow-uuids!
   "Assign workflow uuids from previous submissions"
-  [{:keys [workspace details] :as executor}]
+  [{{:keys [workspace details] :as executor} :executor :as workload}]
```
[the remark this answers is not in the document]
okotsopoulos: I only just learned about it myself.  I've found it easier to read the code, maybe at the expense of reading the arguments initially.

api/src/wfl/executor.clj
```diff
@@ -356,8 +394,8 @@
           submission  (create-submission! executor userComment entity)]
       (allocate-submission executor entity submission)
       (stage/pop-queue! source)))
-  (update-unassigned-workflow-uuids! executor)
-  (update-terra-workflow-statuses! executor)
+  (update-unassigned-workflow-uuids! workload)
+  (update-terra-workflow-statuses! workload)
```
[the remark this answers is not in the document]
okotsopoulos: Same.  Thanks @rfricke-asymmetrik for threading the workload through as part of setting user comments in submissions :)

api/src/wfl/service/cromwell.clj
```diff
@@ -25,6 +25,21 @@
...
+        running   ":cromwell-running:"
+        succeeded ":cromwell-succeeded:"
+        failed    ":cromwell-failed:"
+        default   ":cromwell:"]
+    (case status
+      ("On Hold" "Submitted") submitted
+      ("Running" "Aborting")  running
+      "Succeeded"             succeeded
+      ("Failed" "Aborted")    failed
+      default)))
```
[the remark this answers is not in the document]
okotsopoulos: I agree with you re: bindings, I will debulk.

api/src/wfl/util.clj
okotsopoulos: Yeah, I think repeated keys are allowed (for staged workloads at least):

https://github.com/broadinstitute/wfl/blob/develop/api/src/wfl/module/covid.clj#L68-L73

api/test/wfl/system/v1_endpoint_test.clj
okotsopoulos: Intended more of a heads up, but you're right that a verification would be more appropriate… unfortunately I'm not sure how to do that, the pending Slack test indicates that its attempt at verification is flaky when run in GitHub Actions:

https://broadinstitute.atlassian.net/browse/GH-1553

api/test/wfl/tools/workloads.clj
okotsopoulos: Good point.  We went with channel IDs because they are immutable, the channel names can change.  But it's a PITA to track back to the name, don't think you can do it within the Slack App at least.

Possible improvements --

- Allow a Slack channel name to be specified instead of channel ID at workload creation, internally at that point we map it to its channel ID.  [No Slack API](https://stackoverflow.com/questions/50106263/slack-api-to-find-existing-channel) for fetching a channel by name, so we'd have to fetch all channels and filter ourself.
- Allow a Slack channel name to be specified as decoration in addition to channel ID, we don't do anything with it programmatically or guarantee that it won't change in the future but helps point to where the notifications are going.

api/test/wfl/integration/executor_test.clj
okotsopoulos: I am in the same boat.  When I add new tests these days, I'm more inclined to unit over integration for this reason, but that's me kicking the can.

api/test/wfl/tools/workloads.clj
tbl3rd: Yes. I meant to advocate your "decoration" idea.

tbl3rd: The slack channel ID is just that: an identifier. We should also support a name to support debugging.

okotsopoulos: Sweet.  I think the decorator idea may work out of the box.  I will test it out.

tbl3rd: > Sweet. I think the decorator idea may work out of the box. I will test it out.

The name can be optional, or default to the `::creator`, or …

okotsopoulos: Optional name works fine, this notification was triggered in system tests with the updated `[slack <channel-id> <channel-name>]` watcher.

https://broadinstitute.slack.com/archives/C026PTM4XPA/p1643055770000300

The channel name also appears when fetching the workload:

Screen Shot 2022-01-24 at 3 06 02 PM

I supplemented public-facing documentation to suggest this.

PR #560, opened 2022-01-24
GH-1578 TerraDataRepoSource filters old rows before reducing

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1578

Addresses an edge case bug when `TerraDataRepoSource` polls for new rows to snapshot.

The desired behavior: if a snapshot creation job fails, we should automatically reconsider its rows for snapshotting until they are snapshotted successfully or they age out of our 2 hour check window.

Bug: if a workload's first snapshot creation job fails, we did not reconsider its rows for snapshotting.

Fix: filter old snapshot creation job rows by status prior to reducing to a single object with all old row IDs.  This filtration previously took place within the reduction, but If you pass a singleton sequence to `reduce` it is returned as it's passed in. https://clojuredocs.org/clojure.core/reduce


okotsopoulos:
System tests pass:

```
wm111-e35:wfl okotsopo$ make TARGET=system
export CPCACHE=/Users/okotsopo/wfl/api/.cpcache;            \
	export WFL_WFL_URL=http://localhost:3000; \
	clojure  -M:parallel-test wfl.system.v1-endpoint-test | \
	tee /Users/okotsopo/wfl/derived/api/system.log
WARNING: Specified path is external to project: ../derived/api/src
WARNING: Specified path is external to project: ../derived/api/resources

Ran 32 tests containing 367 assertions.
0 failures, 0 errors.
```

api/test/wfl/unit/source_test.clj
```diff
@@ -33,3 +33,54 @@
...
+            (#'source/filter-and-combine-tdr-source-details
+             [running succeeded failed])]
+        (is (= (concat running-rows succeeded-rows) datarepo_row_ids)
+            "running and succeeded rows should be kept")
+        (is (= (:start_time running) start_time)
+            "Earliest start time should be taken")
+        (is (= (:end_time succeeded) end_time)
+            "Latest end time should be taken")))
+    (testing "Keep no rows"
+      (is (nil? (#'source/filter-and-combine-tdr-source-details
```
okotsopoulos, line 84: This is the case where we experienced the bug, more info in the ticket description.

api/src/wfl/source.clj
```diff
@@ -164,7 +170,7 @@
                     (log-prefix source) (:name dataset) table begin end))
   (let [wfl   (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
                 (postgres/get-table tx details))
-        old   (when (seq wfl) (reduce combine-tdr-source-details wfl))
+        old   (filter-and-combine-tdr-source-details wfl)
```
tbl3rd, line 173: This handles the `nil` case.

api/test/wfl/unit/source_test.clj
tbl3rd: Because `failed` is too old or because it has only `failed-rows`?
(The latter, right?)
Then is there a "snapshot request too old but has succeeded rows" case?

tbl3rd (approved):

api/test/wfl/unit/source_test.clj
okotsopoulos: Correct, the latter.  If rows are too old, either way they will not get considered for reprocessing (expected behavior.)

api/src/wfl/source.clj
okotsopoulos: Yes, and I will add another test case to be explicit.

PR #561, opened 2022-01-25
GH-1433 Test TDR snapshot creation within temporary dataset

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1433

In normal operations, WFL does not programmatically delete TDR snapshots.  The exception: within the `create-temporary-snapshot` fixture used to clean up in testing.

Deleting a snapshot obtains an exclusive lock on the underlying dataset, and can block other concurrent operations on that dataset.  We've seen `wfl.integration.datarepo-test/test-create-snapshot` fail when run concurrently for this reason since it used a fixed existing dataset.

Solution: test snapshot creation within a temporary dataset, per Trevyn's recommendation.

### DIFF
--- a/api/test/wfl/integration/datarepo_test.clj
+++ b/api/test/wfl/integration/datarepo_test.clj
@@ -60,48 +60,53 @@
    :integer "outint"
    :string  "outstring"})
 
-(deftest test-ingest-pipeline-outputs
+(deftest test-ingest-pipeline-outputs-and-snapshot
   (let [dataset-json "testing-dataset.json"
         table-name   "parameters"
         tdr-profile  (env/getenv "WFL_TDR_DEFAULT_PROFILE")]
     (fixtures/with-fixtures
-      [(fixtures/with-temporary-cloud-storage-folder fixtures/gcs-tdr-test-bucket)
+      [(fixtures/with-temporary-cloud-storage-folder
+         fixtures/gcs-tdr-test-bucket)
        (fixtures/with-temporary-dataset
          (datasets/unique-dataset-request tdr-profile dataset-json))]
       (fn [[temp-bucket dataset-id]]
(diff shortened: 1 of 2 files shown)


api/test/wfl/integration/datarepo_test.clj
```diff
@@ -60,48 +60,53 @@
...
+            (is (== 0 bad_row_count))
+            (let [row-ids
+                  (-> dataset
+                      (datarepo/query-table table-name [:datarepo_row_id])
+                      :rows
+                      flatten)]
+              (is (== 1 (count row-ids))
+                  "Single input row should have been written to the dataset")
+              (testing "creating snapshot after completed ingest"
+                (fixtures/with-temporary-snapshot
```
okotsopoulos, line 99: Even though we are creating a snapshot within a temporary dataset -- meaning the dataset will be deleted at the end of the test -- we still need to create a temporary snapshot.  TDR won't let you delete a dataset with snapshots in it.

tbl3rd: And thank you for the tabular formatting (and the = hygiene) too.

api/test/wfl/tools/fixtures.clj
```diff
@@ -193,6 +193,10 @@
    datarepo/delete-snapshots-then-dataset
    f))
 
+;; Recommendation: only call this within a temporary dataset.
+;; If calling within a fixed dataset, running the same test concurrently
+;; may fail due to exclusive dataset locking on snapshot deletion.
+;;
```
tbl3rd, line 199: IWBNI we could enforce this some way, but I guess all those set-up and tear-down calls also have to be synchronous.

tbl3rd (approved):
I hope this is the root of the problem.

PR #562, opened 2022-02-02
GH-1539 Stop started workloads in system tests

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1539

We automate a nightly run of `wfl.system.v1-endpoint-test` on WFL dev.  Lately, the run as a whole fails with scattered "503 Service Unavailable" errors.  [Last night's run](https://github.com/broadinstitute/wfl/runs/5029975554?check_suite_focus=true)

I suspect this is because the volume of active workloads on WFL dev has grown over time.  Each full system test run adds 3 active AoU workloads to the WFL instance.  Because AoU workloads are continuous, they will poll for new inputs until explicitly stopped.

TODO: add WFL dev breakdown here.

(Now, if tests exit atypically it's still possible that they'll have `start`ed a continuous workload without `stop`ping it.  Or any active workload, continuous or otherwise, whose test exits prematurely may never get marked as `finished` if needed executor resources were removed, for example.  But these changes will fix the main cause of unchecked active workload growth in WFL dev.)

### Changes

Wrap tests which`start` AoU workloads with a `(try … finally ...)` to stop them.

### Review Instructions

Note the number of active workloads in your DB:

```
wfl=# select count(*), pipeline from workload where started is not null and finished is null group by pipeline;
 count | pipeline 
-------+----------
(0 rows)
```

Run system tests off of local WFL:

```
make TARGET=system
…
```

After they complete without interruption, observe that the number of active workloads in your DB is unchanged:

```
wfl=# select count(*), pipeline from workload where started is not null and finished is null group by pipeline;
 count | pipeline 
-------+----------
(0 rows)
```

### To Do After Merge

I can manually mark all active test workloads in WFL dev as `finished`, or we can blow away the WFL dev DB.  Let's talk.

### DIFF
--- a/api/test/wfl/system/v1_endpoint_test.clj
+++ b/api/test/wfl/system/v1_endpoint_test.clj
@@ -105,13 +105,16 @@
   [{:keys [uuid pipeline] :as workload}]
   (testing (format "calling api/v1/start with %s workload" pipeline)
     (let [workload (endpoints/start-workload workload)]
-      (is (= uuid (:uuid workload)))
-      (is (:started workload))
-      (let [workflows (endpoints/get-workflows workload)]
-        (is (every? :updated workflows))
-        (is (every? :uuid workflows)))
-      (verify-internal-properties-removed workload)
-      (workloads/when-all-workflows-finish verify-succeeded-workload workload))))
+      (try
+        (is (= uuid (:uuid workload)))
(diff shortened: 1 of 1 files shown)


tbl3rd (approved):
Thanks. I didn't even compile this.

PR #595, opened 2022-03-24
[GH-1433] No TDR temporary snapshots in tests

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1433

Previous work in this space: https://github.com/broadinstitute/wfl/pull/561

I found this new improvement in my "stash graveyard".

In integration tests which interact with TDR, we don't actually have to create temporary snapshots within temporary datasets (and cross our fingers that developers won't use temporary snapshots within fixed datasets).  You might remember that deleting a snapshot obtains an exclusive lock on the underlying dataset, which lead to flapping test failures on concurrent runs.

Instead, we can create snapshots as normal within temporary datasets: temporary datasets automatically delete any of their snapshots before the dataset.

### DIFF
--- a/api/test/wfl/tools/fixtures.clj
+++ b/api/test/wfl/tools/fixtures.clj
@@ -193,18 +193,6 @@
    datarepo/delete-snapshots-then-dataset
    f))
 
-;; Recommendation: only call this within a temporary dataset.
-;; If calling within a fixed dataset, running the same test concurrently
-;; may fail due to exclusive dataset locking on snapshot deletion.
-;;
-(defn with-temporary-snapshot
-  "Create a temporary Terra Data Repository Snapshot with `snapshot-request`"
-  [snapshot-request f]
-  (util/bracket
-   #(datarepo/create-snapshot snapshot-request)
-   datarepo/delete-snapshot
-   f))
-
 (defn with-temporary-workspace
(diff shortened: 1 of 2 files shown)


api/test/wfl/tools/fixtures.clj
```diff
@@ -193,18 +193,6 @@
    datarepo/delete-snapshots-then-dataset
```
okotsopoulos, line 193: Here's where we clean up any snapshots as part of cleaning up a temporary dataset.

api/test/wfl/tools/fixtures.clj
```diff
@@ -193,18 +193,6 @@
...
-;; If calling within a fixed dataset, running the same test concurrently
-;; may fail due to exclusive dataset locking on snapshot deletion.
-;;
-(defn with-temporary-snapshot
-  "Create a temporary Terra Data Repository Snapshot with `snapshot-request`"
-  [snapshot-request f]
-  (util/bracket
-   #(datarepo/create-snapshot snapshot-request)
-   datarepo/delete-snapshot
-   f))
```
tbl3rd, line 206: Thanks again.

tbl3rd (approved):

PR #597, opened 2022-03-24
[GH-1623] workloads/update-workload! should define own transactions

okotsopoulos:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1623

As a prerequisite to factoring out external calls from DB transactions, this PR forces `workloads/update-workload!` implementations to define their own transactions.

With these changes, this decoupling is complete for staged workloads.  Staged workloads can be updated in their own future in a separate PR.

Legacy workload update behavior otherwise remains the same.

### System Tests

Pass:
```
wm111-e35:wfl okotsopo$ make TARGET=system
export CPCACHE=/Users/okotsopo/wfl/api/.cpcache;            \
	export WFL_WFL_URL=http://localhost:3000; \
	clojure  -M:parallel-test wfl.system.v1-endpoint-test | \
	tee /Users/okotsopo/wfl/derived/api/system.log
WARNING: Use of :paths external to the project has been deprecated, please remove: ../derived/api/src
WARNING: Use of :paths external to the project has been deprecated, please remove: ../derived/api/resources

Ran 33 tests containing 374 assertions.
0 failures, 0 errors.
```

### Review Instructions

I recommend hiding whitespace when reviewing.  Wouldn't mind eyes on the code consolidation I did in `wfl.server.clj`.

### DIFF
--- a/api/src/wfl/module/aou.clj
+++ b/api/src/wfl/module/aou.clj
@@ -314,14 +314,18 @@
 
 (defmethod workloads/update-workload!
   pipeline
-  [{:keys [started finished] :as workload}]
-  (letfn [(update! [{:keys [id] :as workload}]
-            (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+  [{:keys [id started stopped finished] :as _workload-record}]
+  (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+    (letfn [(load-workload []
+              (workloads/load-workload-for-id tx id))
+            (update!       [workload]
               (batch/update-workflow-statuses! tx workload)
-              (when (:stopped workload)
+              (when stopped
                 (batch/update-workload-status! tx workload))
-              (workloads/load-workload-for-id tx id)))]
-    (if (and started (not finished)) (update! workload) workload)))
+              (load-workload))]
+      (if (and started (not finished))
+        (update! (load-workload))
+        (load-workload)))))
 
 (defoverload workloads/workflows            pipeline aou-workflows)
 (defoverload workloads/workflows-by-filters pipeline aou-workflows-by-filters)
--- a/api/src/wfl/module/aou.clj
+++ b/api/src/wfl/module/aou.clj
@@ -314,12 +314,13 @@
 
 (defmethod workloads/update-workload!
   pipeline
-  [tx {:keys [started finished] :as workload}]
+  [{:keys [started finished] :as workload}]
   (letfn [(update! [{:keys [id] :as workload}]
-            (batch/update-workflow-statuses! tx workload)
-            (when (:stopped workload)
-              (batch/update-workload-status! tx workload))
-            (workloads/load-workload-for-id tx id))]
+            (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+              (batch/update-workflow-statuses! tx workload)
+              (when (:stopped workload)
+                (batch/update-workload-status! tx workload))
+              (workloads/load-workload-for-id tx id)))]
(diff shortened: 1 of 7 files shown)


api/src/wfl/module/staged.clj
```diff
@@ -125,17 +125,18 @@
...
-            (patch-workload tx workload {:updated now})
-            (when (every? stage/done? [source executor sink])
-              (patch-workload tx workload {:finished now}))
-            (workloads/load-workload-for-id tx id))]
+            (jdbc/with-db-transaction [tx (postgres/wfl-db-config)]
+              (patch-workload tx workload {:updated now})
+              (when (every? stage/done? [source executor sink])
+                (patch-workload tx workload {:finished now}))
+              (workloads/load-workload-for-id tx id)))]
     (if (and started (not finished)) (update! workload (utc-now)) workload)))
```
tbl3rd, line 140: It isn't new in this PR but noticed that update-sink! and patch-workload run in separate TXs?
I guess workload state will converge. (Batch updates are simpler here.)

api/src/wfl/server.clj
```diff
@@ -78,29 +78,23 @@
...
-  (try
-    (log/info "Updating workload"
-              :workload (workloads/to-log workload))
-    (do-update! workload)
+          (slack/notify-watchers workload (.getMessage e)))))
     (catch Throwable t
       (log/error "Failed to update workload"
-                 :workload  (workloads/to-log workload)
+                 :workload  (workloads/to-log workload-record)
                  :throwable t))))
```
tbl3rd, line 98: Now the initial load and the update run in separate TXs.
Would it be safer to run the load in the update's transaction?
(Not sure how those overlap in the schema.)

I prefer a separate backstop try to the nesting, but that looks correct.
I like the workload-record distinction.

tbl3rd (approved):
Not in a place where I can run this now.
I worry about fragmenting the update transaction.
Can overlapping updates on a single workload interfere?

api/src/wfl/server.clj
okotsopoulos: Yeah, good points all around.

I will see about punting the workload load to the update transactions.  Originally I thought I needed the loaded workload to Slack on UserExceptions, but the workload record is sufficient.

I was on the fence about consolidating the backstops anyways (something about try-within-a-try is hard to track) so will un-nest.

tbl3rd: I can manage try-within-try on a good day, but not every day.

okotsopoulos: Workload loading now takes place in the update's transactions, and try-within-a-try is no more.

api/src/wfl/module/staged.clj
okotsopoulos: Yes, all of the individual stages defined transactions as needed to avoid external calls within open transactions.

But yes, there are thread safety concerns where retries are concerned:

https://broadinstitute.atlassian.net/browse/GH-1423

api/src/wfl/module/aou.clj
```diff
@@ -314,13 +314,18 @@
...
+    (letfn [(load-workload []
+              (workloads/load-workload-for-id tx id))
+            (update!       [workload]
+              (batch/update-workflow-statuses! tx workload)
+              (when stopped
+                (batch/update-workload-status! tx workload))
+              (load-workload))]
+      (if (and started (not finished))
+        (update! (load-workload))
+        (load-workload)))))
```
tbl3rd, line 328: Thanks.

tbl3rd (approved):

PR #607, opened 2022-05-13
[GH-1658] Shouldn't wfl.service.rawls/batch-upsert handle Boolean attributes?

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1658

### Changes

- Add `nil?` and `boolean?` attributes to the test entities file.
- Fix what broke.

### Review Instructions

- Ignore whitespace changes.
- Verify that the removed code is not needed.


api/src/wfl/service/rawls.clj
```diff
@@ -136,19 +136,13 @@
...
           (to-operations [attributes]
             (flatten
              (for [[k v] (seq attributes)]
-               (cond (number? v) (add-scalar k v)
-                     (string? v) (add-scalar k v)
-                     (map?    v) (add-scalar k v)
-                     (coll?   v) (add-list   k v)
-                     (nil?    v) (no-op)
-                     :else       (on-unhandled-attribute k v)))))
+               (cond (map?  v) (add-scalar k v)
```
okotsopoulos, line 142: (Not important but curious) is there a reason that this conditional is broken out separately, rather than falling back to the `:else`?

okotsopoulos (approved):
Thanks for making this change, expanding the test resource, and removing unused code.  I like this "innocent until proven guilty" approach when processing attributes.

api/src/wfl/service/rawls.clj
tbl3rd: A map is a collection but we handle it as an "object".
COND processes clauses in order so the map? test has to come before coll?.
(Yes. I had to debug that mistake so it is fresh in mind.)

PR #608, opened 2022-05-23
[GH-1654] Admit memory retry multiplier.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1654

### Changes

- Add a `memory_retry_multiplier` column to the `TerraExecutor` table.
- Map `:memoryRetryMultiplier` workload `executor` attribute to the new column.
- Test that the new attribute round-trips through the database.

### Review Instructions

- No instructions.

### DIFF
--- a/api/src/wfl/executor.clj
+++ b/api/src/wfl/executor.clj
@@ -88,6 +88,7 @@
 ;; Specs
 (s/def ::entity ::all/uuid)
 (s/def ::fromSource string?)
+(s/def ::memoryRetryMultiplier number?)
 (s/def ::methodConfiguration (s/and string? util/terra-namespaced-name?))
 (s/def ::methodConfigurationVersion integer?)
 (s/def ::reference ::all/uuid)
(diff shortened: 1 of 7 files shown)


aednichols (approved):

okotsopoulos (changes requested):
What is this ticket trying to accomplish?  Is it meant to allow a caller to set a memory retry multiplier for Terra submissions?

Presently while this multiplier is saved in the DB, WFL doesn't do anything with it.  The executor will also need to be altered to pass this multiplier to Rawls / Firecloud when initiating a submission.

tbl3rd:
Thanks. Wanna give this another look, @okotsopoulos ?

okotsopoulos (approved):
That looks like it should do it, thanks.

It would be nice to include this new supported executor field in the public-facing docs, I know Morgan and Sushma reference them: https://broadinstitute.github.io/wfl/staged-executor/#terra-executor

tbl3rd:
Good idea. Will do.

PR #612, opened 2022-06-09
[GH-1671] create-submission! swaps arguments to submit-method.

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1671

### Changes

- The `userComment` param is optional and must appear last.

### Review Instructions

- No instructions.


okotsopoulos (approved):
I've been bit by argument reordering before, glad you found it.

PR #614, opened 2022-07-14
[GH-1678] When monitoring source load tags, match prefix instead of exact match. 

tbl3rd:
### Purpose

- https://broadinstitute.atlassian.net/browse/GH-1678

### Changes

- Factor out retry filter keys.
- Rename `terra-executor-workflows-sql-params` to `filter-query-for-unretried-workflows`.
- Privatize some "public for testing" vars.
- Prefix match `loadTag` with `LIKE` on snapshot queries.
- Document changes. 

### Review Instructions

- Start with `datarepo.clj` and `staged-source.md`.
- Scan the rest while ignoring whitespace changes.

### DIFF
--- a/api/src/wfl/api/handlers.clj
+++ b/api/src/wfl/api/handlers.clj
@@ -1,16 +1,17 @@
 (ns wfl.api.handlers
   "Define handlers for API endpoints."
-  (:require [clojure.set                    :refer [rename-keys]]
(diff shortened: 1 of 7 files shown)


nmalfroy (approved):
I think this looks ok?  I'm sorta guessing at a lot of this clojure syntax

PR #636, opened 2024-05-09
WM-2617: Bump markdown to unblock release

cjllanwarne:


docs/requirements.txt
```diff
@@ -1,4 +1,3 @@
-pymdown-extensions==8.0.1
-markdown-include==0.6.0
-mkdocs-material==7.2.4
-jinja2==3.0.3
```
salonishah11, line 4: just to confirm that this package is no longer needed?

salonishah11 (approved):

docs/requirements.txt
cjllanwarne: It no longer needs to be pinned

mspector (approved):

